PR #493, opened 2014-05-17
Feature fraccalc

kuanxu:
Enabling differentiation and integration of fractional order.


nickhale:
I've made a few changes to this, which I've pushed to https://github.com/chebfun/chebfun/tree/feature-frac-calc-nh. I think the changes are a lot better in terms of encapsulation and structure, but there's currently a fairly significant hit in terms of speed. more work is needed.

PR #1197, opened 2014-10-10
Fix trigcf

mhsnjvd:
This branch allows CF approximations for trigonometric functions. Some tests have also been added.


mhsnjvd:
This is now implemented via #1415

PR #1251, opened 2014-11-05
Rapid hacking session fixes

### LINKED ISSUE #1066: Chebfun2 enters infinite loop for functions that cannot be resolved
The following causes the chebfun2 `constructor` to enter an infinite loop.
```
f = chebfun2(@(x,y) sign(x).*sign(y))
```
Use caution running this code as it will eventually use up all available memory on your machine.

The issue appears to be in PHASE 2 of the construction process.  The loop in question (starting at line 290) does not check the `failure` flag as one of the options for loop termination, it only checks `ishappy`.

ajt60gaibb:
This branch contains a bunch of rapid fixes for the issues that came up during the Chebfun2 hacking session. 

Closes #1066.


aaustin141:
Argh... sorry for the delay on this.  I've taken a look at the new fixes, and everything seems in order save for one thing.  Trying @kuanxu's function from the hacking session one more time, I find that it crashes again but with a different error:
```
>> x = ones(1, 2);
>> f = chebfun2(x, 'coeffs');
>> g = chol(f);
Subscript indices must either be real positive integers or logicals.
Error in chebfun2/chol (line 82)
symTest = ( abs(f(r,s) - f(s,r)) < max(epslevel(f.cols),epslevel(f.rows)) );
Error in chebfun2/chol (line 53)
    varargout = {chol( f )};  
```
I think the error here is that `f(r, s)` is meant to do an evaluation of `f` at the point `(r, s)`; however, that won't work inside of a `chebfun2` method because the `subsref` that gets called in such a context will be the built-in MATLAB one and not the `chebfun2` overload.  If I change `f(r, s)` and `f(s, r)` to `feval(f, r, s)` and `feval(f, s, r)`, respectively, this crash no longer occurs; however, I find that the issue I mentioned in my original post still isn't fixed.  I believe this happens because the conditional test for the new "function is not symmetric" error is inverted:  it should be `if ( ~symTest )`, not `if ( symTest )`.

Finally, I observe that the point which is being used to check for function symmetry is selected randomly.  This is great; however, it hinders our ability to reproduce results.  Would it be OK to change it to a "fixed but arbitrary" point, more or less like we do for subdivision when rootfinding?

@ajt60gaibb, can you fix

trefethen:
I believe we have a longstanding policy that Chebfun should not rely on random numbers, since it is disturbing to users when things are not reproducible.  So I would certainly support replacing random by arbitrary.

ajt60gaibb:
Fixed. Thanks. (Apologizes for using `rand`, I should have known better.)

aaustin141:
This is finally closed.  The only additional change I had to make was to slightly loosen the tolerance on the symmetry check to get `tests/chebfun2/test_chol.m` to pass on my machine.  Again, my apologies for taking so long to get these "rapid" fixes merged!

PR #1277, opened 2014-11-20
Feature trigtech plateauCheck

Hadrien-Montanelli:
`plateauCheck` for `trigtech` is now used for ODEs solved with Fourier collocation!


asgeirbirkis:
The main reason why I haven't been happy to merge this up to this date is because of the overlap with the ```chebtech/plateauCheck``` method, as described above. Hopefully, we'll have some new thoughts on how to make it such that we can reuse much of the same code for both classes. Until then, we should not merge this request.

mhsnjvd:
more discussion needed before we can merge this.

PR #1327, opened 2014-12-07
Feature doublelength option

hrothgar:
This branch adds support for a `doubleLength` option in the Chebfun constructor, which when supplied returns a chebfun of length twice what the constructor decides.

`doubleLength` is currently not supported in combination with `splitting on` because the necessary changes to the construction process are more far-reaching.


nickhale:
05a0975 attempts to deal with the `f = chebfun(1, 10)` issue.

For scalars it would have been easy, but we should also allow `chebfun([1 2], 10)` to return something meaningful, so I prolonged the result after the construction, rather than extending the vector beforehand.

asgeirbirkis:
@hrothgar could you please look at the change made by @nickhale above, so that we can attempt to merge this asap?

hrothgar:
@nickhale, what did 05a0975 fix? I am getting the same errors for each of these:
```
f = chebfun(1, 'doublelength');
f = chebfun(1,10);
f = chebfun([1; 2], 10);
```

hrothgar:
Doublelength always returns a function with even length, so for trigtechs the result looks asymmetric on plots:

doublelength_trig

Should doublelength return a chebfun of length 2*n+1 to make this look nicer? @trefethen, what do you think?

(I notice there is a bug in plotcoeffs there with the epslevel line, but that is a separate matter.)

trefethen:
Yes, I agree.  Trigfuns normally have odd length and that should happen when they're produced with `doublelength` too.  Thanks for noticing.

hrothgar:
This pull request is ready for review. Any takers?

nickhale:
It seems the 'double-plus-one' length is used for both `chebtech` and `trigtech` based `chebfuns`. is that what we want?

@gradywright  - Would it be better to always force `trigtechs` to have odd lengths?

gradywright:
I think it is better to force `trigtechs` to have odd lengths after double length is called.  As double length corresponds to doubling the degree.  It really only makes sense to talk about the degree of an odd-length trigonometric interpolant.

nickhale:
I'm not sure that answered my question.

How it works at the moment is that we construct as normal, we then query the length, and construct again with `2*length(f)+1`. My point is that we don't want the `+1` for `chebtech`s, but we do for `trigtech`s, but I don't want to make that distinction at the `chebfun` level. What I'd like to do is pass `2*length` down to the constructor, and for the `trigtech` constructor to say "Hey, even lengths don't make sense to me, so I'mma add one to that."

Should the length here be 10 or 11?
```
>> f = chebfun(@sin, [-pi, pi], 'trig', 10)
f =
       interval       length   endpoint values trig
[    -3.1,     3.1]       10  -1.2e-16  1.2e-16 
Epslevel = 1.167358e-15.  Vscale = 9.510565e-01.
````

**Comment by gradywright** [comment]
I don't think it is a good idea to force every `trigfun` to be of even length.  This may generate confusion for users.  For example, the following code should naturally return a `trigfun` of even length:
```
f = chebfun(rand(20,1),'trig')
```
I'm curious what @trefethen thinks.

I'm happy to have double length simply return a `trigfun` with double its actual length for now.

trefethen:
No, please don't change how trigfun is set up.  In writing the code and the paper and the docs we thought this through pretty carefully I think: ordinary constructions always give trigfuns of odd length, but one can specify even length and in some cases that is a good idea.  See the first ten pages of our paper at https://people.maths.ox.ac.uk/trefethen/trigpaper.pdf and the specific comment in the middle of p. 10.

I think `doublelength` for a trigfun should always effectively double the DEGREE: i.e., return a trigfun of odd length.

PR #1356, opened 2015-01-22
Feature fraccalc new

nickhale:
Add support for fractional integrals and derivatives of **smooth** `chebfun` objects.

The key idea is to express the function as a Legendre series and then use explicit expressions for the fractional integrals of these guys from §18.17 on NIST.

For half-integrals and derivatives, this is very fast as we need only `cheb2leg()`. For arbitrary values of `mu` it's a little slower as we must convert Jacobi coefficients back to Chebyshev coefficients. This is currently done in `jac2cheb()` using an O(n^2) direct approach (i.e., using the standard recurrence relation). However, since here `mu` is in (0,1), it should be no problem to use the H-T asymptotic approach that is used in `leg2cheb`.

When it comes to adding support for piecewise smooth functions, this approach can be used to compute the 'local' contributions. 

There is some scope for computing fractional derives of functions with endpoint singularities at the left end point using NIST (18.17.9) (and flipping using symmetry), but I haven't investigated this further. I'm not sure that there are formulas for the case when the singularity is on the right.

It's not clear what to do on unbounded domains (although NIST (18.17.10) may be useful) and I'm not sure what should happen in the `trigtech` case (currently I throw an error).

Edit: typos.

### DIFF
--- a/@bndfun/fracInt.m
+++ b/@bndfun/fracInt.m
@@ -1,0 +1,19 @@
+function f = fracInt(f, mu)
+%FRACINT  Fractional integral of a BNDFUN. 
+%   FRACINT(F, MU) gives the order MU fractional integral of a BNDFUN object F.
+
+% Copyright 2014 by The University of Oxford and The Chebfun Developers.
+% See http://www.chebfun.org/ for Chebfun information.
+
+% Scaling for domain:
+scl = (diff(f.domain)/2)^mu;
+
+if ( ~isa(f.onefun, 'singfun') )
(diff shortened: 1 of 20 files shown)


nickhale:
Update: There's now support for `singfun` objects where the singularity is on the left. Currently an error is thrown if there is a non-zero exponent on the right (after simplification). We'll probably have to resort to brute-forcing the integral in this case.

There's still no support for piecewise smooth functions (we could perhaps attempt to merge?) or unbounded domains.

PR #1359, opened 2015-01-23
Fix singfun chebcoeffs

### LINKED ISSUE #1358: Unexpected behaviour from @SINGFUN/CHEBCOEFFS().
`@singfun/chebcoeffs()` does not return the correct coefficients:

```
n = 14;

x = chebfun('x');
f1 = sqrt(1+x).*exp(x);
c1 = chebcoeffs(f1, n);

f2 = chebfun(@(x) sqrt(1+x).*exp(x), 1000);
c2 = chebcoeffs(f2, n);

f3 = chebfun(@(x) sqrt(1+x).*exp(x), 'trunc', n);
c3 = chebcoeffs(f3, n);

[c1 c2 c3]
```
```
ans =
    2.2992    1.4638    1.4638
    2.3195    1.7951    1.7951
    0.8176    0.4702    0.4702
    0.1380    0.1064    0.1064
    0.0203    0.0037    0.0037
    0.0031    0.0073    0.0073
   -0.0011   -0.0042   -0.0042
    0.0014    0.0032    0.0032
   -0.0013   -0.0025   -0.0025
    0.0012    0.0020    0.0020
   -0.0012   -0.0016   -0.0016
    0.0011    0.0013    0.0013
   -0.0011   -0.0011   -0.0011
    0.0005    0.0010    0.0010
```

I first suspected that it was simply returning the coefficients of `exp(x)`, but that's not the case either.

An additional problem is that if no `N` is given to `chebcoeffs()` then the length of the **smoothPart** of the function is given to 'singfun/chebcoeffs', even though the function will typically have a non-compactly supported Chebyshev expansion. Here's an extreme case:
```
>> chebcoeffs(sqrt(1+x))
ans =
    1.4142
```

I suggest that perhaps `chebcoeffs()` should **always** require an `N` to be passed, otherwise we're expecting the user to know too much about the underlying representation.

nickhale:
This partially fixes #1358.

This is branched off of fix-chebpoly.

### DIFF
--- a/@chebfun/chebcoeffs.m
+++ b/@chebfun/chebcoeffs.m
@@ -73,19 +73,7 @@ if ( any(isinf(f.domain)) )
         'Infinite intervals are not supported here.');
 end
 
-%%
-% We compute 2nd-kind coefficients by computing the 1st-kind coefficients and
-% using a recurrence relation.  The recurrence for the coefficient of U_n
-% requires the coefficients of T_n and T_{n + 2}, so we need to compute two
-% extra 1st-kind coefficients if 2nd-kind coefficients have been requested.
-%
-% TODO:  This is OK for smooth functions but will be expensive for
-% piecewise-smooth ones, which require integrals.  If the function is only
-% piecewise smooth, we should compute the exactly N inner products for the
-% second-kind coefficients directly.
-if ( kind == 2 )
-    N = N + 2;
-end
+%% Merge:
 
 % Try to merge out breakpoints if possible, since integrals are expensive:
 if ( numFuns ~= 1 )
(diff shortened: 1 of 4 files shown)


kuanxu:
@nickhale Shall we throw an error for an unrecognized type (e.g. the second argument 5 in the example below)?
```
>> chebpoly(3,5)
ans =
[      -1,       1]        4        -1        1 
Epslevel = 2.220446e-15.  Vscale = 1.
```

nickhale:
Good idea. Done.

It might be nice to eventually add support for 3rd and 4th kind, but that's far from urgent.

Issue #1026, opened 2014-06-27
Sum and vscale for functions with endpoint singularities on unbounded domains


davidagross:
Based off [another old Trac ticket](https://svn.maths.ox.ac.uk/trac/chebfun/ticket/215) I wanted to check what the current state of the `sum` is by running:
```
f1 = chebfun(@(t) t.^0.5./exp(t), [0,inf], 'exps', [0.5 0])

f2 = chebfun(@(t) t.^0.5./exp(t), [0,1,inf], 'exps', [0.5 0 0 0])

f3 = f1; f3(1) = f3(1)

[sum(f1); sum(f2); sum(f3)]
```
which yields:
```
f1 =
 
   chebfun column (1 smooth piece)
       interval       length   endpoint values   endpoint exponents
[       0,     Inf]       44         0 -6.7e-16         [0.5      0]  
Epslevel = 1.000000e-14.  Vscale = 2.738613e+00.

 
f2 =
 
   chebfun column (2 smooth pieces)
       interval       length   endpoint values   endpoint exponents
[       0,       1]       12         0     0.37         [0.5      0]  
[       1,     Inf]       60      0.37    1e-15         [0      0]  
Epslevel = 1.000000e-14.  Vscale = 7.071068e-01.  Total length = 72.

 
f3 =
 
   chebfun column (2 smooth pieces)
       interval       length   endpoint values   endpoint exponents
[       0,       1]    65537   1.2e-05     0.37         [0      0]  
[       1,     Inf]    65537         0 -4.5e-13         [0.5      0]  
Epslevel = 0.  Vscale = Inf.  Total length = 131074.


ans =

   0.885858294730405
   0.886226925452745
   0.853327814564061
```
`f1` has the wrong vscale and integral, `f2` has the wrong vscale only, and `f3` has the wrong endpoint values along with the wrong vscale and a wrong-er integral.

I tracked this down to [`@unbndfun/restrict`](https://github.com/chebfun/chebfun/blob/development/@unbndfun/restrict.m#L83-L89) which needs something like / at least an
```
...
else
    data.exponents = [];
end
```
to clear the exponents from being erroneously used later in the loop.  Making this change yields:
```
f3 =
 
   chebfun column (2 smooth pieces)
       interval       length   endpoint values  
[       0,       1]    65537   1.2e-05     0.37 
[       1,     Inf]       58      0.37 -7.3e-16 
[...]

PR #1360, opened 2015-01-23
Naively restore exponents before integrating singfun

davidagross:
With reference to #1026, this change and test does the trick of dealing with the accuracy of `sum(f1)`, with the caveat that it feels like the checking for exponents is inelegant and that the tolerance increase to keep all tests passing seems too large.

### DIFF
--- a/@unbndfun/sum.m
+++ b/@unbndfun/sum.m
@@ -9,7 +9,7 @@ function out = sum(g)
 % See http://www.chebfun.org/ for Chebfun information.
 
 % Get the domain.
-dom = g.domain.';
+dom = g.domain;
 
 % Cancel vanishing boundary values with negative exponents but save them
 % for later restoration:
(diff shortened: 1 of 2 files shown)


nickhale:
I made a few changes to this in this [here](https://github.com/chebfun/chebfun/compare/fix-singfun-exponents) (I don't have commit access to @davidagross's branch).

If theses seem OK, then we can probably merge. Perhaps @davidagross and @kuanxu could take quick a look.

davidagross:
Thanks @nickhale for these changes.  I made one additional change to yours re-aligning the dimensions of the exponents and domain.  Please check and make sure I did that in the right place (as opposed to transposing the domain and then transposing it back).

davidagross:
Also, bringing your changes @nickhale down to my branch and then pushing was very straightforward, but let me know if there's a better workflow to either give you access to my branch directly or to do future work like this with a different process.  Thanks!

PR #1362, opened 2015-01-27
Hotfix display non differentiable

asgeirbirkis:
Make sure that operators with non Frechet differentiable components get marked as nonlinear, rather than an error being thrown.


asgeirbirkis:
This should have been branched off master, another request coming up.

Issue #1339, opened 2015-01-05
Reverse coefficients in the help text of `chebcoeffs`


bhashemi:
The help text of `chebcoeffs` gives wrong information. It says that 
`A = CHEBCOEFFS(F)` means that
 F = A(1) T_{N-1}(x) + ... + A(N-1) T_1(x) + A(N) T_0(x)
which is not correct. It should be
 F = A(1) T_0(x) + A(2) T_1(x) + ... + A(N) T_{N-1}(x)

Here is an example showing this:
`>> f = chebfun('2 + x'); a = chebcoeffs(f)`
`a =`
`     2`
`     1`

PR #1370, opened 2015-01-30
Fix chebcoeffs

### LINKED ISSUE #1358: Unexpected behaviour from @SINGFUN/CHEBCOEFFS().

nickhale:
Closes #1358.

### DIFF
--- a/@chebfun/chebcoeffs.m
+++ b/@chebfun/chebcoeffs.m
@@ -1,12 +1,13 @@
 function out = chebcoeffs(f, varargin)
 %CHEBCOEFFS   Chebyshev polynomial coefficients of a CHEBFUN.
-%   A = CHEBCOEFFS(F, N) returns the first N Chebyshev coefficients of F ib
+%   A = CHEBCOEFFS(F, N) returns the first N Chebyshev coefficients of F is
 %   the column vector A such that F = A(1) T_0(x) +  A(2) T_1(x) + ... + 
 %   A(N) T_(N-1)(x), where T_M(x) denotes the M-th Chebyshev polynomial of the
 %   first kind.
 %
-%   If F is a smooth CHEBFUN (i.e., with no breakpoints), then CHEBCOEFFS(F) is
-%   equivalent to CHEBCOEFFS(F, LENGTH(F)).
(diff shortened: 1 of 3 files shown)


gradywright:
I looks like the documentation for `@chebtech/chebcoeffs' and `@trigtech/chebcoeffs` has the ordering of the coefficients incorrect.  My understanding is coefficients at the `chebtech` level are now stored in increasing order, not decreasing.

aaustin141:
@gradywright:  You're right; this was reported in #1339.  The documentation was also wrong in another place:  these methods return column vectors now instead of rows.  I've added a commit that fixes this, so closing this request will also close that issue.

PR #1374, opened 2015-02-02
Fix doublelength

nickhale:
I think this is the desired behaviour?


hrothgar:
I'm confused. Why would this be the desired behavior? Doublelength of a trigtech should double the degree, so N –> 2*N+1.

nickhale:
It's still doing that.

However I misunderstood and thought perhaps we didn't want to do double the degree for non-trigtechs.

I blame the confusing preference name, the lack of documentation, and the Mondays.

PR #1383, opened 2015-02-05
fix-dct-real

### LINKED ISSUE #1380: Ensure real output in CHEBFUN.DCT() et al.
We current make little attempt to ensure that the DCT/DST/IDCT/IDST of real vectors returns a real output. 

(The reason this is required is that we have to use the FFT to compute these, which can introduce small imaginary errors.)

For example:
```
>> chebfun.idst(ones(10, 1))
ans =
  1.264573231231541 - 0.000000000000000i
  0.000000000000000 - 0.000000000000000i
  0.398126284179942 - 0.000000000000000i
  0.000000000000000 - 0.000000000000000i
  0.209829367369638 + 0.000000000000000i
 -0.000000000000000 - 0.000000000000000i
  0.116847450394242 + 0.000000000000000i
  0.000000000000000 + 0.000000000000000i
  0.053386635079703 + 0.000000000000000i
  0.000000000000000 + 0.000000000000000i
```

nickhale:
Ensure real (imaginary) output for real (imaginary) input in DCT and DST

Closes #1380.

(This actually required very little effort, as kinds II-VIII are essentially implemented via kind I in both cases.)

### DIFF
--- a/@chebfun/dct.m
+++ b/@chebfun/dct.m
@@ -32,10 +32,11 @@ switch type
         
         % Equivalent to evaluating a ChebT expansion at 2nd kind points 
         % (up to a diagonal scaling). Implemented using the connection to 
-        % a real-even DFT, see CHEBTECH2.COEFFS2VALS().
-        
-        y = fft( [ u ; u(end-1:-1:2 ,: ) ]/2 ); 
-        y = y( 1:n, : );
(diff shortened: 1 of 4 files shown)


nickhale:
8542049 gives a significant performance boost when performing many transforms of the same length.

Issue #1382, opened 2015-02-04
Testing of chebfun outerproduct


asgeirbirkis:
I'm working on a feature branch restoring the behaviour of calling ```kron``` with three arguments, and getting a ```chebmatrix/linop``` output. Going through the current tests, I noticed that currently, there don't appear to be any test for ```chebfun/outerProduct```, or for the behaviour of calling ```f*g'```. There is a reminder in ```tests/chebfun/test_mtimes``` saying ```% TODO: Test Chebfun2 outerproducts!```, It's worth mentioning that ```test_kron``` also tries them indirectly.

I'm happy to write some as part of my pull request, however, it'd be nice if @ajt60gaibb could confirm that the tests I write are indeed testing the correct things.

ajt60gaibb:
I can take a look at the tests when you are ready.

asgeirbirkis:
Closed via #1385

PR #1385, opened 2015-02-05
Feature outer product operator

### LINKED ISSUE #180: Outer product operator
It's useful and conceptually interesting to use the outer product of two chebfuns to create a rank-1 operator (dyad). We have a traditional way, and now maybe a chebfun2 way...

asgeirbirkis:
Introducing capabilities for taking the outer product of two CHEBFUNs and getting an OPERATORBLOCK, rather than a CHEBFUN2.

Closes #180 (an issue that's precisely one year old today!) and #1382.

### DIFF
--- a/@blockFunction/blockFunction.m
+++ b/@blockFunction/blockFunction.m
@@ -118,9 +118,9 @@ classdef blockFunction
         end
 
         function F = mult(A, f)
-            %MULT
-            %
-            % Returns a BLOCKFUNCTION corresponding to a call to CHEBFUN/TIMES.
+        %MULT
+        %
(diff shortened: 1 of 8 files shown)


ajt60gaibb:
Though I am not going to merge the pull request, I have had look through the code and would support that button being pressed.  Looks like great code.  Thanks for strengthing the testing of the outerproduct Chebfun2 code too!  

Since Asgeir asked me, the linear differential operators from `ultraS` are not quite of the correct flavor for the `kron` command to make mathematical sense.  As it stands the domain of an `ultraS` operator is always in the Chebyshev T basis and I think this would need to change to more general ultraspherial bases.

ajt60gaibb:
I feel I do not have the authority to merge into development anymore.

PR #1386, opened 2015-02-07
Fix chebfun vertcat

### LINKED ISSUE #1234: error when creating a chebmatrix using a vector of doubles
When creating a chebmatrix, it is currently required that each scalar is inserted individually. There is no conceptual reason why concatenating with a vector shouldn’t work.
```
>> x = chebfun('x');
>> [x; 0; 0];         % works
>> [x; [0; 0]];       % breaks
Error using chebmatrix>parseData (line 743)
Entries in cell input to CHEBMATRIX constructor must be scalar.
Error in chebmatrix (line 91)
            data = parseData(data);
Error in chebfun/vertcat (line 50)
    out = chebmatrix(varargin.');
```

nickhale:
Closes #1234.


asgeirbirkis:
@hrothgar do you want to deal with this request since you filed the original issue?

Issue #1130, opened 2014-08-25
collection of examples of >3 slowdowns of v5 compared with v4


trefethen:
Except for codes like `conv`, `polyfit`, and `inv` that are benefitting from new algorithms, most computations take longer in Chebfun version 5 than they did in version 4.  We know that this is partly a result of v5 using more classes than v4 in its object-oriented design, but our understanding of the matter is thin.  In the upcoming year, we hope to find ways to recover some of this speed.

To help our investigation, I offer below an example of a code segment that takes 2.9s on my desktop in v4 and 10.4s in v5.  I would like to suggest that when others encounter examples of a slowdown by a factor of 3 or more, they add them as comments to this issue.  If we accumulate a collection of such examples, that will help us.  And of course, if readers of these examples examine them and spot good speedups to fix the problem, that will be marvelous!

```
tic, x=chebfun('x',[-2 1]); f=exp(.5*x).*(sin(5*x)+sin(101*x)); r=roots(f-round(f)); toc
```

PR #1403, opened 2015-02-19
feature-chebtech-lvalrval

### LINKED ISSUE #1171: Avoid using FEVAL() in @CHEBTECH/GET(f, 'RVAL').
It would be more efficient to simply sum the Chebyshev coefficients.

One could do something similar for `@chebtech/get(f, 'lval')`.

(Looking at profile reports from @hrothgar's v4-v5 comparisons (see #1130), often a lot of time was spent in `clenshaw_scl.m`. The fix suggested above might fix a lot of this.)

nickhale:
Add chebtech/lval() and chebtech/rval(). 

Closes #1171.

### DIFF
--- a/@chebtech/chebtech.m
+++ b/@chebtech/chebtech.m
@@ -281,6 +281,9 @@ classdef chebtech < smoothfun % (Abstract)
 
         % A 'loose' (i.e., not too strict) check for happiness.
         [ishappy, epslevel, cutoff] = looseCheck(f, values, pref)
+        
+        % Evaluate a CHEBTECH at -1.
+        out = lval(f)
 
         % Convert an array-valued CHEBTECH into an ARRAY of CHEBTECH objects.
         g = mat2cell(f, M, N)
@@ -356,6 +359,9 @@ classdef chebtech < smoothfun % (Abstract)
         
         % Round a CHEBTECH towards nearest integer.
         g = round(f)
+        
+        % Evaluate a CHEBTECH at +1.
+        out = rval(f)
 
         % Test an evaluation of the input OP against a CHEBTECH approx.
         pass = sampleTest(op, values, f)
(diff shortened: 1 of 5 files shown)


asgeirbirkis:
Looking at #1130, which inspired #1171 which in turn this pull request is aimed at, it would be nice if we had an automated way of comparing running times between different commits. Essentially something that would automatically generate the table that @hrothgar posted on #1130, and compare between runs. Thoughts @hrothgar or @aaustin141 ?

hrothgar:
It wouldn't be too hard to put together a script to do that, although it is slightly complicated by the necessary switching between branches, both because that (probably) has to be done manually and because all the variables need to be cleared when doing so.

Has anyone thought of adding V4 as a branch on this repo to make it easier to switch back and forth? Might be a good idea, too.

asgeirbirkis:
I think adding v4 to the repo is a good idea, not least for ensuring that the code doesn't get lost (not that it's a high risk event, but it'd be nice to have all chebfun code available at one place).

I think that the script doesn't have to be too complicated, and I'd be happy with the following scenario, at least for a start:

1. Run the timing script, output a log file with the times each method took.
2. Switch branches manually, clear all variables manually
3. Run the timing script again, but this time, with the output from (1) as an input
4. Output a log file with the comparision table from @hrothgar's comment in #1130 on 18/09.

To begin with, I don't think it's necessary to include v4 in the comparison, comparing between different versions of v5 would be fine. Essentially, a mechanism for checking whether changes we're doing to the development branch are speeding up or slowing down the code.

asgeirbirkis:
(Sorry @nickhale for hijacking this thread)

nickhale:
No problem - I think it's a good idea. 

I don't expect this particular pull request to have a massive effect, however.

hrothgar:
@nickhale, I've taken the liberty of putting your function `compare_branches` on Gist to keep track of it.

https://gist.github.com/hrothgar/51b10c867d3dcef0c026

asgeirbirkis:
I like this script, maybe we could discuss it on the meeting next week and if everyone is happy, get it into development?

PR #1409, opened 2015-02-20
increasing the tolerance for a few tests to pass in chebfun/test_mtimes

mhsnjvd:


hrothgar:
I was having the same issue. Thanks @mhsnjvd.

Issue #1392, opened 2015-02-13
Way to tell if a linop has cumsum operators in it


asgeirbirkis:
At the moment, once we start combining integral and differential operators, there is no way of telling that a linop has a cumsum operator in it. In other words, there is no integration analogue of ```diffOrder```, we use negative ```diffOrder``` to indicate that integration is involved, but that breaks down when we start combining operators. The following demonstrates the issue
```
>> C = operatorBlock.cumsum([-1 1]);
>> C.diffOrder
ans =
    -1
>> D = operatorBlock.diff([-1 1]);
>> D.diffOrder
ans =
     1
>> E = C + D;
>> E.diffOrder
ans =
     1
```
I know this is something @hrothgar was interested in last year, is that still the case?  Should we introduce ```cumsumOrder``` to the ```linBlock``` class?

asgeirbirkis:
I'd even stick a bug label to the following:
```
>> C = operatorBlock.cumsum([-1 1]);
>> I = operatorBlock.eye([-1 1]);
>> E = C+I;
>> E.diffOrder
ans =
     0
```

nickhale:
Is there a reason you want to be able to do this? Obviously `.diffOrder` is required for boundary conditions, but I'm not sure there's a need for a `.cumsumOrder`.

Furthermore, I don't think your example above is a bug at all. `.difforder` has always held the largest differential order in the operator, and here 0 > -1.

Probably the main problem here is that `.diffOrder` is barely documented at all:
```
    %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
    %% LINBLOCK CLASS PROPERTIES:
    %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
    properties ( Access = public )

        ...
 
        % Used to track differential order.
        diffOrder = 0;

        ...

    end
```

asgeirbirkis:
This came up when trying to fix #1209. To more carefully treat unknown parameters as opposed to functions, we should categorise any variable that's not acted on by differentiation or integration as parameter. The way we have for that now is to check whether the corresponding ```diffOrder``` is 0. However, when I then try
```
L = chebop(@(u) cumsum(u) + u)
```
we get ```L.diffOrder = 0```, which breaks the test.

[...]

Issue #1209, opened 2014-10-15
parameter-dependent problems where parameter appears in a boundary condition


hrothgar:
I am having no luck trying to solve a parameter-dependent problem in which the parameter appears in the boundary condition. Is there a way to get it to work, or is this feature not supported? I see no reason why it shouldn’t work.
```
op = @(x,u,p) diff(u,2) + u;
bc = @(x,u,p) [ u(0)
                u(1) - 2
                feval(diff(u),0) - p ];
A = chebop(op, [0 1], bc);
sol = A\0
```
I get the following error:
```
Error using cat
Dimensions of matrices being concatenated are not consistent.
Error in cell2mat (line 87)
        m = cat(2,m{:});
Error in chebDiscretization/matrix (line 63)
        M = cell2mat(A);
...
```
Passing an initial guess with correct dimensions does not help:
```
op = @(x,u,p) diff(u,2) + u;
bc = @(x,u,p) [ u(0)
                u(1) - 2
                feval(diff(u),0) - p ];

x = chebfun('x', [0 1])
init = [0*x; 0];

A = chebop(op, [0 1], bc, init);
sol = A\0
```

asgeirbirkis:
Here are some more lines of code related to this issue that came from the help mailing list:
```
%% Example 1 (fine)
L = chebop(@(t,x,p) diff(x)-1,[-1 1]);
L.lbc = @(x,p) x;
L.rbc = @(x,p) x-p;
X = L\0;
[x, p] = deal(X);
plot(x)
p
%Clearly, the solution is p=2 and x(t) is a straight line from (-1,0) to (1,2). 
% So far, Chebfun works fine. But now we try the same problem but using .bc syntax:
%% Example 2 (not working)
L = chebop(@(t,x,p) diff(x)-1,[-1 1]);
L.bc = @(t,x,p) [x(-1);x(1)-p];
X = L\0;
[x, p] = deal(X);
plot(x)
p
%Chebfun fails with a complicated error message, which I do not understand. However, if I slightly %change the problem so that the parameter p appears in the operator (replace -1 by -p in the first 
%line) and in turn removing the p from the bc (replace -p by -2 in the second line), then Chebfun 
%again works fine and gives the same x(t) and p=1 as expected:
%% Example 3 (fine)
L = chebop(@(t,x,p) diff(x)-p,[-1 1]);
L.bc = @(t,x,p) [x(-1);x(1)-2];
X = L\0;
[x, p] = deal(X);
[...]

PR #1418, opened 2015-03-02
Fix derivative dimensions

### LINKED ISSUE #1176: Dimension of derivatives
Something seems to be broken in deciding what dimensions derivatives should have for coupled systems. Here are few examples:

**Example 1:**
```
>> L = chebop(@(x,u,v) u + v);
>> LL = linop(L); LL.blocks
ans = 
    [0x0 operatorBlock]    [0x1 chebfun]
```
But we should expect both blocks to be ```chebfuns```.

**Example 2:**
```
>> L = chebop(@(x,u) u{1} + u{2});
>> LL = linop(L); LL.blocks
ans = 
    [0x0 operatorBlock]    [0x1 chebfun]
```
But we should expect both blocks to be ```chebfuns```.

**Example 3:**
```
>> L = chebop(@(x,u,v) diff(u) + v);
>> LL = linop(L); LL.blocks
ans = 
    [0x0 operatorBlock]    [0x1 chebfun]
```
This is actually correct.

**Example 4:**
```
>> L = chebop(@(x,u,v) u + diff(v));
>> LL = linop(L); LL.blocks
ans = 
    [0x0 operatorBlock]    [0x1 chebfun]
```
Here, the (1,1) block should have been a ```chebfun```, and the (1,2) block an ```operatorBlock```.

**Example 5:**
```
>> L = chebop(@(x,u,v) diff(u) + diff(v));
>> LL = linop(L); LL.blocks
ans = 
    [0x0 operatorBlock]    [0x1 chebfun]
```
Here, both blocks should have been ```operatorBlocks```.

asgeirbirkis:
A lot of changes to do more clever things with the dimensions of derivatives in Chebfun.

I'd like to go through this at a Chebfun meeting before merging, as it brings some subtle changes to chebops.

Closes #1176,  #1209 , #1392 (and potentially some more).

### DIFF
--- a/@adchebfun/adchebfun.m
+++ b/@adchebfun/adchebfun.m
@@ -1029,8 +1029,13 @@ classdef (InferiorClasses = {?chebfun}) adchebfun
             % F = PROD(F)       PROD of an ADCHEBFUN
             % 
             % See also chebfun/prod.
-            
-            f = exp(sum(log(f)));
+                        
+            % Temporary step to get dimensions correct
+            f = sum(log(f));
+            % Update FUNC part
+            f.func = exp(f.func);
+            % Update JACOBIAN part
+            f.jacobian = f.func*f.jacobian;
         end     
         
         function f = rdivide(f, g)
--- a/@adchebfun/adchebfun.m
+++ b/@adchebfun/adchebfun.m
@@ -1161,7 +1161,7 @@ classdef (InferiorClasses = {?chebfun}) adchebfun
 
                 % Convert the cell-array to a CHEBMATRIX and assign to the
                 % derivative field of U:
-                u.jacobian = linop( blocks );
+                u.jacobian = chebmatrix( blocks );
                 
                 % Initalise linearity information. The output is linear in all
                 % variables.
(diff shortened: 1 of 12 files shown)


hrothgar:
I've reviewed this code but am going to hold off for a day to merge it in case anyone wants to comment, since it does involve a somewhat significant change for ODEs.

To summarize, the main change this pull request introduces is the following. Until now, parameters (scalar unknowns) in ODEs were detected as the final variables in an op (e.g. in `@(x,u,p) diff(u)+p` the variable `p` is detected as a parameter), whereas with this change parameters are detected as _any_ variable that is not integrated or differentiated. (So with the change, the above example could also be written `@(x,p,u) diff(u)+p`.)

One concern I bring up is that a trivial example mentioned in #1209 still will not work with the changes in this pull request: see https://github.com/chebfun/chebfun/issues/1209#issuecomment-63653265. However, since the user can specify dimensions by passing an initial guess, I think it's fine that the above example doesn't work. What isn't fine is that passing an initial guess with correct dimensions still doesn't work! Is this a bug, @asgeirbirkis?

```
% Gives an error, as expected:
L = chebop(@(x,u,c) u, [-1 1], @(x,u,c) u(0)-c);
f = chebfun(@exp);
u = L \ [f; 0];

% Gives a different error, but should work:
L.init = [chebfun('x'); 0];
u = L \ [f; 0];
```

asgeirbirkis:
@hrothgar , could you please make a separate issue out of the trivial example from #1209 and your comment about it above?

PR #1420, opened 2015-03-04
fix jaccoeffs

nickhale:
Add more efficient version of jaccoeffs for chebtech-based chebfuns. 

Also fix a few errors in documentation for related methods.

### DIFF
--- a/@chebtech/jaccoeffs.m
+++ b/@chebtech/jaccoeffs.m
@@ -1,16 +1,18 @@
 function c = jaccoeffs(f, n, alp, bet)
-%JACCOEFFS   Compute Legendre series coefficients of a CHEBTECH object.
-%   B = JACCOEFFS(F) returns the Legendre series coefficients of CHEBTECH F, so
-%   that F = B(1)*P_0 + ... + B(N)*P_(N-1), where P_k is the kth Legendre
-%   polynomial. B is a vector of the same length as that of F.
+%JACCOEFFS   Compute Jacobi series coefficients of a CHEBTECH object.
+%   A = JACCOEFFS(F, N, ALPHA, BETA) returns the first N+1 coefficients in the
+%   Jacobi series expansion of the CHEBTECH F, so that such that F approximately
+%   equals A(1) J_N(x) + ... + A(N) J_1(x) + A(N+1) J_0(x) where J_N(x) denotes
+%   the N-th Jacobi polynomial with parameters ALPHA and BETA. A is a column
+%   vector. If length(F) < N then the additional entries of A are padded with
+%   zeros.
 %
-%   B = JACCOEFFS(F, N) returns the first N coefficients. If length(F) < N
-%   then the additional entries of B are padded with zeros.
-%
+%   A = JACCOEFFS(F, ALPHA, BETA) will assume that N = length(F) - 1;
+%   
 %   If F is an array-valued CHEBTECH, then a matrix of coefficients is returned
-%   so that F(:,k) = B(1,k)*P_0 + ... + B(N,k)*P_(N-1).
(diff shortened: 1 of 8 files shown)


ajt60gaibb:
I was interested to know what changed here, anything algorithmic? 
Using Clenshaw-Curtis-Jacobi? 

Could one do `cheb2jac` with `0<=alpha<=1` and `0<=beta<=1' using 
asymptotics and then use banded operators to do integer moves around the Jacobi 
plane?  This would cost `O(N (log N)^2/loglogN)' instead of `O(N^2)`. 
Useful for your fractional derivative stuff, perhaps? Worth it, maybe not?

nickhale:
Yes, the algorithmic change is just the switch to Clenshaw-Curtis-Jacobi. 

I thought the problem with the asymptotic based transforms in the Jacobi case is that we don't have a good error formula?

ajt60gaibb:
Oh, in the range `-1/2<alpha<1/2` and `-1/2<beta<1/2` the error is `2x` the first neglected term. See NIST, just under (18.15.1). 

When we were chatting before, I was speaking about general `alpha` and `beta`.

nickhale:
I have a working code for the US case, but not for general Jacobi. I can imagine it being a bit of a mess to write.. Feel free to have a try!

ajt60gaibb:
Yes, you do!  We need to find a cool application for that.  We can arbitrarily move up and down the diagonal of the Jacobi plane... hmmm...  And you are telling me it is not useful for fractional derivatives?  (Feels like it should be.)  

If you tell me it's worth it, then I will do it. I do not plan to do it as a sick idea of a fun few days (though it would be satisfying at the end). This is why Alex Barnett would tell us to construct the asymptotics numerically... he has a point with the applications he has in mind...

nickhale:
If you want to do fractional derivatives of high degree polynomials, then it could be useful.

The only algorithmic change in this pull request is that `cheb2jac()` is used when computing Jacobi coefficients of a chebtech-based chebfun, so please don't let the possibility of improving `cheb2jac()` hold up this pull request.

ajt60gaibb:
What @nickhale did is very good. It is what should have been done in the first place. I support the big green button being pressed.

asgeirbirkis:
I'm looking at this code and assuming no issues arise, will get it merged today or tomorrow.

@chebtech/jaccoeffs.m
```diff
@@ -0,0 +1,38 @@
...
+    bet = alp;
+    alp = n;
+    n = length(f);
+end
+
+% Convert from Chebyshev to Jacobi coefficients.
+c = cheb2jac(f.coeffs, alp, bet);
+
+% Truncate / pad to length n:
+if ( nargin > 1 )
```
asgeirbirkis: When would nargin not be greater than 1? We'd get errors above for alp and beta missing otherwise?

ajt60gaibb: The help text says that if `nargin=1`, then it computes Legendre coefficients. The code does not do this. 

I think it should return an error if `nargin<4`.

asgeirbirkis:
@nickhale , please look at the changes I made to the documentation in 3f44451. A lot of the jaccoeffs methods had help text that seemed to be more suited to legcoeffs methods. Please let me know if I was correct in fixing this, and have a look if I overlooked anything else that should have been changed.

Also, cf. the comment I made on the outdated diff above (and the comment from @ajt60gaibb on the issue).

nickhale:
Yes, sorry. Looks like I forgot to clear out a lot of the `legcoeffs()` stuff from the documentation.

The changes all look good to me -- thanks!

PR #1429, opened 2015-03-17
fix-chebfun-dst

nickhale:
DST-II and DST-III were the wrong way around. 

Fixed and improved the implementation.


ajt60gaibb:
Thanks.

PR #1432, opened 2015-03-20
Feature ivp splitting off

### LINKED ISSUE #1431: change default ODE IVP mode to "splitting off"
At present when you solve an ODE IVP, the Chebfun that is produced is constructed in splitting on mode with a `splitLength` of 300.  This has a very awkward consequence: if you take a couple of derivatives of the solution, little delta functions appear.  And these are precisely the functions a Chebfun user is most likely to want to take derivatives of!  This has come up in writing the ODE book.

I've discussed the matter with @asgeirbirkis, and so far as he is aware, there is no compelling reason to have "splitting on" as the default construction mode here.  (Historical comment: for a year or so way back when, Chebfun *function* construction had splitting on by default, but the lack of global smoothness drove us crazy and we turned it off.)  We accordingly propose to make the minor changes in the code necessary to change the IVP solver construction default to splitting off.

If nobody objects, @asgeirbirkis will proceed with a branch that does this, testing to make sure there are no untoward consequences, and hopefully before long we will have newly smooth outputs from our IVPs.

Comments anyone?

asgeirbirkis:
Closes #1431

### DIFF
--- a/@chebfun/odesol.m
+++ b/@chebfun/odesol.m
@@ -65,9 +65,10 @@ end
 %% Create a CHEBFUN object.
 p = chebfunpref();
 p.techPrefs.eps = max(relTol); % Use the same tolerance for each column.
-p.splitting = true;            % use splitting, always, or there is no hope.
-p.splitPrefs.splitLength = 300;
-p.splitPrefs.splitMaxLength = p.techPrefs.maxLength;
+p.splitting = false;
+% p.splitting = true;            % use splitting, always, or there is no hope.
+% p.splitPrefs.splitLength = 300;
+% p.splitPrefs.splitMaxLength = p.techPrefs.maxLength;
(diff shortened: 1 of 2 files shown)


trefethen:
I understand from talking with @asgeirbirkis that before this branch, the splitLength 300 was actually hardwired into our code! -- i.e., inaccessible to various preferences.  This branch not only sets the default to splitting off, but it also regularizes that situation: now, when you specify splitting on for an IVP solution, the splitLength will be the standard one (160 by default, overrideable).  I am delighted with these changes, and I've checked an example from the ODE book happily and successfully.  @hrothgar, maybe you would like to review this?

asgeirbirkis:
Thanks @hrothgar for looking at these recent pull requests!

trefethen:
Thanks from me too @hrothgar.

PR #1435, opened 2015-03-23
Feature chebop vectorize

### LINKED ISSUE #1430: Automatic vectorizing of the `op` field of a chebop
For the ODE book whose preparation is getting underway, I am obsessive about making all the syntax as simple as possible; this must be a book about ODEs, not Chebfun.  Indeed, this obsession is what led to the new IVP capabilites that @asgeirbirkis has worked on so hard in the past year.

I think it would be extremely nice if, in defining in-line `op` fields for chebops, one could write things in format like 
```
N.op = @(x,u) diff(u,2) + a*u^3
```
rather than
```
N.op = @(x,u) diff(u,2) + a*u.^3
```
Note that chebgui already has this capability.
 
I've talked with @asgeirbirkis and he has already been making progress on a method for achieving this.  (It's not trivial because of the proper treatment of a parameter like the constant `a` in the example above.)  If nobody objects, he and I will continue to devleop and test this -- but it seems important to file this issue because this is a significant user-level item.  Comments?

asgeirbirkis:
Closes #1430.

By default, CHEBOPs will try to vectorize anonymous functions passed to them. It is possible to turn off this feature, either globablly via cheboppref or by setting the ```vectorize``` property of CHEBOPs. This is documented in the help text for CHEBOPs.

### DIFF
--- a/@chebop/chebop.m
+++ b/@chebop/chebop.m
@@ -126,7 +126,7 @@ classdef (InferiorClasses = {?double}) chebop
 % 
 % Example (solving an IVP by automatically converting it to first order form):
 %
-% % Solve the van der Pol equation u'' - 25*(1-u^2)u' + u = 0, u(0)=2, u'(0)=0
+% % Solve the van der Pol equation u'' - 20*(1-u^2)u' + u = 0, u(0)=2, u'(0)=0
 % vdpFun = @(u) diff(u, 2) - 20*(1-u.^2).*diff(u) + u;
 % dom = [0 100];
 % N = chebop(vdpFun, dom);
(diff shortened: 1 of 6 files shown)


aaustin141:
I'm not saying it's not a good/clever idea, but we should at least be forewarned that the `functions` function called [here](https://github.com/chebfun/chebfun/blob/f28709a3424fa10a9ebb50b96cd8ac7526d9c6ae/%40chebop/vectorizeOp.m#L31) is marked by MATLAB as an internally-facing API and isn't meant to be used by outsiders.  In particular, MathWorks reserves the right to change its behavior without warning in future releases of MATLAB; see `help functions`.

asgeirbirkis:
Thanks @aaustin141 for pointing this out. I first used this functionality of ```functions``` at around 2009, and it hasn't changed since then. Furthermore, according to the Matlab documentation, http://uk.mathworks.com/help/matlab/ref/functions.html this has been in Matlab since prior to version 2006a , so I think we're on the safe side to make use of it!

PR #1438, opened 2015-03-26
Making isMult a property of linBlock, rather than operatorBlock.

asgeirbirkis:
This means that functionalBlock now has this property.

Fixes a crashing chebtest (chebop/test_autoVectorize).


asgeirbirkis:
@aaustin141 I renamed the property and have now added tests, hopefully this addresses the concerns you raised in your review.

Issue #1440, opened 2015-04-01
Trigfun `cumsum`


jaurentz:
In the current version of Chebfun the trigtech/cumsum command throws an error if the function
has a nonzero mean. The line between zero and nonzero mean can be difficult to determine. 
Consider the following example:

``` matlab
 >> f = chebfun([1;0;1],[-1,1],'coeffs','trig'); cumsum(f)    

ans =

   chebfun column (1 smooth piece)
       interval       length   endpoint values trig
[      -1,       1]        3  -1.6e-16        0 
Epslevel = 2.220446e-16.  Vscale = 5.513289e-01.

>> g = chebfun([1;1e-14;1],[-1,1],'coeffs','trig'); cumsum(g)
Error using trigtech/cumsum>cumsumContinuousDim (line 84)
Indefinite integrals are only possible for TRIGTECH objects with zero mean.

Error in trigtech/cumsum (line 64)
    f = cumsumContinuousDim(f, m);

Error in bndfun/cumsum (line 29)
    g = cumsum(f.onefun);


Error in chebfun/cumsum>cumsumContinousDim (line 85)
        newFuns = cumsum(f.funs{j});

Error in chebfun/cumsum (line 51)
        f(k) = cumsumContinousDim(f(k), m);
```

Even in this simple example it is not clear what to do. One option might be to
return a chebfun instead of an error.

trefethen:
I've been talking with @jaurentz to understand his concerns here (which were prompted by a test that failed in the branch `feature-prescribed-eps`, namely `trigtech/test_cumsum`).  It seems there are three concerns:

(A) A user encountering the message above may be bewildered.

To address this, I suggest the following.  First of all, `@chebfun/cumsum` currently does not mention that zero mean is needed for `cumsum` applied to a trigfun.  Such a comment must be added.  (There is such a comment in `@trigtech/cumsum`, but users cannot be expected to look there.)

Second of all, the error message above could say more: e.g., perhaps

```
Error using trigtech/cumsum>cumsumContinuousDim (line 84)
Indefinite integrals are only possible for TRIGTECH objects with zero mean: see help chebfun/cumsum
```

(B) The logic of how close to zero counts as a "zero mean" may not be optimal.  (In the test mentioned above, the tolerance may be too strict.)

(C) If the mean is not zero, should we return a chebfun rather than just an error message?

mhsnjvd:
[...]

PR #1442, opened 2015-04-02
making the zero mean test at the Chebfun level tighter

mhsnjvd:
This addresses #1440


mhsnjvd:
@jaurentz  How about now?

Issue #1445, opened 2015-04-07
display `vscale` and `epslevel` rather than `Vscale` and `Epslevel` 


trefethen:
In Chebfun, our convention is that the correct spelling of `epslevel`, `vscale`, and `hscale` is all lower case.  However, when we display  a chebfun, they come out with initial upper-case:
```
>> x = chebfun('x')
x =
   chebfun column (1 smooth piece)
       interval       length   endpoint values  
[      -1,       1]        2        -1        1 
Epslevel = 1.110223e-15.  Vscale = 1.
```
This happens though even if you try to access such variables, it won't work:
```
>> x.Epslevel
Error using chebfun/get (line 180)
'Epslevel' is not a recognized CHEBFUN property name.
Error in chebfun/subsref (line 144)
        out = get(f, idx); 
>> x.Vscale
Error using chebfun/get (line 180)
'Vscale' is not a recognized CHEBFUN property name.
Error in chebfun/subsref (line 144)
        out = get(f, idx); 
```
I think the output should be simply `epslevel` and `vscale`.  I've taken the liberty of assigning this issue to our style-wallah, @mhsnjvd.  If nobody objects, Mohsin, would you be willing to make this change?

Hadrien-Montanelli:
In that case, would it be possible to get rid of the full stop after the displayed value? 

Hadrien

trefethen:
Yes please!  I had not dared to ask for so much :) .

aaustin141:
Closed in #1446.

PR #1446, opened 2015-04-07
some minor display changes.

mhsnjvd:
This addresses #1445. This is what things look now:
```
>> x
x =
[      -1,       1]        2        -1        1 
epslevel = 1.110223e-15   vscale = 1
>> h = heaviside(x)
h =
[      -1,       0]        1         0        0 
[       0,       1]        1         1        1 
epslevel = 2.220446e-16   vscale = 1   Total length = 2
```

### DIFF
--- a/@chebfun/disp.m
+++ b/@chebfun/disp.m
@@ -76,7 +76,7 @@ end
 [extraItem, extraData] = dispData(f);
 
 % Loop through each of the funs to display the following information:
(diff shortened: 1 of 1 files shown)


trefethen:
I might add that it is likely that before too long `epslevel` will no longer appear, as is already the case in `v5b` mode of `feature-prescribed-eps`.

PR #1447, opened 2015-04-07
Fix ICs diff-orders

### LINKED ISSUE #1444: bug in ODE interpretation of BCs
In writing an example in the ODE book I found that the following code produces two identical plots, whereas they should be different:
```
L = chebop(0,50);
L.op = @(t,y) diff(y,2) + y;
subplot(2,1,1), L.lbc = @(y) [y-1; diff(y)]; plot(L\0)
subplot(2,1,2), L.lbc = @(y) [y-1; diff(y,2)]; plot(L\0)
```
Somehow or other, chebfun interprets `diff(y,2)` (also `diff(y,77)`) as the same as `diff(y)`.

Ideally, Chebfun should solve the problem stated.  If that's not feasible, one would hope it would at least detect that `diff(y,2)` has been written and generate an error message.  @asgeirbirkis, can you take a look?

asgeirbirkis:
Closes #1444.

We now throw meaningful error messages when unsupported initial conditions are specified. Examples:

Specifying the value of a derivative of a too high order
```
L = chebop(0,50);
L.op = @(t,y) diff(y,2) + y;
L.lbc = @(y) [y-1; diff(y,2)]; y = L\0
Error using treeVar.sortConditions (line 103)
The value of a derivative of a too high order was specified.
Error in chebop/solveivp (line 212)
    idx = treeVar.sortConditions(N.lbc, N.domain, diffOrders);
Error in  \  (line 40)
    [varargout{1:nargout}] = solveivp(N, rhs, pref, varargin{:});
```

Specifying multiple values for the same derivative:
```
>> L.lbc = @(y) [y-1; y-2]; y = L\0
Error using treeVar.sortConditions (line 109)
Multiple initial conditions on the same variable/derivative specified.
Error in chebop/solveivp (line 212)
    idx = treeVar.sortConditions(N.lbc, N.domain, diffOrders);
Error in  \  (line 40)
    [varargout{1:nargout}] = solveivp(N, rhs, pref, varargin{:}); 
```

Updated tests to check that we're throwing the correct errors.


nickhale:
(Without thinking about it too carefully) I don't see why the following boundary conditions wouldn't be valid:
```
>> L = chebop(0,50);
>> L.op = @(t,y) diff(y,2) + y;
>> L.lbc = @(y) [diff(y)-y; diff(y)+2*y];
Error using treeVar.sortConditions (line 71)
Initial/final condition not supported. Please ensure that the
unknown function(s) appear only once in each condition,
e.g. not as u + u' - 1, and that the coefficient of the unknown
function is 1, e.g. not 5*u-1.
Error in chebop/solveivp (line 212)
    idx = treeVar.sortConditions(N.lbc, N.domain, diffOrders);
Error in  \  (line 40)
    [varargout{1:nargout}] = solveivp(N, rhs, pref, varargin{:}); 
```

asgeirbirkis:
I agree that the example you post above makes sense, and in collocation mode, it works:
```
L = chebop(0,50);
L.op = @(t,y) diff(y,2) + y;
L.lbc = @(y) [diff(y)-y; diff(y)+2*y];
cheboppref.setDefaults('ivpSolver','collocation');
u = L\1;
lbcRes = L.lbc(u);
feval(chebfun(lbcRes), 0)
ans =
   1.0e-12 *
    0.1082   -0.0525
```

It's to do with the fact that by default, we call ode113, which requires the following calling syntax:
```
[TOUT,YOUT] = ode113(ODEFUN,TSPAN,Y0)  with TSPAN = [T0 TFINAL] integrates 
    the system of differential equations y' = f(t,y) from time T0 to TFINAL 
    with initial conditions Y0.
```
I.e. `Y0` is a vector of initial conditions, where in your example above, the first element prescribes u(0), the second element u'(0), and your conditions can't be written on that form. However, the error message should probably point users to collocation mode, I'll change that now.

asgeirbirkis:
With 27b3dd4, the error message now reads
```
L = chebop(0,50);
L.op = @(t,y) diff(y,2) + y;
L.lbc = @(y) [diff(y)-y; diff(y)+2*y];
u = L\1;
Error using treeVar.sortConditions (line 71)
Initial/final condition not supported. Please ensure that the
unknown function(s) appear only once in each condition,
e.g. not as u + u' - 1, and that the coefficient of the unknown
function is 1, e.g. not 5*u-1. Alternatively, try solving the
problem in collocation mode (see help cheboppref for details).
Error in chebop/solveivp (line 218)
    idx = treeVar.sortConditions(N.lbc, N.domain, diffOrders);
Error in  \  (line 49)
    [varargout{1:nargout}] = solveivp(N, rhs, pref, varargin{:}); 
```

PR #1458, opened 2015-04-20
Add ability to compute the transpose of the leg2cheb operator.

nickhale:
This is useful in computation of the inverse discrete Legendre transform, which we hope to add soon.

This commit also addresses some small bug fixes left over from the switch in order of coefficients in Chebfun, as well as some general tidying up.

Tests have been added for the new features.

### DIFF
--- a/cheb2leg.m
+++ b/cheb2leg.m
@@ -1,4 +1,4 @@
-function c_leg = cheb2leg(c_cheb, normalize, M)
+function c_leg = cheb2leg(c_cheb, varargin)
 %CHEB2LEG   Convert Chebyshev coefficients to Legendre coefficients. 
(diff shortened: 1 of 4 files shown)


ajt60gaibb:
Before this added, it should be checked that transposed leg2cheb is not related to cheb2leg via a discrete orthogonality relation.  The connection is not obvious because of the doubling up that has to be done inside `cheb2leg`, but transposes are so often diagonally-scaled inverses in this proportion of the universe.

nickhale:
@ajt60gaibb and I have now agreed that it is _not_ possible to compute a transpose `leg2cheb` via a suitably scaled `cheb2leg`.

ajt60gaibb:
Yes, I agree. It is not possible because of the doubling up.

Issue #1460, opened 2015-04-20
Should the periodic flag give an error when given an even number of samples?


ajt60gaibb:
So, I just found out that trigtech only works with odd length samples. Fine, I understand that parity is annoying here.  But, I do not get an error if I try: 
```
f = chebfun( ones(32,1), 'periodic'  );
g = cumsum(f);
```
Now, `g` is the zero function! I was expecting an error or for it to be cast to a chebfun (with a warning?).

gradywright:
trigtechs should work completely fine for even length samples.  Is there somewhere in the trigtech code that would lead you to believe otherwise @ajt60gaibb?

The bug you found is a bug with @chebfun/cumsum.  It is incorrectly checking the mean of an even length trigfun (another bug from when we switched the order of the coefficients).  I will have a fix (and test) soon.

ajt60gaibb:
Oh, `trigtech/cumsum` made me think that because of the way it was checking the coefficients.

ajt60gaibb:
Also, everytime I tried to make a `trigfun` from a function handle it came out with an odd number of coefficients. This seems to happen also when an even number would do.

ajt60gaibb:
Also, things like this: 
```
f = trigtech(@(x) exp(1i*pi*x) + exp(-2i*pi*x)); 
f.coeffs
ans =

   1.0000 - 0.0000i
  -0.0000 - 0.0000i
   0.0000 + 0.0000i
   1.0000 - 0.0000i
   0.0000 - 0.0000i
```

gradywright:
Unless you explicitly construct from samples, a `trigfun` will always be odd.  I'm pretty sure more people than just me know this and that we decided this would be fine since even length trigfuns are odd (pun-intended)--they break symmetry between numbers of cosine/sine modes.   Do others have comments?

trefethen:
About even vs odd, yes, trigfuns are supposed to have odd length unless you explicitly force it to be even.  This is discussed carefully in our paper on the subject, `https://people.maths.ox.ac.uk/trefethen/trigpaper.pdf`; see in particular remark 3 on p. 2 and the text on p. 10.  Speaking philosophically, let me comment that whereas of course in principle everything about even vs. odd trigonometric expressions is utterly elementary, in practice this is tricky subtle stuff and our paper is possibly the only place on earth that really discusses it carefully -- ultimately because Chebfun in unusual in being forced to confront such matters.

mhsnjvd:
[...]

PR #1461, opened 2015-04-20
Fix cumsum of an even trigfun (and trigtech)

gradywright:
Fixed bug where cumsum of an even trigfun (and trigtech) was not working properly.
Added tests in chebfun/cumsum for cumsums of trigfuns (nothing was there previously).

Partially address #1460.


ajt60gaibb:
Close #1460.

PR #1467, opened 2015-04-23
Feature dlt

nickhale:
A fast implementation of the discrete Legendre transform (DLT) and its inverse (IDLT).

The DLT is analogous to the `@chebtech2/coeffs2vals()` method but evaluates a Legendre series at Legendre nodes (rather than a Chebyshev at Chebyshev nodes). The IDLT is analogous to `@chebtech2/vals2coeffs()`.

The algorithm combines the existing `leg2cheb()` with a 'non-uniform discrete Cosine transform' (NDCT), which evaluates a Chebyshev series at a Legendre grid. 

The details are contained in a forthcoming paper by Hale (@nickhale) and Townsend (@ajt60gaibb) . I'll add a link once a pre-print is ready.

Note 1: This branch is based off of the `feature-leg2cheb-transpose` branch.
Note 2: Tests are included for both the DLT and IDLT.


ajt60gaibb:
I hope the Chebfun team merge the `dlt` and `idlt` commands, though at the moment we do not have a big Chebfun application.   I will remind you that a similar situation was true when `cheb2leg` and `leg2cheb` went into Chebfun.  A year later, the commands `polyfit` and `conv` were using them...  And this would have not happened without a `legpts` command back in 2011 (which was a standalone command back then). 

In a year I could see Chebfun using `dlt` and `idlt` for fractional derivatives and integro-differential equations, perhaps. Hard to predict.

aaustin141:
I've now taken a brief look at this.  I'm not in any way qualified to assess the mathematics (and especially not without the preprint), but I've played with it some, and it seems to work!  The tests for the new stuff exercise everything fully, and I don't find that it causes anything to break (naturally, since nothing uses this yet).

There are one or two coding style things that need fixing, but I can take care of those.

Does anyone object to merging this?

trefethen:
Yes, I object! -- but only in the following very upbeat sense.  When something like this gets into Chebfun that could be interesting to users, we need ways for the team to know about it and then eventually users.  So might I ask, before we merge, if @nickhale would be willing to write an Example showing off this functionality?  Looking at such an example is a pretty healthy part of a review process, and it would give us something to go over at the next Chebfun team meeting 13 days from today.  Of course if anyone wishes to propose a comment or to that could go into the Guide too, that would be great.

ajt60gaibb:
I just uploaded a Chebfun example of `dlt` and `idlt` here: 
https://github.com/chebfun/examples/blob/add-dltExample/cheb/fastdlt.m

trefethen:
Thank you very much @ajt60gaibb!  I am not sure how quickly I can look at this just now since I'm traveling, but @aaustin141, if you can check out the example to make sure it doesn't raise any question marks, then by all means go ahead and merge.  I will be sure to find time to look at the Example at some point in the upcoming week.

nickhale:
@aaustin141 : Thanks. Feel free to make any coding style changes.

@ajt60gaibb : I'm not entirely sure the `ndct()` functionality needs to be separated out. If it is, then the help text needs to be updated to reflect the fact that it's a static method.

nickhale:
I moved @ajt60gaibb 's Example to here:
https://github.com/chebfun/examples/blob/add-dltExample/cheb/FastDLT.m

PR #1471, opened 2015-04-27
Abstract chebfun2

ajt60gaibb:
This pull request adds an abstract class called `lowrankapprox`. The only subclass at the moment is `chebfun2`. The idea is to extract out as much functionality as possible so that the 2D side of Chebfun can work more seamlessly with periodic functions, functions on the sphere, functions on the disc, etc.  The lowrankapprox requires that a 2D function is written as (columns)x(diagonal)x(rows) in standard CDR form, where the columns and rows are techs. 

This is what was intended during the rewrite of version 5, except we never got around to it. 

Users will essentially be unaware of the changes (except when reading some of the help text).


trefethen:
I have no direct opinion on this apart from the suspicion that it's probably a good idea if you think it's a good idea.  However, I have a meta-concern.  There's not much development happing in 2D Chebfun these days, yet the team needs some awareness of what is going on, especially as @bhashemi is beginning 3D experiments.  So with a fundamental design change like this, I think it's highly desirable to spread awareness among our team of what is going on so we can have a substantive a discussion as possible.  Any thoughts?  Might you e.g. want to connect to our Oxford team meeting Wednesday nine days from now?

ajt60gaibb:
I issued a pull request because I have something working and I would like to open up the discussion. I am not expecting this to be merged immediately because it is a design change. 

I am happy to join the Chebfun meeting 9 days from now.  I will start talking to members of the Chebfun team because this is the first time I have used abstract classes.   

@gradywright, his student Heather Wilber, and I are working on structure-preserving low rank function approximations on the sphere and disc.  By having a `lowrankapprox` class we do not have to start again on the implementation and we get lots of the uninteresting algorithms for free.  That is the motivation for this and it will test the design structure.

ajt60gaibb:
OK so `lowrankapprox` has changed to `separableApprox`.  I believe we are happy with the general idea behind this design change. I hope it will be merged into development soon.  All tests pass for me.

gradywright:
I have finished my initial review of the new `@separableApprox` class.  Below are some comments/suggestion/corrections regarding the new code (in no particular order):

1.  Should `chebpolyplot` and `chebpolyplot2` be in the `separableApprox` class?  These are deprecated methods and thus it seem to me that they can just go under the `@chebfun2` class.

2.  I'm not fond of the `diffx` and `diffy` methods being in `separableApprox` class.  The names of these methods do not make sense in all contexts. 

3.  Should discriminant and `hessian` be abstract `methods`?  They will be heavily dependent on topology of the underlying domain of the `separableApprox`.  For example, the disk or sphere would have different implementations of these methods.

4.  Same goes for the `fred` and `volt` methods.  Should these be moved to `chebfun2`?

5. I don't think `ellipsoid` should be allowed for anything that is of `separableApprox` type.  The current code is very specific to `chebfun2` objects.  Also, `sphere` was moved to be a strictly `chebfun2` function.

6.  Need to change the help text in `separableApprox/integral2`:  "This is a very slow feature, and is only fast for small degree bivariate polynomials."

7.  The implementation in `separableApprox/integral2` for integration along a complex-valued path is very specific to chebfun2.

8.  The help text of `separableApprox/length` makes references to polynomial degree.  This should be changed since `separableApprox` should not know what the underlying discretization is.

9.  `separableApprox/max`, `separableApprox/median` use `chebpolyval2`.  They should instead call an equivalent method of the @separableApprox class.

10.  The same is true for `separableApprox/vscale`.  This needs to be fixed as this code is called a lot.

11. `separableApprox/minandmax2` has code that is very specific to the underlying tech being a `chebtech` (see lines 84-105).  It would be nice if there was a  `getPoints` method for techs...

12.  Should we only allow operations like `mtimes`, `plus`, `power`, `rdivide`, and `times` to be done when the input arguments are of the same child class?  Otherwise how can we handle these operations when `f` is a `chebfun2` and `g` is a `trigfun2` (or something else)?

13.  We need to figure out a proper implementation for `separableApprox/plotcoeffs2`.  If there was a `chebfun` method for getting the coefficients of whatever tech is being used, then we could easily make `plotcoeffs2` general.  I'm not sure what a good workaround is.

14. In `separableApprox/plus` the method `compression_plus(f, g)` is used.  Is this a general technique for anything that is a `separableApprox`?

15.  It think `pol2cart`, `cart2pol`, `sph2cart`, and `cart2sph` should be part of `chebfun2`, not `separableApprox`.

16.  The method `separableApprox/power` always returns a `chebfun2` object.  Is this the desired behavior?

17.  I think `separableApprox/restrict` should be an abstract method since restriction to a subdomain is very specific to the type of tech and the domain being used.

18. `separableApprox/std` explicitly constructs `chebfun2` objects to do the computations.  Can this be made more general?

19.  Several method of `separableApprox` return a `chebfun` as output (e.g. `sum`, `diag`, `volt`, `fred`).  However, the result often does not respect the tech type used in the rows/cols of the `separableApprox` representation.  

20.  Lines 44-45 of `separableApprox/svd` explicitly construct a `chebfun2` object in the case that the function to compute the SVD of is exactly zero.  Can this be made more general so that the underlying child of `separableApprox` is constructed?

21.  **The remainder of the comments/suggestion/corrections are in regards to `chebfun2v`**.  What are we going to do about `chebfun2v`?  Can we also make an abstract class `separableApproxv`?  The answer to these questions may take care of the ones remaining below.

22.  How do we handle line 124 of `separableApprox/feval` when the input is a chebfun2v object?

23.  `grad` and `gradient` return chebfun2v objects.  This seems suggests that we should move them to `chebfun2`.  Additionally, these functions will depend heavily on the topology of the underlying `separableApprox` domain.

24. `separableApprox/jacobian` is not working when the inputs are not empty.

25.  Need to figure out a general way for `separableApprox/mtimes` to handle the case where the input is a chebfun2v object.

26. `separableApprox/quiver`, `separableApprox/quiver3.m`, and `separableApprox/roots` make heavy use of `chebfun2v`, again suggesting a abstract separableApproxv class.

27.  The help text and code in `separableApprox/subsref` make explicit use of `chebfun2v`.

28.  `separableApprox/vertcat` explicitly constructs chebfun2v objects.  Can this be made more general?

gradywright:
29:  Should we add `display` as an abstract method of `separableApprox`?

ajt60gaibb:
I did the easy things above:   

1. OK. Done. Thanks. 
2. I have no real opinion on this because they do not always make sense in the chebfun2 setting either. I have moved them to the `@chebfun2` class. 
4. Moved. 
5. Er. OK. Moved. 
6. Updated. Thanks. 
8. Done. 
15. Moved `pol2cart`. We do not have the other command, see #1541.
16. Done. Thanks. 

Left is: 3, 7, 9, 10, 11, 12, 13, 14, 17-29.   Though, 21-28 should probably be left for a later pull request.

gradywright:
I am finally back to reviewing the code and have made some changes to address the issues raised in my first review that @ajt60gaibb did not get to.  Issues 3, 9, 10, 12, 13, 20, 24, and 29 are now taken care of.  

After some contemplation (and time spent implementing a new subclass of `abstractChebfun2`), I don't think that issues 7, 17, 18, and 19 are important.  These can always be overridden by the subclass and the implementations of the functions mentioned in these issues will still work as defaults.

The remaining issues are 21-28, which all deal with handling vector valued functions, i.e. `chebfun2v`.  As @ajt60gaibb suggested, I think these should be addressed in a separate pull request.  I will post them to the tracker so we do not forget.

Before I merge the pull request, I do have a question.  Should there be tests specifically for the `abstractChebfun2` class?  Right now, the `chebfun2` test suite touches all these function, except the ones pertaining to plotting.

ajt60gaibb:
@gradywright: This is great!  Nice.  It will make `spherefun` and `diskfun` so much neater. 

In this pull request, the only concrete class beneath `separableApprox` is `chebfun2`. Therefore, I believe it is fine to keep the tests unchanged.  Once we put in `spherefun` and `diskfun`, then we may want to change the structure of the tests. I am not sure yet.   I hope you think this is reasonable.

PR #1474, opened 2015-05-04
Feature prescribed eps

jaurentz:
This pull request effectively disables the use of epslevels, in the sense that a chebfun's epslevel is enforced to be Matlab eps. The bulk of the changes in this branch correspond to loosening tolerances in tests. The rest of the changes involve loosening tolerances in subroutines that make decisions based on epslevels.


asgeirbirkis:
I've had a look at this request, it looks pretty straightforward to me and simply does what we've getting descriptions of at meetings. I have couple of questions for @jaurentz and @trefethen before proceeding with the pull request:

1. In a lot of places in the code, you're renaming `epsLevel` to `epslevel`. Could you please remind me whether that's something that's been brought up that we should do, i.e., why are we abandoning camel-casing here? Perhaps @mhsnjvd can help freshen my memory?

2. In a lot of places in the code, you're increasing tolerance by one order of magnitude, e.g. line 13 of `@chebfun/thresholdBreakpointValues.m`:
```
-    breakVals(abs(breakVals) < vscale(f(k))*epslevel(f(k))) = 0;
+    breakVals(abs(breakVals) < 10*vscale(f(k))*epslevel(f(k))) = 0;
```
More dramatically, here's a chance from `tests/bdnfun/test_diff` (line 94);
```
-pass(10) = err < 100*tol;
+pass(10) = err < 1e6*tol;
```
I presume this is all because previously, the epslevels of chebfuns (and bndfuns and chebtechs) were several orders of magnitude larger than machine eps, and now that everything is at machine eps, the tolerance simply has to be scaled accordingly. However, it'd be nice to get a concise demonstration that the error has not increased by making the change of throwing away epslevel, in other words, that `100*tol` on the current development is actually the same number as `1e6*tol` on this new branch.

jaurentz:
With regards to 1), I must admit I am not entirely clear on what the standards are. If there is a clear 
rule I am happy to make the necessary changes. 

With regards to 2), this particular function had bizarre behavior only on my laptop. In fact it missed the
tolerance exactly in the last bit! I added the factor of 10 to avoid such a tight decision. To me, this 
problem is more of a practical nature than a theoretical one. Adding the factor of 10 seems just as 
reasonable as a factor of 5, in the sense that in practice there is no discernible difference. If there is
a good reason not to do this I would be happy to hear it. 

With regards to the bndfun example, this particular test takes the 4th derivative of a Chebfun approximation to sin(x). If we perform this experiment outside the test we get the following:
```matlab
>> f = chebfun('sin(x)'); g = diff(f,4); chebcoeffs([f,g])

ans =

            0            0
   8.8010e-01   8.8010e-01
            0            0
  -3.9127e-02  -3.9127e-02
            0            0
   4.9952e-04   4.9952e-04
            0            0
  -3.0047e-06  -3.0045e-06
            0            0
   1.0499e-08   1.0581e-08
            0            0
  -2.3960e-11            0
            0            0
   3.8538e-14            0
```
Here we see that the coefficients of f and g differ by roughly 6 orders of magnitude, which is exactly the size of the fudge-factor in the tolerance. Based on this experiment, I feel that this tolerance is
acceptable.

trefethen:
Thanks @jaurentz.  About (1), i.e. `epslevel` vs. `epsLevel`, this is a long-established exception (also `hscale` and `vscale`).  For those with access to the `discussion_documents` folder in the dropbox at Oxford, you can find a thorough discussion in item 6 of `21-epslevel-progress.pdf`.

asgeirbirkis:
@jaurentz: OK, let me first recap what I believe are the changes you're making with this request:

a) Changing `epsLevel` to `epslevel`. As @trefethen points out below, this is one of our exceptions for camel casing, so I'm fine with this.

b) Changing a lot of lines regarding checking error/sanity checking in chebtests.

c) You're only making few changes in the actual Chebfun code. The changes include:
In `@chebfun/thresholdBreakpointValues.m`:
```
-    breakVals(abs(breakVals) < vscale(f(k))*epslevel(f(k))) = 0;
+    breakVals(abs(breakVals) < 10*vscale(f(k))*epslevel(f(k))) = 0;
```
In  `@chebtech/isdecay.m`:
```
-    mask = abs(get(g, 'lval')) < 1e3*tol;
+    mask = abs(get(g, 'lval')) < 1e4*tol; 
+    % tolerance loosened in epslevel-dependent code
```
In `@singfun/plus.m:
```
-tolSmth = 1e2*eps;
+tolSmth = 1e3*eps; % tolerance loosened in epslevel-dependent code
```
In `@unbndfun/sum.m`:
```
-maskLeft = ( abs(vends(1, :)) > 1e4*tol );
-maskRight = ( abs(vends(2, :)) > 1e4*tol );
+maskLeft = ( abs(vends(1, :)) > 1e5*tol );
+maskRight = ( abs(vends(2, :)) > 1e5*tol );
```

Looking at my original question (2) above, I'm happy to clarify what I was asking about, but first, could you explain that if `epslevel` is having such a limited effect, why these changes are having to be made (I acknowledge that they're not dramatic, but at the same time, they must have been made for some reason)?

jaurentz:
@asgeirbirkis the few changes to the actual code were the result of the code behaving inappropriately on specific machines. The `@chebfun/thresholdBreakpointValues.m` change was only required for one machine (mine). Every other machine that we tested did not have this problem. Similar statements can be made for the other three changes you listed. For whatever reason (matlab version, C libraries, etc.), these subroutines did not behave in a standard way on different machines when `epslevel` was changed to `eps`. Changes by factors of 10 mean that results only differed in the last few bits on different machines. This does not, in my opinion, imply that `epslevels` was tracking machine specific behavior. It is more likely that the loose estimates propagated by `epslevels` made us oblivious to these discrepancies.

asgeirbirkis:
OK, thanks for clarifying.

Now returning back to my original question above.  Take for example lines 82-87 in `tests/bndfun/diff`:
```
f = bndfun(@(x) x.*atan(x) - x - 0.5*log(1 + x.^2), struct('domain', dom), ...
    pref);
df2 = diff(f, 2);
df2_exact = @(x) 1./(1 + x.^2);
err = df2_exact(x) - feval(df2, x);
pass(9) = (norm(err, inf) < 1e7*get(df2, 'vscale')^2*get(df2, 'epslevel'));
```

Previously, the last line read
```
pass(9) = (norm(err, inf) < 1500*get(df2, 'vscale')^2*get(df2, 'epslevel'));
```

Now, I've been looking at the code, and as far as I can tell, `get(df2,'vscale')` still returns the same number,  and `norm(err,inf)` still gives the same answer. So my question was simply to confirm, have you changed `1500` only because previously, `get(df2, 'epslevel')` would return `2.4692e-10`, but now it gives machine epsilon, `2.2204e-16`? If that's the case, I'm happy. Obviously, I'm not talking just about this test, but all the tests you changed in a similar manner.

In other words, the only reason why you're doing those changes in the tests is because now `get(f,'epslevel')` always returns machine epsilon, and not some estimate of the accuracy, and all `vscales` and errors like computed above have remained the same? Hope this question makes sense?

On another subject, I now count 334 instances of the comment `% tolerance loosened in epslevel-dependent code` in the code. I don't think we should be cluttering the main Chebfun code with this, would you be happy to remove it? In case you believe this will be useful info later on, maybe store it in a file somewhere before deleting it?

jaurentz:
You're absolutely correct. The only difference is that `get(f,'epslevel')` always returns `eps`. In order to have the tests pass we had to loosen the tolerances whenever `epslevel` was previously large. 

As for the comments, I am perfectly happy removing them, but maybe others should chime in on this.

trefethen:
I am happy with the idea of removing those comments.  If we ever want to get this information back -- which let's face it, is pretty unlikely -- we always have Git.

@chebfun/disp.m
```diff
@@ -113,8 +113,7 @@ function disp(f)
 end
 
 % Display epslevel:
-s = [s, sprintf('epslevel = %i', epslevel(f))];
-s = [s, sprintf('   vscale = %i', vscale(f, 'sup'))];
+s = [s, sprintf('Vscale = %i.', vscale(f, 'sup'))];
```
asgeirbirkis: @jaurentz this got changed to `vscale` in #1446 , so I'm changing it back.

tests/chebtech/test_extractBoundaryRoots.m
```diff
@@ -90,9 +92,9 @@
     [g, l, r] = extractBoundaryRoots(f, [ml+1; mr+2]);
     gexact = testclass.make(@(x) exp(x), [], pref);
     err = feval(g, x) - feval(gexact, x);
-    pass(n, 9) = ( (norm(err, Inf) < (1e1^(ml + mr))*f.epslevel) && ...
-        ( l == ml ) && ( r == mr ) );    
-    
+%     pass(n, 9) = ( (norm(err, Inf) < (1e1^(ml + mr))*f.epslevel) && ...
```
asgeirbirkis: @jaurentz any reason why this test got commented out?

tests/trigtech/test_compose.m
```diff
@@ -12,12 +12,16 @@
 f = testclass.make(@(x) pi*cos(pi*(x-0.1)));
 g = compose(f, @sin, [], [], pref);
 h = testclass.make(@(x) sin(pi*cos(pi*(x-0.1))));
+n = max(length(g),length(h));
+g = prolong(g,n); h = prolong(h,n);
```
asgeirbirkis: @jaurentz: Why is it only now that we need to call `prolong`? Similar for `tests/trigtech/test_real`.

Issue #1473, opened 2015-05-01
Plotting chebfun to specific axes is not supported


asgeirbirkis:
While investigating #1465, I came across the following bug, which I believe was discussed a long time ago. Essentially, it's not possible to command Chebfun to plot to specific axes, like is possible to do in standard Matlab. These lines confirm that the first plotting command plots to the first axes created, while the second command causes an error:
```
fig1 = figure(1);
title('Figure 1','handlevisibility','off')
ax1 = gca;
set(ax1, 'nextplot','replacechildren')
fig2 = figure(2);
title('Figure 2');
ax2 = gca;

xx = linspace(-1, 1, 100);
plot(ax1, xx, sin(xx))

x = cheb.x;
plot(ax2, cos(x))

Error using plot
A numeric or double convertible argument is expected 
```

asgeirbirkis:
Unless I'm mistaken, then based on my experiments, we don't stand a great chance to ever successfully overload `plot` so that `plot(ax2, cheb.x)` works. However, we should be able to achieve something like `plot(cheb.x, 'axes', ax2)`.

asgeirbirkis:
It appears that I've found a way to circumvent this issue (see an upcoming pull request), so I'm demoting it to a low-priority one.

PR #1476, opened 2015-05-05
Enhance chebgui pde speedup

### LINKED ISSUE #1465: Allen-Cahn demo in chebgui is slow
As pointed out by @trefethen, the Allen-Cahn demo in chebgui runs much slower in v4 than it did in v5.

My investigation leads me to think it’s actually not a v4/v5 issue, but more to do with plotting at each timestep. If I export the problem from the GUI, and run in v5.0, it takes 4.00 seconds, while it takes 4.07 seconds on the current version of development. But if I run the demo straight from the GUI, rather than an m-file, it takes much longer.

These are the times I get when solving the problem in chebgui:
With plotting of every time step enabled: Elapsed time is 1.434559 seconds.
With plotting of every time step disabled: Elapsed time is 8.025111 seconds.

It appears that the slowdown we’re seeing is due to changes we made in order to make the code (and in particular, chebgui) work for 2014b graphics of Matlab.

The code does not do anything special (yet) about periodic BCs in PDEs, so it’s not due to a change in the way we treat those.

For reference, solving the exported script from v4 takes 6.129 seconds, so if we crack this, we’ve actually gained a decent amount of speed!

asgeirbirkis:
Various speedups and improvements for the PDE side of chebgui, and pde15s in general. These include:

* Significant speedups for pde15s, addressing a lot of slowdowns caused by plotting each timestep.
* Correct colormaps for waterfall plots for pdes in chebgui.
* Correct variable names for pdes when pde15s is called outside of chebgui.
* Fixing bug in exporting PDE systems from chebgui.

This closes #1465 .


nickhale:
@asgeirbirkis - Can you please supply some further detail about what the changes to `pde15s` are?

asgeirbirkis:
@nickhale , sure.

1. Moving  `parseFun` higher up in the code, to get the correct variable names in the non-gui case (for printing the correct legend, xlabel and ylabel).
2. Rather than calling chebfun plot, we evaluate each solution on a 300 point grid and plot the discrete values. This was necessary because as described in #1473 , we can't plot a chebfun directly to axes.
3. Setting correct strings for legends, xlabels and ylabels.
4. Cleaning up plotFun. E.g., rather than repeatedly setting the legend, we set it once at the start. This saved a lot of time. Then, also, rather than repeatedly calling plot, I simply update the y-values of the lines, which is a much faster operations. Also, similarly to the legend, xlims, ylims, xlabels and ylabels are specified at the start, before timestepping.

The combination of 2 and 4 was what allowed the significant speedups, as tabled in #1465 .

nickhale:
I'm mostly happy for this to be merged. My only two questions are:
 * Is it right to fix `nPlot = 300;`? Why not let it depend on the length of the solution.
 * Aren't we duplicated some behavious of `plotData()` by calling `Udata = feval(U, plotGrid);`?

Do we lose much by doing the following instead? (At line 388)
```
        % Get plotting data:
        Udata = plotData(U);
        UdataX = Udata.xLine;
        UdataY = Udata.yLine;
        
        % Update the YData of the lines plotted
        if ( ~doHold )
            for hCounter = 1:length(hLines)
                set(hLines(hCounter), 'XData', UdataX);
                set(hLines(hCounter), 'YData', UdataY(:, hCounter)');
            end
        else
            % Reset color cycle prior to point plot if running R2014b.
            if ( newMatlabVersion )
                set(axesSol, 'ColorOrderIndex', 1);
            end
            hLines = plot(axesSol, UdataX, UdataY, plotOpts{:});
        end
        
```

asgeirbirkis:
* Regarding `nPlot = 300`, that gave me something that looked smooth enough on my laptop with 2014b, retina display, so I figured that it should work for other resolutions as well.
* `plotData` does a lot of stuff that isn't needed here, deal with breakpoints, infinite domains, finding xlims and ylims, deltas, and x points and y points. I'd regard it to be a massive overkill in this case.
* `xData` shouldn't need to get changed between timesteps, at least not if we keep the plotting resolution fixed, hence we get away with simply evaluating the solution on the grid, rather than calling `plotData`.

asgeirbirkis:
Actually, experimenting with it more carefully, it seems that using `plotData` does not cause a significant slow down (somewhere around 5-20 hundredths of a second), so for the sake of code reuse and nice software design, I'd be OK to use that. Commit 3143122 does that.

nickhale:
:thumbsup:

PR #1477, opened 2015-05-05
Fixing broken EIG problems in chebgui.

asgeirbirkis:
Problems were incorrectly being determined to be generalised EIG problems when in fact they were not.

### DIFF
--- a/@chebguiExporterEIG/exportInfo.m
+++ b/@chebguiExporterEIG/exportInfo.m
@@ -137,13 +137,12 @@ catch ME
 end
 
 % Check for a generalised problem.
-% TODO: This should be a joint method with solveGUIeig
 if ( ~isempty(rhsString) )
     RHS  = eval(rhsString);
     N_RHS = chebop(RHS, d);
 
     try
-        B = linop(N_RHS);
+        B = linop(N_RHS, 0);
     catch ME
         MEID = ME.identifier;
         if ( guiMode  && ~isempty(strfind(MEID, 'linop:nonlinear')) )
--- a/@chebop/linop.m
+++ b/@chebop/linop.m
@@ -1,4 +1,4 @@
-function [L, f, FAIL] = linop(N)
+function [L, f, FAIL] = linop(N, paramReshapeFlag)
 %LINOP   Convert a CHEBOP to a LINOP.
 %   L = LINOP(N) converts a CHEBOP N to a linop L if N is a linear operator. If
 %   N is not linear, an error message is returned.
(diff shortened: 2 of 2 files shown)


nickhale:
* Can you please expand on `% For generalized eigenvalue problems, it is useful to pass PARAMRESHAPEFLAG = 0.`
 * Passing the flag option to the `linop` constructor this way means we're stuck with the syntax for ever. Wouldn't `linop(N, 'param', 1)` (or similar) be better / more flexible?

ajt60gaibb:
I thought all problems were discretized as generalized eigenvalue problems now. (This could be a good idea because of balancing and the fact that collocation matrices usually have poor col/row scaling.)

asgeirbirkis:
@nickhale 
* With the merging of #1418, the following now happens when we try to linearize a chebop where no derivatives appear:
```
>> A = chebop(@(u) u);
>> L = linop(A);
>> matrix(L, 6)
ans =
    1.0000
    1.0000
    1.0000
    1.0000
    1.0000
    1.0000
```

Compared with what happens on the master branch:
```
>> A = chebop(@(u) u);
>> L = linop(A);
>> matrix(L, 6)
ans =
    0.7337    0.3187   -0.0761    0.0392   -0.0282    0.0127
   -0.1707    0.9813    0.2512   -0.0984    0.0660   -0.0293
    0.0337   -0.0909    0.9961    0.0881   -0.0468    0.0199
    0.0199   -0.0468    0.0881    0.9961   -0.0909    0.0337
   -0.0293    0.0660   -0.0984    0.2512    0.9813   -0.1707
    0.0127   -0.0282    0.0392   -0.0761    0.3187    0.7337
```

Refer to #1418 to see why those changes were made, and all the issues it fixed.

For problems that are posed as generalized eigenvalue problems, but aren't really in fact, that is, when the RHS operator is the identity operator, we actually still want the old-style version of the discretization. That's what the `paramReshapeFlag` achieves in `chebop/linearize` (the passing of that argument to `chebop/linop` is simply being put forward to `chebop/linearize`).

Come to think of it, this might actually be better dealt with in the chebgui code, where we try to figure out whether the problem posed is generalized or not. So please hold off any merging until I've explored this further.

@ajt60gaibb : Not sure, looking at `linop/eigs` makes me think that's not the case. In any case, we could do that internally, i.e. pose the problem as standard eigs problem, but if no RHS operator is passed, just take it internally to be the identity operator. What's going wrong here is that problems are incorrectly being passed as generalized eigs-problem, which looks wrong in the exported m files (and at the moment, breaks the code anyway).

asgeirbirkis:
Closing this pull request, there's a chance for a lot of code simplification and improved treatment for EIG problems in chebgui that I want to carry out, will do so in an upcoming pull requests.

ajt60gaibb:
@nickhale  Can you clarify what is happening here?  Where did the rectangular projection matrix go (if it is a standard eigenproblem)?

nickhale:
I'm not sure that I understand the question, but it seems as though we _always_ solve generalized problems. This is necessary when using cheb-based discretisations, but perhaps not so when using trig-based (which was implemented later).

PR #1478, opened 2015-05-06
Fast tall-skinny QR

ajt60gaibb:
I am excited by this.  It is the first algorithmic speed up of the quasimatrix QR factorization in Chebfun for years.  It is also the first `dlt()` and `idlt()` Chebfun application.  

Let `A` be a `inf x m` quasimatrix with columns of maximum polynomial degree `n`.  The current QR factorization algorithm has a complexity of    `O( n^2m + nm^2 )`.   Using `ndct` and `idlt`, I changed the algorithm slightly so the QR factorization now costs  `O(nm (log n)^2/loglog n + nm^2 )` operations.  This is a big win for tall-skinny quasimatrices. 

For example (do not try this if you are not on this branch!): 
```
A = chebpoly(1e4:1e4+1);        % inf x 2 quasimatrix 
tic, [Q, R] = qr( A ); toc 
Elapsed time is 0.843294 seconds.
```
In `development` this crashes MATLAB as it tries to construct an `10000x10000` dense matrix. 

We can work out orthogonal bases for sets of functions with *extremely* high degree: 
```
x = chebfun('x'); 
f = cos( x * (1e4:1e4:5e4) );
tic, [Q, R] = qr( f ); toc 
Elapsed time is 9.336177 seconds.
```
Here, the maximum degree in a column of f was 50336... so I am happy with 10 seconds. 

The algorithm currently in development cannot get over `n = 5000`.  I wish I knew about this a week ago, but it will probably appear in the DLT paper after it is back from the review.

### DIFF
--- a/@chebtech/qr.m
+++ b/@chebtech/qr.m
@@ -110,91 +110,89 @@ if ( n < m )
     n = m;
 end
 
-% Project the values onto a Legendre grid: (where integrals of polynomials
-% p_n*q_n will be computed exactly and on an n-point grid)
-if ( (length(WP) ~= n && n <= 4000) || ((~isempty(type) && isa(f, type)) && n <= 4000)  )
-    xc = f.chebpts(n);
-    vc = f.barywts(n);
-    [xl, wl, vl] = legpts(n);
-    P = barymat(xl, xc, vc);     % Map from Chebyshev values to Legendre values.
-    W = spdiags(sqrt(wl.'), 0, n, n); % Weighted QR with Gauss-Legendre weights.
-    Winv = spdiags(1./sqrt(wl.'), 0, n, n);    % Undo the weighting used for QR.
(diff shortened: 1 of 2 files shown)


nickhale:
This is neat. I refactored things a little and added a few more inline comments to make things clearer.

My only question is that since the algorithms for n <= 4000 and n > 4000 are so similar, with the only difference being how some of the transforms (essentially mat-vecs) are performed, is it worth factoring these out? My own opinion is that it's probably not worthwhile here as the amount of duplication is quite small, but I thought I'd run it past @aaustin141 first.

ajt60gaibb:
Thanks.

aaustin141:
I agree with @nickhale's assessment that trying to factor things out here is probably not worthwhile.  The duplication is annoying but not really severe, and there's a chance that trying to factor things out here could make the code less clear.

Issue #405, opened 2014-05-09
pde15s taking an incredibly long time for Maxwell example


gradywright:
The following code is used in the first part of the Maxwell's equation example (note it has been modified to use v5 syntax):

```
pdefun = @(t,x,E,B) [ diff(B), diff(E) ];
bc.left = @(t,x,E,B) E;
bc.right = bc.left;
d = [-2,2];
x = chebfun('x',d);
E0 = exp(-16*x.^2);
t = 0:0.05:5;
soln = pde15s(pdefun,t,[E0,-E0],bc,pdeset('plot','on'));
```
This code runs really fast in v4, but runs incredibly slow in v5.  In v5 on my machine, the solution is computed up to about t=0.5 in a reasonable amount of time, but then the code slows down considerably and appears to get stuck without advancing any further.  Eventually I had to kill the process.

asgeirbirkis:
With the introduction of `pde23t`, are we happy to close this issue?

nickhale:
:+1:

asgeirbirkis:
The introduction of `pde23t` in #1481 now means that we can happily solve the problem above.

Issue #1191, opened 2014-10-05
Rewrite PDE15S to use coefficients rather than values?


nickhale:
`pde15s()` typically struggles when presented with a problem with little or no diffusion (see #405). My suspicion is that the main cause of this is that the coefficient 'tail' used to determine happiness becomes polluted by noise which leads to intractably large discretisation lengths. It's possible that staying in coefficient space will help to alleviate this problem.

nickhale:
I finally got around to trying this. The figures below are `plotcoeffs()` plots of the Chebyshev coefficients from (a) the current `pde15s()` implementation* and (b) a branch where I stay in coefficient space.

The problem solved is a simple constant-coefficient Dirichlet advection diffusion problem. (I chose this because Dirichlet conditions are trivial to implement and because I couldn't be bothered to implement multiplication in coefficient space - although this can be done as in `@chebtech/times()`.)

![coeffs1](https://cloud.githubusercontent.com/assets/2035010/11146902/f11bf972-8a1a-11e5-8abe-52822605b5a7.png)

ajt60gaibb:
Multiplication in coefficient space won't help the "polluted by noise" problem.  

I only see one plot, not sure what I am looking at...

ajt60gaibb:
This is glorious! :+1:  I assume one can make diffusion even weaker so that `standardchop()` won't work as the plateau becomes higher.   I would be interested to see your code. 

Does this need variable coefficients that are polynomial and known exactly?

nickhale:
I've been working on the branch feature-pde15s-coeffs. So far it only works with constant coefficients. multiplication in coefficient space is there now (but only for term `u.^2` - i was lazy). The following shows that staying in coefficient space alleviates the plateauing problem even here.

`pdefun = @(t,x,u) 0.1.*diff(u,2) + diff(u) - u.^2;`

![coeffs1](https://cloud.githubusercontent.com/assets/2035010/11176436/ac2eaec2-8c45-11e5-8bb8-cab65d661430.png)
![coeffs2](https://cloud.githubusercontent.com/assets/2035010/11176437/b2434b24-8c45-11e5-9985-4232992db66a.png)

ajt60gaibb:
Great!   I am a little surprised this worked for a nonlinear pde.  Any insight?

nickhale:
[...]

PR #1481, opened 2015-05-11
Add pde23t() method which is better than pde15s() for advection-dominated problems.

nickhale:
We know that `pde15s()` does pretty well for diffusive problem, but can struggle otherwise. The archetypal example is Maxwell's equation (see #405.)

While experimenting with a coefficient-based version of `pde15s()` I noticed that a lot of noise was being placed in to the highest-order Chebyshev coefficients by `ode15s()` during the time-stepping phase. On a whim I decided to try one of the other MATLAB solvers, `ode23t()`. (This might seem a strange choice, but because of the way we enforce boundary conditions we require the ability to solve with singular Mass matrices, and `ode23t()` and `ode15s()` are the only built in solvers which can do this.)

It turned out this was a *lot* faster on the Maxwell's example than `pde15s()`, even for the existing values-based approach. I decided to create a `pde23t()` method so that this can be explored further. Below is the comparison I made:

```
pdefun = @(t, x, E, B) [ diff(B), diff(E) ];
bc.left = @(t, x, E, B) E;
bc.right = bc.left;
d = [-2, 2];
x = chebfun('x', d);
E0 = exp(-16*x.^2);
t = 0:0.05:1;
opts = pdeset('plot', 'off');
```
```
>> tic, soln = pde15s(pdefun, t, [E0, -E0], bc, opts); toc
Elapsed time is 11.051078 seconds.
```

```
>> tic, soln2 = pde23t(pdefun, t, [E0, -E0], bc, opts); toc
Elapsed time is 1.327562 seconds.
```

Plotting the Chebyshev coefficients gives an indication as to why `pde15s()` is so much slower:

```
plotcoeffs(soln)
```
coeffs1

```
plotcoeffs(soln2)
```
coeffs2


asgeirbirkis:
@nickhale should we make this an option in Chebgui as well, i.e., allow users to select between pde15s and pde23t for their problems?

nickhale:
Sure, that sounds sensible.

asgeirbirkis:
OK, I'm happy to write the functionality for the GUI option.

However, I'm looking at the pull request now, and I have an issue with `pde23t` simply calling `pde15s` with a different option. Surely, there should be a method at top, e.g. `solvePDE`, that both methods should call with different options?

Along those lines, maybe we should use this opportunity to refactor the `pde15s` method? As we all know, it's a beast of an m-file, as I recently got reminded of while working on #1476. It'd make a lot of sense to me to pull out most of the subfunctions and put in separate files, perhaps via introducing a new helper class?

On a more ambitious and interesting note, how realistic would it be to try to diagnose PDEs before calling the solvers, and choose pde15s or pde23t according to whether the problem is diffusive or not?

nickhale:
Thanks.

I thought about completely refactoring things, but I decided that perhaps before expending too much effort on that it would be best to get some idea of how useful the new functionality would be and whether a more extensive rewrite would be worthwhile. (In principle there's no reason to to allow user defined ode solvers, as long as they have the required functionality (i.e., allowing singular mass matrices)).

Again, I'm not sure if refactoring `pde15s()` in its current state is worthwhile. In particular, if a lot of effort was being expended on this, I would almost certainly advocate moving to a coefficient-based approach rather than values-based (see #1191). At one point my understanding was that `pde15s()` was going to be made obsolete by spectral deferred correction, but I haven't heard anything from that direction in a long time.

MATLAB makes no attempts to diagnose ODEs, it just gives you the plethora of `ode***()` routines and tells you which is best in certain situations. I don't think we need to try and automate the process for ODEs.

asgeirbirkis:
OK, maybe we should hold on with the complete refactoring, but as a compromise in the meantime, I think it'd be worth introducing the `pdeSolve` method that both `pde15s` and `pde23t` would call -- I found this confusing when I first looked at it, and that's even after having spent decent amount of time in the PDE district of Chebfunland, so anyone else looking at this will probably find it confusing as well. I think all it'd require is essentially to rename the current `pde15s` to the new method, and then make `pd15s` look similar to the new `pde23t`.

nickhale:
One possible reason for the difference in performance between the two solvers (although I stress that I have essentially no evidence to back this up) is that `ode15s()` introduces high frequency noise at each time step whereas `ode23t()` introduces low frequency noise. 

The former is much worse in this setting, both because the higher-degree Chebyshev coefficients then grow rapidly as we take further derivatives and because it causes 'unhappiness' and forces the `length` of the chebfun to grow.

PR #1488, opened 2015-05-12
Feature pde15s periodic

nickhale:
In the past we dealt with periodic problems by enforcing
certain derivative conditions on the solution. A much
better way is to use a periodic basis, thus negating the
need to enforce any boundary conditions at all.

The following, which hangs on development after t = .5,
now works perfectly:

```
dom = [-1 1];
t = 0:.01:1;
pdefun = @(t,x,u) diff(u);
bc = 'periodic';
x = chebfun(@(x) x, dom);
u0 = sin(3*pi*x);
opts = pdeset('Eps', 1e-6, 'Ylim', [-1.2,1.2]);
[t, sol] = pde15s(pdefun, t, u0, bc, opts);
```

Note, this branch is based off of `feature-pde23t` (and based off of an earlier implementation in `feature-fourtech-pde`). I propose we merge this branch with `feature-pde23t` before merging to development.


asgeirbirkis:
Closing this request and issuing an request between the two branches, cf. #1489.

asgeirbirkis:
Reopening this now that #1498 has been merged.

@chebfun/pdeSolve.m
```diff
@@ -20,10 +20,11 @@
 plotOpts = {'-'};       % Plotting style
 adjustBCs = true;       % Adjust inconsistent BCs
 throwBCwarning = true;  % Throw a warning for inconsistent BCs
+timeChunks = 51;        % Default number of time slices if not specified
 
 % Parse the variable inputs:
 if ( numel(varargin) == 2 )
-    opt = varargin{1};
+    O = varargin{1};
```
asgeirbirkis: @nickhale should this be `opt`, or should the line just be deleted? In any case, the variable `O` does not appeared to be used anywhere else in the code.

nickhale: Should be opt.

PR #1489, opened 2015-05-12
Feature pde15s periodic

asgeirbirkis:
Replacing #1488 with this one. Per @nickhale's original message:

In the past we dealt with periodic problems by enforcing
certain derivative conditions on the solution. A much
better way is to use a periodic basis, thus negating the
need to enforce any boundary conditions at all.

The following, which hangs on development after t = .5,
now works perfectly:

```
dom = [-1 1];
t = 0:.01:1;
pdefun = @(t,x,u) diff(u);
bc = 'periodic';
x = chebfun(@(x) x, dom);
u0 = sin(3*pi*x);
opts = pdeset('Eps', 1e-6, 'Ylim', [-1.2,1.2]);
[t, sol] = pde15s(pdefun, t, u0, bc, opts);
```

Note, this branch is based off of `feature-pde23t` (and based off of an earlier implementation in `feature-fourtech-pde`). I propose we merge this branch with `feature-pde23t` before merging to development.

### DIFF
--- a/@chebfun/pdeSolve.m
+++ b/@chebfun/pdeSolve.m
@@ -68,6 +68,7 @@ else
 end
 
 userMassSet = false;
+isPeriodic = false;
 
 %% %%%%%%%%%%%%%%%%%%%%%%%%%%  PARSE INPUTS TO PDEFUN  %%%%%%%%%%%%%%%%%%%%%%%%%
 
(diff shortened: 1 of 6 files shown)


asgeirbirkis:
On a second thought, we came to the conclusion that it'd be better to first get pde23t merged, then deal with the periodic PDEs (which require a bit more changes), so I'm closing this request. Once the periodic PDE functionality is ready for a review, #1488 should be reopened.

PR #1492, opened 2015-05-13
Implement trigonometric version of barycentric interpolation formula.

nickhale:
This isn't being used anywhere yet, but I'll need it for the periodic version of `pde***()`.

(Also, I think it's nice to have.)

As a side note, it might be nice if the Horner scheme was factored out of `trigtech/feval()` in the same was as `clenshaw` is separate from `chebtech/feval()`, but I haven't done this here.


aaustin141:
How is this different from / how does it interact with the `trigBary.m` and `trigBaryWeights.m` stuff that we have in the "trunk" folder?

nickhale:
Ah. Poop.

PR #1496, opened 2015-05-13
Feature trigspec

Hadrien-Montanelli:
This branch implements a Fourier spectral method in coefficient space for solving linear, nonlinear and system of periodic ODEs. It uses the class `trigspec`, the equivalent of `ultraS` in the Fourier world.
I added a class between `chebDiscretization` and `ultraS`/`trigspec`, called `spec`, because as we could expect, these two classes share many methods.  

By default, periodic ODEs will be solved with `trigcolloc`, but one can use `trigpsec` as follows:
```
dom = [0 2*pi];
a = chebfun(@(x) sin(x) + 2, dom);
f = chebfun(@(x) cos(10*x), dom, 'periodic');
L = chebop(@(u) diff(u) + a.*u, dom);
L.bc = 'periodic';
pref = cheboppref();
pref.discretization = 'trigspec';
u = solvebvp(L, f, pref); 
norm(L*u-f,inf)
```

```
ans = 1.0134e-14
```


asgeirbirkis:
I'm happy to review the code; however, have we converged on the names that this pull request introduces?

Hadrien-Montanelli:
Thanks @asgeirbirkis! No we haven't. I sent an email around but didn't get any answer.

trefethen:
I can live with the name `trigspec` provided that's just the name of the class, nothing to do with what we tell users.  Specifically, we should not tell users that "spectral methods" are distinct from "collocation methods", for they are not.  Everything is spectral; it's just a distinction between values-based and coefficients-based.

asgeirbirkis:
The user will have to specify that he/she wants a coefficient based periodic discretization at some point in the code for this new class to be utilized, i.e.
```
p = cheboppref();
p.discretization = 'trigspec';
u = solvebvp(N,0,p);
```
At the moment, for value based periodic discretization, we have `trigcolloc`. The way I could thing of to convey the message that @trefethen writes above would be to name the two classes `trigval` and `trigcoeff`, i.e.
```
p.discretization = 'trigval'/'trigcoeff';
```
(`trigcoeffs` is already a name of a method), although then we're in conflict with the names `chebcolloc1/2`.

Basically, I don't think there is a perfect solution to this problem, we just need to decide in what norm we want to minimize our frustration...

Hadrien-Montanelli:
I let you chose the norm! I don't think `trigval` and `trigcoeff` are particularly better than `trigcolloc` and `trigspec` but I really don't have a strong opinion on that.

trefethen:
Would
```
p.discretization = 'vals'/'coeffs'
```
work or perhaps
```
p.discretization = 'values'/'coeffs'
```
?

asgeirbirkis:
That's certainly feasible; when we encounter `N.discretization = 'periodic'`, we automatically switch to periodic discretizations, so `p.discretization = 'values'/'coeffs'` could decide between `trigcoeff/trigspec` in the periodic case, and `chebcolloc2/ultraS` in the non-periodic case. Of course, we could still offer more fine-tuning for advanced users, i.e.
```
p.discretization = 'chebcolloc1';
```
if someone really wants to use that one, or
```
N.bc = 'periodic';
p.discretization = 'chebcolloc2';
```
for overwriting periodic discretizations.

I'm happy with this solution, @Hadrien-Montanelli, if you make these changes, I can review this request. This will also require some changes to chebgui which I guess I should write, but I'll submit a separate pull request for it.

Hadrien-Montanelli:
Ok, great!

ajt60gaibb:
I have been using `trigspec` for about a week for solving PDEs on the sphere. It's excellent! Thanks @Hadrien-Montanelli. 

(It's great to have the periodic analogue of the ultraspherical spectral method in Chebfun.)

Hadrien-Montanelli:
I am glad it's working well! 

Hadrien

aaustin141:
FYI:  This no longer merges cleanly into development, owing to some (easy-to-fix) conflicts with the changes to `chebop` preferences made in #1562.

asgeirbirkis:
@Hadrien-Montanelli as @aaustin141 points out, this no longer merges cleanly into development; could you please merge development into this branch, sort out the conflicts and push, so that I can finish the code review and hopefully get it merged soon!

Hadrien-Montanelli:
I fixed that -- thanks @asgeirbirkis and @aaustin141.

@chebop/solvebvp.m
```diff
@@ -189,8 +189,19 @@
 
 % Solve:
 if ( all(isLinear) )
+    % Ensure that rhs-residual is of correct discretization, and convert it to a
```
asgeirbirkis: @Hadrien-Montanelli this is also done in lines 227 and below of the code, so couldn't we move it ahead of the `if( all(isLinear)) ` block?

@chebop/solvebvp.m
```diff
@@ -227,6 +238,15 @@
 % simplify output
 u = simplify(u,pref.errTol/200);
     
+    % Enforce the function to be real if the imaginary part is small if using a
```
asgeirbirkis: @Hadrien-Montanelli  I would have thought this to be better suited for the toFunctionOut methods, could you please explain why you moved it out of those methods and into here in 3f6b808 ?

@coeffsDiscretization/coeffsDiscretization.m
```diff
@@ -0,0 +1,80 @@
+classdef coeffsDiscretization < opDiscretization
+
```
asgeirbirkis: @Hadrien-Montanelli I would have expected documentation at the start of this file, similar to what we have at the start of `valsDiscretization`. Could you please add such documentation?

@opDiscretization/testConvergence.m
```diff
@@ -1,65 +1,65 @@
-function [isDone, epslevel, vscale, cutoff] = testConvergence(disc, values, vscale, pref)
```
asgeirbirkis: @aaustin141 these two versions look pretty similar to me, so is some whitespacing issue tricking git here? Is it worth trying to sort out so that the history of this file doesn't get further corrupted (the culprit seems to be ef119d43 by @jaurentz ), or should we not worry about it?

aaustin141: It's a line-endings issue.  The version prior to ef119d4 has mixed CRLF and LF endings, and the version after it has LF endings only.  I'm not exactly sure how it got the mixed endings (which is waaay messed up); my best guess is that the file was created on Windows but then got hit by a Unix utility like `sed` that operates line-wise, giving LF terminators to whatever lines it hit and leaving the CRLF elsewhere.  Whatever editor was used to save the file for ef119d4 then converted it to use LF only instead.

I don't think we should worry too much about this at the moment.  The fact that we haven't had many problems with this until now is a sign that most of the editors we're using deal with line endings transparently.  It's only when they get mixed like this that problems occur.  If we run into more issues in the future, we can normalize the endings in the repository and then have people configure `core.autocrlf` to let Git handle them as needed.

asgeirbirkis: Great, thanks for the clarification, happy to let this pass!

asgeirbirkis:
OK @Hadrien-Montanelli , here are the four outstanding issues I find left with this PR, once they've been fixed, I'll be very happy to merge:

1. There is no top level documentation in `coeffsDiscretization/coeffsDiscretization.m`.

2. I think that the following `ultraS` and `trigspec` methods could be moved to the `coeffsDiscretization` level, as they look identical to me: `getOutputSpace.m` and `scale.m` .

3. `trigspec/instantiate->instantiateOne` does not deal with items that are `functionalBlocks`. `ultraS` does so (by going back and forth between value and coeffs space). Is it possible to do so here as well? Furthermore, I think there might be a chance of moving `instantiate` up to the `coeffsDiscretization` level, as the methods look pretty similar for `trigspec` and `ultraS`.

4.  Some of the new methods have copyright at 2014 instead of 2015.

I merged development into the branch this afternoon and pushed some coding style changes. Running chebtest on this now, glad to find that the tests that were failing the other day are passing now!

Hadrien-Montanelli:
@asgeirbirkis Thank you for the feedback. I've pushed 2 commits which address the 4 issues.

asgeirbirkis:
@Hadrien-Montanelli great, thanks, just merged, very glad to finally have this in development!

ajt60gaibb:
Nice. :+1:

Hadrien-Montanelli:
Great thanks @asgeirbirkis for the hard work!

PR #1501, opened 2015-05-18
Make chebdouble/feval more efficient.

nickhale:
We were wasting a **lot** of time computing `chebpts()` at each time step here.


@chebdouble/chebdouble.m
```diff
@@ -241,7 +241,15 @@
         function out = feval(u, y)
             %FEVAL   Evaluate polynomial interpolant of data {X_cheb, U} at a
             % point y using barycentric interpolation.
-            [x, w, v] = chebpts(length(u.values), u.domain);
+            
+            persistent x v
+            
+            N = length(u.values);
+            
+            if ( length(x) ~= N )
```
asgeirbirkis: @nickhale this might sound far fetched, but shouldn't we also be storing the previous domain in a persistent variable, and then check
```length(x) ~= N && all(dom == u.domain)```
Otherwise, we might end up solving PDEs on different domains, but because the `N` of the new PDE matches that of the old, `x` doesn't get recomputed?

nickhale: Ugh. Maybe.

Alternatively we could just store the [-1,1] values and rescale each time.

asgeirbirkis: OK cool, will implement one or the other.

Issue #1232, opened 2014-10-23
chebfun2v.roots fails on a very easy (bivariate linear!) example.


aaustin141:
The following two `chebfun2` objects have an isolated common root at (-1/4, 1/4), but the resultant-based bivariate rootfinding method doesn't seem to find it:

```
>> p = chebfun2(@(x, y) x - y + 0.5);
>> q = chebfun2(@(x, y) x + y);
>> roots([p ; q])
ans =
     []
```

I've managed to track this down to [these lines](https://github.com/chebfun/chebfun/blob/347b601f132690f8b4541312240ebcb07ea6c409/%40chebfun2v/roots.m#L385-389) in `@chebfun2v/roots.m`.  It looks like they are thresholding away something that is important for this problem.  If I comment them out, I get the expected answer:

```
>> roots([p ; q]) 
ans =
  -0.250000000000000   0.250000000000000
```

I'm marking this "high priority" because the example is really basic, but if someone feels that this is unjustified and wants to remove the label, I won't object.

PR #1508, opened 2015-05-26
Fix reversed arguments to eig() call in chebT1rtsmatgep() in chebfun2v.roots().

### LINKED ISSUE #1505: Chebfun2v roots fails on linear example
I have been playing a lot with `chebfun2v/roots` recently and there is an example where it misses a well-conditioned root: 

```
p = @(x,y) y + x/2 + 1/10; 
q = @(x,y) y - 2.1*x + 2; 
r = roots( chebfun2v( p, q ) )
r =
     []
```

It actually has a solution inside `[-1,1]x[-1,1]`:

```
[ .5 1 ; -2.1 1] \ [.1 ; -2]
ans =
   0.807692307692308
  -0.303846153846154
```

See also: 
```
plot(roots(chebfun2(p))), hold on, plot(roots(chebfun2(q)))
```
intersection

aaustin141:
These should be the other way around:  the leading-order coefficient should be the "B" argument, while the constant term should stand in for "A".  This closes #1505 and is the right resolution for #1232.  `chebfun2v.roots` now handles bivariate linear problems correctly.

This commit also reverts the changes made in 8bcf6c4d and 6bb4be81 to address #1232, as the cutoff tolerance is not the issue here.  The reason that making that tolerance smaller appeared to fix things is that it sometimes caused a small leading-order coefficient that in principle _should_ be stripped to not be stripped, so that when the coefficients were passed to `chebT1rtsmatgep` later, the input would have size at least 3 in the third dimension.  This caused the handler for the general case, which does things correctly, to get called instead of the handler for the linear case, which was faulty prior to the changes in this commit.

A test has been added to hopefully prevent this behavior from recurring in the future.

### DIFF
--- a/@chebfun2v/roots.m
+++ b/@chebfun2v/roots.m
@@ -384,7 +384,7 @@ else
     % cutoff negligible B
     nrmB = norm(B(:,:,end),'fro');
     for ii=1:size(B,3)
-        if norm(B(:,:,ii),'fro')/nrmB > eps/2,    break;    end
+        if norm(B(:,:,ii),'fro')/nrmB > 10*eps,    break;    end
     end
     B = B(:,:,ii:end);
     ns = size(B);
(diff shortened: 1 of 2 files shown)


ajt60gaibb:
OK. I totally agree that the order of the coefficients in the bilinear case was incorrect. Thanks @aaustin141 for your very impressive detective skills.

ajt60gaibb:
The battery test of rootfinding examples in `chebtest chebfun2v` passes for me so I am happy.

PR #1515, opened 2015-05-29
Warning for f.epslevel and get(f,'epslevel').

### LINKED ISSUE #1513: What should we return if user calls f.epslevel?
Currently on the master branch, we try to give meaningful answer when a user enquires the epslevel of a function (although exactly how meaningful it is, and what the actual meaning is is debatable):
```
>> f = chebfun(@(x) sin(100*x));
>> f.epslevel
ans =
   1.1057e-14
```

However, with the merge of #1474, which disabled epslevels but didn't remove them, it is still possible to enquire a `chebfun` about its epslevel, however now, the result will be (even more) gibberish:
```
>> f = chebfun(@(x) sin(100*x));
>> f.epslevel
ans =
   2.2204e-16
```

Given the chance that there actually might be some users using the epslevel information, should we at least issue a warning if someone asks about the epslevel of a chebfun?

asgeirbirkis:
After adding a warning in `chebfun.get`, when I then run chebtest, I get a lot of warnings! This is because in a lot of places, the chebtests still include lines such as
```
tol = 10*get(f, 'epslevel')*get(f, 'hscale');
```
and
```
pass(im+1,k) = norm(feval(f,xx) - F(xx), inf) < 1e2*max(f.epslevel.*f.vscale);
```
which were kept because we wanted to keep the option open to easily reintroduce epslevels.

Since removing those lines would make the reintroduction of epslevels a lot trickier (even though it's unlikely we want to do so), I didn't want to change the tests, instead, I turned off the warning inside chebtest.

Closes #1513.

### DIFF
--- a/@chebfun/chebfun.m
+++ b/@chebfun/chebfun.m
@@ -286,9 +286,6 @@ classdef chebfun
         % Display a CHEBFUN object.
         display(f);
 
-        % Accuracy estimate of a CHEBFUN object.
-        out = epslevel(f, flag);
-        
         % Evaluate a CHEBFUN.
         y = feval(f, x, varargin)
         
@@ -473,6 +470,9 @@ classdef chebfun
         % Compare domains of two CHEBFUN objects.
         pass = domainCheck(f, g);        
         
+        % Accuracy estimate of a CHEBFUN object.
+        out = epslevel(f, flag);
+
         % Extract columns of an array-valued CHEBFUN object.
         f = extractColumns(f, columnIndex);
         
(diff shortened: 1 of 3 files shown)


aaustin141:
I presume we're OK with not issuing any sort of warning for `epslevel(f)`?  That's the internal function used to implement `f.epslevel` and `get(f, 'epslevel')`.  I don't think we've advertised it anywhere, so we should be OK.  If we have to issue warnings on it as well, then we're going to need to add many more statements in other places that disable the warning, since `epslevel(f)` is called all the time internally.

trefethen:
Sounds right to me.

asgeirbirkis:
Sounds good to me as well!

PR #1519, opened 2015-05-30
fix trigstripplot

kuanxu:
Fixing the documentation in chebellipseplot and trigstripplot, as well as the formula used in trigstripplot to compute the estimated strip of analyticity.

@mhsnjvd You may want to take a look and let me know your comments.


kuanxu:
I'm closing this as it is now replaced by #1524.

PR #1524, opened 2015-06-01
Overhaul of plotregion

### LINKED ISSUE #1518: chebellipseplot and the estimated strip of analyticity/chebstrip for a trigfun
This issue was encountered when @aaustin141 and I were investigating the stability of trigonometric barycentric formulae. For example:
```
f = chebfun(@sin, [0, 2*pi], 'trig')
chebellipseplot(f)
```
gives an ellipse, instead of a strip, which is not very reasonable. 
ellipse

Should we make `chebellipseplot` work for both trig and non-trig cases by returning a strip for the trig case and an ellipse for the polynomial case? Or we throw an error when a trigfun is passed to `chebellipseplot` and create a new function (maybe called `chebtripplot`) to handle the trig case?

Notice that though @mhsnjvd has implemented `trigstripplot` in `@trigtech` to handle the actual computation of the strips, it has never been used.  

@aaustin141, @mhsnjvd, @Hadrien-Montanelli, @gradywright, Comments?

kuanxu:
1. Enabling plot of the analytic strip of a periodic function in chebfun level.
2. Renaming a bunch of files to better describe the methods.
3. Improving the plots of regions of analyticity.
4. Improving the help text in plotcoeffs@chebfun.

This is to close #1518 and replace #1519.


kuanxu:
Here is an example:
```
>> f = chebfun(@sin, [2*pi 4*pi], 'trig');
>> g = chebfun(@cos, [0 2*pi]);
>> h = join(g, f);
>> plotregion(h)
```
gives
ellipse

All tests passed.

PR #1527, opened 2015-06-05
Refactor bndfun/power.m and unbndfun/power.m

nickhale:
`bndfun/power()` and `unbndfun/power()` were identical.


asgeirbirkis:
@nickhale I'm getting a lot of chebtests failing on this branch on the form:
```
Error using  .^ 
Method 'power' is not defined for class 'bndfun' or is removed
from MATLAB's search path.
Error in chebfun/power>columnPower (line 169)
                g.funs{k} = power(f.funs{k}, b);
```
Even though `bndfun` is clearly supposed to be inheriting the `classicfun` `power` method. Are things working for you?

PR #1533, opened 2015-06-12
Pass hscale information to tech constructors in chebfun2 constructor.

nickhale:
These lines in the `chebfun2` constructor seemed odd:
```
    colData.vscale = dom(3:4);
    tech = pref.tech();
    colChebtech = tech.make(sum(colValues,2), colData);
```
I think they're trying to accound for scale invariance in the domain. I changed it to 
```
    colData.hscale = norm(dom(3:4), inf);
```
which I think is right. 

All the tests pass either way.


ajt60gaibb:
I agree. Thanks.

PR #1540, opened 2015-06-16
Fix slow evaluation method for real-valued trigtechs

gradywright:
The `trigtech/feval` method is noticeably slower than `chebtech/feval` (up to two-times for scalar-valued functions and more for array-valued functions) for similar length real-valued functions.  The issue stems from the use of complex arithmetic in the `trigtech` evaluation routine, which is just standard Horner's method with the polynomial `z=exp(1i*pi*x)`.

This branch implements Horner's method for evaluating real-valued and complex-valued `trigtechs`.  The code (now contained in `horner.m`) checks if the `trigtech` is real and calls the method for doing all computations in real floating point arithmetic.  There are also different routines for array-valued real/complex `trigtech's` to optimize the computations.

The most noticeable difference in speed comes when evaluating an array-valued, real-valued, `trigtech` (something that occurs quite often in `chebfun2`).  For example:
```
>> rv = rand(1,100);
>> x = rand(100,1);
>> f_op = @(x) abs(sin(x-0.3)).^3;
>> f = chebfun(f_op,[-pi pi],'trig')*rv;
```
On the development branch, I get the following on my MacBook Pro:
```
>> tic, val = f(x); toc
Elapsed time is 0.405161 seconds.
```
With the new code, I now get:
```
>> tic, val = f(x); toc
Elapsed time is 0.230274 seconds.
```
I have generally seen speed-ups in the range of 1.5 to about 2 with the new code.

I also tried implementing Clenshaw's method for summing a Fourier series in order to make the evaluation code of a `trigtech` as fast as that as a `chebtech`.  While this method does indeed speed up the code to the level of `chebtech`, it turns out to be unstable for certain ranges of the input (see for example, A. C. R. Newbery, "Error analysis for Fourier series evaluation." Mathematics of Computation 27.123 (1973): 639-644).  The attached plot illustrates this instability by comparing a few evaluation techniques for the function: `f = @(x) exp(cos(1e4*pi*(x-0.35)))`.  The plot shows the difference between evaluating `f` and evaluating the `trigfun` and `chebfun` versions. 

accuracycomparison

There are ways to fix this instability (see referenced article above), but the logic involved in implementing this would seem to diminish any speed-ups.  However, I did not check to see if this would indeed be the case.


trefethen:
Interesting.  Impressive speedup!

Issue #1542, opened 2015-06-19
Why is chebtech/qr not done like this? 


ajt60gaibb:
Hi @nickhale and @trefethen,

Is this a viable way to do `qr()` in `chebtech`?   On my machine it is x2 slower than the `chebtech2/qr()`, but with a bit of Nick Hale magic that could change.  It is certainly a simple approach. 

Idea:  Use the Legendre basis to relate the continuous QR to the discrete QR. Then, use discrete `qr()`.

```
function [Q, R] = myqr( A )
chebA = cheb2leg( A.coeffs );               % Get Legendre coefficients of array-valued chebtech
D = diag( 1./sqrt((0:size(legA, 1)-1)+.5) );% Orthonormalization constants.
[ Qleg, R ] = qr( D*legA, 0);               % Discrete QR
Q = chebtech2( { [], leg2cheb( D\Qleg ) } );% Construct continuous QR
Ds = diag( sign( diag( R ) ) );             % Ensure diagonals of R are of +ve sign.
R = Ds * R; Q = Q * Ds;
end
```

You can check that it computes the right thing with this code: 
```
A = chebtech2(@(x) [cos(100*x.^2), sin(100*x)]); 
[Qexact, Rexact] = qr( A );
[Q, R] = myqr( A ); 
normest( A - Q * R )
normest(Q - Qexact)
normest(R - Rexact)
```

aaustin141:
To be clear---`chebtech` already does QR; there aren't specific `chebtech1` and `chebtech2` overrides for the `qr` method.  What you're suggesting is that we move to doing QR in coefficient space instead of in value space like it currently is, right?

ajt60gaibb:
`chebtech1` and `chebtech2` both use the same Chebyshev basis so the code above works equally in both cases. 

Yes. I was asking why not use Legendre coefficients rather than values taken at Legendre points.

nickhale:
Your way is certainly viable. I would expect the timings to be about the same. (You can probably speed up your code a little by doing the multiplication by `Ds` before you construct the `chebtech2`. Also the multiplication by diagonal matrices might be made faster using sparse or bsxfun.)

Values requires:
 * ndct, idlt and leg2cheb
Coeffs requries:
 * cheb2leg and leg2cheb

Since the idlt is a transposed cheb2leg, this would suggest your approach would be marginally faster (although the ndct is so fast that you probably wouldn't notice). If we has a _really_-fast indct, then values would definitely be preferable.

[...]

PR #1543, opened 2015-06-19
Fix caching in chebtech/qr().

nickhale:
Looking at #1542 made me notice this.

Before this fix:
```
>> tic, for k = 1:100, qr(chebtech2(rand(1000,20))); end, toc
Elapsed time is 2.551894 seconds.
```
After:
```
>> tic, for k = 1:100, qr(chebtech2(rand(1000,20))); end, toc
Elapsed time is 0.868198 seconds.
```

Explanation: We cache the matrix which converts Chebyshev values to Legendre values. This is useful if we are doing many transforms of the same length (which is typical in many `chebfun2` commands).

We could remove the problem of checking for type if we _always_ used cheb1 or cheb2 points locally in `qr`. However, this pull request doesn't address this (and it's actually closely related to #1542)


ajt60gaibb:
This looks great to me!  I note that `chebfun2/plus()` speeds up a little.

PR #1547, opened 2015-07-01
Fix a vscale that should be an hscale.

nickhale:


asgeirbirkis:
Closing this pull request, instead, opening up another one into the release branch (running chebtest on it as we speak).

PR #1548, opened 2015-07-01
Feature remove hscale

nickhale:


trefethen:
Removing `hscale` from chebtech sounds like a fundamental conceptual change.  Can you point us to any information about why it was there once, why it shouldn't be there now, what the idea is, why you think this change is appropriate?

I'm not saying I disagree -- in general my Chebfun bias is always towards simplifying the code where possible -- but I'd like some information about what's going on.  This is of more than passing interest to me since @jaurentz and I are currently working through various matters related to Chebfun construction.

Regarding `vscale` vs. `hscale` removal from `chebtech` (see #1549), am I right in understanding that these are quite non-analogous, because a `vscale` can still be inferred from the data in the `chebtech` whereas an `hscale` can't?

nickhale:
In theory it's a fundamental change, whilst in practice it's not really since the `.hscale` property of a `chebtech` was hardly being used anywhere. (The only place being `chebtech/roots()`.)

I think probably the reason we included it in the v5 rewrite was because it was there in v4 and we weren't really sure what to do about it. Clearly we had some confusion, since the class def files state that the `.hscale` is alway 1 (in which case, why have it?) Probably another was the way it related to updating `.epslevel`, but since that's now essentially a constant, this requirement is removed.

My view is that the `.hscale` of a `chebtech` should by definition (at least the definition we are currently using) be 1. It's the responsibility of the class above to deal with domain scaling, so this should information should be passed down when/if required. I guess this is my answer to your final question: No, the `hscale` can't be inferred from the `chebtech` data. However, it shouldn't be, it should be inferred from the `bndfun` data.

aaustin141:
With #1562 now in place, this no longer merges cleanly into development.  Remarkably, the conflicts don't seem to be that bad to handle except in `@chebDiscretization/testConvergence.m`, for which my diff shows that the entire file changed.  (This could be due to an issue with line-endings.)

asgeirbirkis:
At least according to https://github.com/chebfun/chebfun/commit/ef119d43a06bff77fa898f1e5e24f5654af0b9e3#diff-291e55094123415cfc171483fd36151c , there was a commit by @jaurentz that touched the formatting of `@chebDiscretization/testConvergence.m`, so I find it likely it's a formatting issue.

aaustin141:
This pull request has been superseded by #1584.

PR #1555, opened 2015-07-07
Fix chebgui cheboppref

asgeirbirkis:
Various changes to chebgui to address changes introduced in #1550 .

PR #1557, opened 2015-07-08
Fix linop pref

Hadrien-Montanelli:
The methods `eigs`, `expm`, `matrix`, `null` and `mldivide` are, for the user, CHEBOP methods. Inside the code, the operator is linearized and their analogue LINOP methods are called. At the CHEBOP level, the preferences are parsed and passed to the LINOP level. This pull request gets rid of the preference parsing/checking at the LINOP level, i.e., we assume now that the preferences have been correctly set and passed to the LINOP level. It makes the code clearer and avoids to check things twice.


asgeirbirkis:
@Hadrien-Montanelli I get the following three chebtests failing:
```
   chebfun/test_kronOp.m
   chebgui/test_toFileEIG.m
   chebmatrix/test_plot.m
```
Could you please ensure they pass so that I can look at getting this request merged?

Hadrien-Montanelli:
@asgeirbirkis I fixed `chebfun/test_kronOp.m` and `chebmatrix/test_plot.m` but I don't understand what's going on with `chebgui/test_toFileEIG.m`.

asgeirbirkis:
OK, fixed the broken EIG stuff, however, until #1555 gets merged, we'll keep getting warnings from `cheboppref/parsediscretization`. So I added few warnstate offs in `chebgui/solveGUI` and `tests/chebgui/solvegui_eig`, which should be removed as a part of merging #1555.

PR #1558, opened 2015-07-08
Ensure that initial guess for chebop methods is a CHEBMATRIX.

asgeirbirkis:
This was causing an issue found by @trefethen, namely that "Use latest solution as initial guess" option in Chebgui was crashing.

### DIFF
--- a/@chebop/solvebvp.m
+++ b/@chebop/solvebvp.m
@@ -91,7 +91,11 @@ if ( isempty(N.init) )
     u0 = chebmatrix(u0);
 else
     % Get the initial guess.
-    u0 = N.init; 
+    u0 = N.init;
+    % Ensure that initial guess is a CHEBMATRIX (as later code assumes it is):
+    if ( isa(u0, 'chebfun') )
+        u0 = chebmatrix(u0);
+    end 
 end
 
 % Initialise the independent variable:


aaustin141:
I can't seem to reproduce the issue here on release-v5.2.x.  On development, if I run the "Nonlinear pendulum (third solution)" demo, hit the button to use the latest solution, and then run it again without changing anything else, I get the error "incorrect number of outputs".  Running back through history, I find that this behavior began occurring after b7f505a8.  That commit (part of the multiple-outputs-from-backslash feature) is on development but not the release branch.  If I cherry-pick the commit up for inclusion in release-v5.2.x here (8e8e90b) onto development, the issue is resolved.

@asgeirbirkis, can you confirm that this assessment is correct?  If so, let me know, and I can take care of the rest.

asgeirbirkis:
@aaustin141 yes, your assessment is correct, I was under the impression that the multiple outputs from chebops had been introduced in time for the last release, but misremembered. So this pull request really should be aimed at the development, not release branch, if you're happy to take care of it, that'd be great!

aaustin141:
Per the above discussion, I'm closing this pull request without merging it as-is.  The changes have been incorporated into development in 7074ad05.

PR #1567, opened 2015-07-21
Loosening the tolerance used for sampleTest for the construction of `smoothPart` of singfun. 

kuanxu:
To do so, a new field `sampleTestEps` is added to techPref@chebtech and techPref@trigtech. All tests passed.

### DIFF
--- a/@chebtech/chebtech.m
+++ b/@chebtech/chebtech.m
@@ -13,7 +13,7 @@ classdef chebtech < smoothfun % (Abstract)
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % CHEBTECH Class Description:
 %
-% The CHEBTECH class is an abstract class for representations of smootha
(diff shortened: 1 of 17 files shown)


ajt60gaibb:
I was expecting you to just make a specialised `sampleTest` for `singfun`.  Does this simple change really need a new preference? 

(I will agree with whatever @aaustin141 thinks about it.)

kuanxu:
@ajt60gaibb It seems to me that adding this new preference can make things easier. Otherwise, I need add a new input argument in each layer to pass down a specified tolerance from constructSmoothPart@singfun all the way down to sampleTest@chebtech. But I'm not very sure. I'll see what @aaustin141 say too.

ajt60gaibb:
Oh, I see. @aaustin141 is omniscient.

bhashemi:
All the tests pass on my machine.

ajt60gaibb:
What did @aaustin141 think?

kuanxu:
We just had the discussion in the weekly meeting and I think this fix made both @aaustin141 and @trefethen happy.

ajt60gaibb:
OK. Great!

PR #1568, opened 2015-07-26
Simplify trigtech/sum. 

ajt60gaibb:
There is no need to use the trapezium rule in `trigtech/sum`, requiring `O(n)` operations, when the zeroth Fourier coefficient is already known.   The algorithm is now `O(1)` operations.  (This speed up can be seen with very high rank periodic `chebfun2` objects.)


trefethen:
Thanks for noticing this @ajt60gaibb.

PR #1569, opened 2015-08-03
Vary tolerance in sampleTest() with the construction tolerance.

aaustin141:
The adjustments to the tolerances in `sampleTest` made in 71f7c8d5 cause the
example ode-eig/OpticalResponse.m to virtually hang on my machine.  (Actually,
it doesn't, but it takes just over 20 minutes to finish executing.)  According
to the example text, the function being constructed there has a fair amount of
inherent noise, and it relaxes the construction tolerance accordingly by
adjusting `pref.eps`.  Unfortunately, prior to this commit, the construction
tolerance was not used as a factor in determining the tolerance for
`sampleTest`, and this caused it to be too stringent.

In the past, the tolerance was relaxed for functions like these by `f.epslevel`.
One may then wonder why this example did not start hanging sooner when
epslevels were "deactivated" with the incorporation into development of
61dddd9a.  The reason seems to be that there were a handful of places that
evaded deactivation previously.  These were fixed in the recent merge of the
new `standardCheck` happiness checker in 7541c9c0, and as a result, the example
began to hang.

Note that the changes in this commit mean that `pref.sampleTestTol` (currently
adjusted only by `singfun`) can only be used to relax the tolerance for
`sampleTest` compared to the construction tolerance, not tighten it.  This does
not seem likely to cause any issues, as it doesn't make a whole lot of sense to
sample test more stringently than you construct.  The relationship between
`pref.sampleTestTol` and `pref.eps` and how both of these get incorporated into
choosing tolerances should be revisited when we take a look at `sampleTest`
again in the future.


trefethen:
Thanks @aaustin141.  Yes, it is certainly odd if Chebfun was attempting to use tighter tolerances in `sampleTest` than in construction.

PR #1572, opened 2015-08-05
Fix simplification of an even length trigtech

gradywright:
Even length `trigtechs` are not being simplified correctly.  For example:
```
>>f = trigtech(ones(8,1));
>> simplify(f)
ans = 
  trigtech with properties:
      values: 0
      coeffs: 0
      vscale: 1
      hscale: 1
     ishappy: 1
    epslevel: 2.2204e-16
      isReal: 1
```
This pull request fixes the issue, which turned out to be a mistake in the index for the non-symmetric mode.  This is a pretty trivial pull request (only one line and a new test).

### DIFF
--- a/@trigtech/simplify.m
+++ b/@trigtech/simplify.m
@@ -82,7 +82,7 @@ end
 % Now put the coefficients vector back together.
 coeffs = f.coeffs;
 if ( isEven )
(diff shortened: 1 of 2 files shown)


gradywright:
I should add that the bug was introduced after the current release (with the introduction of standardChop).

PR #1578, opened 2015-08-18
Allow subsref to return empty. Closes #1574.

### LINKED ISSUE #1574: Non-empty chebfun after subsref
Consider the following:
```
f = chebfun(@(x) cos(x))

f =

[      -1,       1]       15      0.54     0.54 
vscale = 1.000000e+00.
```
Now suppose we try to remove the first (and only column) of f using `subsref` beyond that column
```
f(:,2:end)

ans =

[      -1,       1]       15      0.54     0.54 
vscale = 1.000000e+00.
```
I would have expected to get back an empty `chebfun`.   This is what happens for vectors in matlab:
```
a = rand(5,1);
b = a(:,2:end)

b =

   Empty matrix: 5-by-0
```

nickhale:
Allow subsref to return empty. Closes #1574.

### DIFF
--- a/@chebfun/subsref.m
+++ b/@chebfun/subsref.m
@@ -77,6 +77,8 @@ switch index(1).type
             elseif ( max(idx{2}) > columnIndex(end) )
                 error('CHEBFUN:CHEBFUN:subsref:badsubscript', ...
                     'Index exceeds CHEBFUN dimensions.');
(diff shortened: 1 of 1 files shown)


asgeirbirkis:
Got this merged into the release branch, then merged that into development. Also added a test, ca08c1d9 .

PR #1583, opened 2015-08-21
Feature path following

asgeirbirkis:
Finally, the pathfollowing capabilities of chebop are ready to go into the main system (assuming they pass the review).


nickhale:
I make no pretence at understanding what's going on here, but (aside from a few changes I just made) the coding looks good.

If the testing is made more extensive then I'd probably be OK with this being merged. (Particularly since it's essentially independent of the rest of the system.)

asgeirbirkis:
Thanks @nickhale for looking at the code and the code review commit. I'll have a think about how to further test the code, and push some more tests soon!

asgeirbirkis:
@nickhale Just added some more tests if you want to have a look!

nickhale:
I'm happy to merge this, but it would be nice to see an Example to go with it. Eventual inclusion of this functionality in the GUI might also be nice.

Before I merge though I'll quickly question the naming of the method `followpath()`. To me it seems OK, but I know that some people care a lot more about names than I do. Any objections?

trefethen:
Ah, I certainly care about names!  @nickhale, do you have an idea for a better name?

asgeirbirkis:
@nickhale I have an example ready for this functionality which I believe I presented at a meeting some months ago, I'll polish it up and present again at next weeks meeting.

nickhale:
Not really. One possibility is to have a generic `solve` method that takes a `'continuation'` flag, but I'm not sure I like that idea much.

PS. I just notice that `newtonBVP` and `tangentBVP` are not consistent with `solvebvp` and `solveivp`.

asgeirbirkis:
`newtonBVP` and `tangentBVP` are only ever used internally, hence the camel casing, while `solvebvp` and `solveivp` could potentially be called by a user, hence, the camel casing. I don't feel too strongly about this though, so if you want it changed, I'm fine with that.

trefethen:
I have no particular view on this, so long as the user-facing name remains lower case.

PR #1500, opened 2015-05-18
Remove .vscale property from chebtechs.

nickhale:
See discussion in #1479.

This was complicated somewhat by 
 * a) The need to pass vscales in the construction process which do not correspond to function values.
 * b) The fact that we sometimes modified the coefficients, then called f.vscale to look at the old vscale.

As noted in #1479 if `.epslevels` are fully removed, then even more references to `.vscale` can go.

Performance seems comparable (at least in timing `chebtest chebtech`).

Issue #1479, opened 2015-05-07
Is the .vscale needed in chebtech objects?


nickhale:
Since it seems the `.epslevel` property of `chebtech` objects is about to go the way of the Dodo, should we also consider scrapping the `.vscale` field? 

In terms of efficiency, I imagine around 50% of the time we will lose by having to compute the same vscale more than once, and the other 50% we will save by having not computed it at all.

In terms of the code, it would almost certainly make things simpler: `vscale()` would become a method rather than a property and only called when the vertical scale is required. Strictly speaking, the vscale can always be recovered from the `.coeffs`, so storing it is a duplication of information.

I think the only reason we _have_ a `.vscale` currently is a hang-over from v4, where it represented something other than the current magnitude of the chebfun.

trefethen:
A very interesting idea.  @jaurentz and I will bear it in mind as we work further on matters of constructors etc.  Thanks.

jaurentz:
Right now we've only disabled `epslevels`, with plans for removal in the future. Is a straight forward thing to disable `vscales` before completely removing it?

aaustin141:
@jaurentz:  Just to be clear, the proposal is not to eliminate the `vscale` concept (like we've more or less done with `epslevel`; we've just kept the code lying around because we may try to reintroduce it in the not-so-distant future).  It's just to compute `vscale` only on-demand instead of updating it at the end of every operation and storing the result.

I'm generally in favor of doing this, assuming it doesn't too negatively impact performance.  Getting rid of all the code required to update it would definitely simplify things and eliminate improperly updated `vscale` fields from our list of sources of bugs.  On the other hand, `vscale` is probably the most frequently requested property of a tech, so we should measure some experimental timings first before we commit to doing this.

nickhale:
> On the other hand, vscale is probably the most frequently requested property of a tech, so we should measure some experimental timings first before we commit to doing this.

But since accessing properties is so slow in MATLAB, we might actually _gain_ from this..

trefethen:
I very much support the idea: I like simplicity.

aaustin141:
>> On the other hand, vscale is probably the most frequently requested property of a tech, so we should measure some experimental timings first before we commit to doing this.

> But since accessing properties is so slow in MATLAB, we might actually _gain_ from this..

I'm not contradicting that; to the contrary, I think this is probably true.  I'm just saying that we should measure to be sure.

nickhale:
[...]

PR #1584, opened 2015-08-23
fix-remove-unneeded-properties

aaustin141:
These changes accomplish three related tasks:

 - Remove the `hscale` property from `chebtech` and `trigtech`.
 - Remove the `vscale` property from `trigtech` (for `chebtech`, see #1479 and #1500).
 - Completely purge `epslevel` from all layers of the system, finishing the work begun in #1474.

Of course, the _concepts_ of `vscale` and `hscale` haven't disappeared; rather, these quantities are now computed on-demand as needed.  On the other hand, `epslevel` has now completely disappeared, even as a concept.  The only place it shows up anymore is within the bodies of the old happiness-checking functions (`classicCheck`, `plateauCheck`, etc.) and within a couple of internal (i.e., non help-text) comments.

These changes eliminate over 800 lines of code and considerably simplify some of the methods.  (Compare, e.g. this branch's `chebtech.cumsum` with the one on development.)

Most of the commits have my username attached to them, but this work is as much (if not more) @nickhale's as it is mine.


aaustin141:
With all the discussion that's been happening about performance this week, I thought I'd mention that this branch seems to have some performance improvements behind it, at least on my machine under R2015b and measuring in the (admittedly flawed) `chebtest` metric.  On this branch, I'm consistently seeing `chebtest` runtimes of roughly 650-660s for this branch, compared with 780-790s on development.

I'm at a loss for an explanation as to the details of this.  It could be something as mundane as the time gained by losing all the calls to `epslevel` when determining test tolerances, though I must admit that a gain of 100s+ seems too large for it to be just that.  What's clear is that we didn't lose much, if anything, by computing `vscale

asgeirbirkis:
@aaustin141 could you please check whether you're happy with my commit 8b7dbee , I replaced a lot of `.*eps` with `*eps` in the tests, as it better reflects that we don't have to deal with vectors of epslevels anymore?

nickhale:
@asgeirbirkis : I haven't checked each and every line of those commits, but it seems a good idea. 

The only place it would cause an error would be if we had somewhere defined `eps` as a vector. Since we shouldn't be doing that anyway, that should be fine.

aaustin141:
@asgeirbirkis:  Just checked over the diffs.  They look OK to me!

asgeirbirkis:
OK, now I've similarly replaced all `eps.*` with `eps*` in both class methods and tests. Sorry this is taking such a long time to review, the diff is more than 18000 lines long, but I'm gradually working my way through it.

asgeirbirkis:
OK, gone through all the class methods, only the tests left. Have made some changes along the way to further simplify the code.

@aaustin141 here are some outstanding issues with the code, hopefully, I won't come across too many in the tests (have already fixed most of the issues I encountered there):

* In `chebtech/chebtech`, a TODO is left for `parseDataInputs`.

* The documentation for `@chebtech/parseDataInputs.m` is quite limited, I didn't really understand what's going on in the method and why it's necessary from reading it -- just an explanation of what the DATA struct actually is would be really helpful.

* `@chebtech/parseDataInputs.m` is missing copyright info.

* Same applies for the `trigtech/parseDataInputs` method.

* (I realise that those methods did exist before, but we should use this opportunity to improve the documentation.)

* In `chebtech/simplify`, we're always creating a `cheboppref`, even though `tol` got passed (it becomes important in line 43 to have `p` if `tol` didn't get passed). But in line 48, there is no way for a user to affecting the value of `p.eps`, except for changing the global `cheboppref`. Also, `tol` in line 48 doesn't have to be a row vector anymore, as all the entries will be identical. I suspect that the logic for this method can be simplified a lot.

* The same as above applies to `trigtech/simplify`.

* There are various places in the code where we construct a cheboppref or techpref, just to read out the current pref.eps. E.g. lines 41-42 in `chebtech/strictcheck`. I think I recall seeing this causing quite a bit of overhead in the profiler, so we might want to find a simpler way of just returning the pref.eps . However, this is more of a performance related issue, so should be dealt with separately (just wanted to bring this up now while it was fresh on my mind).

aaustin141:
I've just pushed some commits addressing these comments.

Regarding `chebtech.simplify()`, I've removed the use of p.eps from line 48, as doing so doesn't seem to break anything.  I've also moved the call to `chebtech.techPref()` inside the if-block for determining the default tolerance so we aren't wasting time creating them unnecessarily.  As for your remarks about the conversion of tol to a row vector, look again:  that's necessary in the event that you want to use a different simplification tolerance for each column.  I've added documentation clarifying that the `tol` input to the function may be a vector.

Similar changes have been applied to `trigtech.simplify()`.

> we might want to find a simpler way of just returning the pref.eps

I don't know a way to do this that wouldn't also either (a) violate the rule that `chebtech.techPref()` be the single-point-of-truth regarding the default value for the preference or (b) add input-parsing overhead to `chebtech.techPref()` that would offset (or maybe even totally erase) any performance gain.  Do you have anything specific in mind?  I agree that we don't want to try to address this issue in this pull request.

asgeirbirkis:
Thanks @aaustin141 for the recents commits and comments above, in particular, sorting out the correct calling sequence to the happinessChecker.

I don't have a concrete idea in mind on how to more cheaply get the `pref.eps`, perhaps something along the lines of putting the responsibility for each option in different method, but that might cause a different type of overhead and we'd end up potentially worse situation. In any case, it's not worth worrying about this now, just something we might try to address if profiler results indicate that it's something taking up nontrivial time.

asgeirbirkis:
Sry about 5b29cf1 (and having to revert with 1eb6fe6 ) , got too aggressive on code simplifications.

asgeirbirkis:
Image of Frodo

aaustin141:
Great!  Thank you @asgeirbirkis for all your hard work on looking at this!

nickhale:
Lol. Thanks @asgeirbirkiss!

PR #1588, opened 2015-09-02
Feature twelve transforms

nickhale:
The twelve flavours of Chebyshev-Legendre transforms requested by @trefethen .


trefethen:
Thanks @nickhale!  I'm delighted you've done this.  I'm on holiday this week and won't look properly, but my impression is you've now got all twelve codes with names like chebcoeffs2legcoeffs, and that these don't need to be prefaced by anything like cheb or chebfun.  Right?  If so I'm very pleased and look forward to playing with them.

nickhale:
That's correct - although this can readily be changed. (I have no strong feelings on the matter.)

Enjoy your holidays!

PR #1599, opened 2015-09-08
Eliminate calls to chebfun/subsref in treeVar/tree2infix output.

asgeirbirkis:
Another PR, which like #1341 was inspired by an example by @trefethen . By avoiding calls to `chebfun/subsref` in output from the `treeVar` class, nice speedups can be obtained. Take for example:
```
L = chebop(0,8); L.lbc = 0;
L.op = @(t,y) diff(y) + y;
t = chebfun('t',[0 8]);
% Smooth RHS
g = sin(t.^2); 
tic, 
for i =1:5
    y1 = L\g;
end
time = toc; fprintf('smooth: %4.3fs.\n', time)
% Piecewise RHS
g = sign(sin(t.^2));
tic
for i=1:5
    y2 = L\g;
end
time = toc; fprintf('piecewise: %4.2fs.\n', time)
```

On current release branch, I get the timings on my laptop on Matlab 2014b:
```
smooth: 3.482s.
piecewise: 27.03s.
```
On this branch, I get:
```
smooth: 2.515s.
piecewise: 20.27s.
```

On the 2015b prerelease, the timing results are also pretty nice (notice how much faster 2015b appears to be in general). On current release branch:
```
smooth: 1.404s.
piecewise: 10.03s.
```
On this branch:
```
smooth: 0.815s.
piecewise: 5.90s.
```

Profiling reveals to for this particular example, 639 calls to `chebfun/subsref` are saved for the smooth example, and 4453 for the piecewise one.

PR #1604, opened 2015-09-09
Simplification of code in chebfun/feval

### LINKED ISSUE #1600: Unnecessary call to chebfun/size in fevalColumn
Looking at the profiler output when studying #1599, I noticed a lot of calls to `chebfun/size` inside `chebfun/feval/fevalColumn`. These are in [line 85](https://github.com/chebfun/chebfun/blob/fix-chebop-shortPulses/%40chebfun/feval.m#L85).

However, we're only really interested in the result of `chebfun/size` with the second argument equal to 2, which is just what `numColumns` returns (and `numColumns` gets called inside `chebfun/size`). By changing line 85 from
```
numCols = size(f, 2);
```
to
```
numCols = numColumns(f);
```

and running the profiler for `chebtest('chebfun')`, it turns out that it saved more than 13000 calls to `chebfun/size`, while all the tests still passed.

I'm wondering whether there's a reason for this call to `size` rather than `numColumns` before I issue a PR addressing it.

I'm also wondering about the line that follows in `fevalColumn`:
```
numFuns = numel(f.funs);
```
Will this ever be anything else than 1, as things with `numel(f) > 1` only live at the `chebfun` level?

asgeirbirkis:
(This PR replaces #1603 )

* Call numColumns instead of size in columnFeval.

* Deal with row chebfuns in feval, rather than columnFeval.

This closes #1600. I didn't do a profiling on the whole of chebtest, but for and chebtest('chebfun'), it turns out that it saved more than 13000 calls to chebfun/size.

Also, for the example listed in #1599, these are the times I get on 2015b prerelease before and after these changes.

Before:

smooth: 0.820s.
piecewise: 5.62s.
After:

smooth: 0.716s.
piecewise: 4.93s.

### DIFF
--- a/@chebfun/feval.m
+++ b/@chebfun/feval.m
@@ -88,7 +88,6 @@ x = x(:);
 numCols = numColumns(f);
 numFuns = numel(f.funs);
 
-out = zeros(size(x, 1), numCols);
 funs = f.funs;
 dom = f.domain;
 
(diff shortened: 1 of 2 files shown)


asgeirbirkis:
With the current state of this branch (911a009), merge into `fix-treeVar-subsref`, the smooth example listed above goes from 0.716 seconds to 0.615s, compared with 1.6s on current development!

PR #1612, opened 2015-09-25
Fix chebop short pulses

### LINKED ISSUE #1512: IVP solver is wrong for short forcing pulses
There is something wrong with the IVP solver with certain forcing functions:

```
t = chebfun('t',[0 2]);
L = chebop(@(t) diff(t)+t,[0 2],1,[]);
u = L\0
```
Obviously, this is just exp(-t). 

If we put in a pulse for the forcing function:
```
>> pulse = 10*(t>1).*(t<1.1);
>> u_p = L \ pulse
u_p =
   chebfun column (3 smooth pieces)
[       0,       1]       10         1     0.37 
[       1,     1.1]        7      0.37      1.3 
[     1.1,       2]       10       1.3     0.52 

>> difference = norm(u - u_p)
difference =
    0.6394
```
The new solution checks out. But if the pulse duration is too short, the solution doesn't even see it.

```
>> pulse = 20*(t>1).*(t<1.05);
>> u_p = L \ pulse
u_p =
   chebfun column (3 smooth pieces)
[       0,       1]       10         1     0.37 
[       1,     1.1]        6      0.37     0.35 
[     1.1,       2]        9      0.35     0.14 
  Vscale = 1.  Total length = 25.

>> difference = norm(u - u_p)
difference =
     0
```

asgeirbirkis:
This closes #1512 . The trick is to restart the ODE solvers at each breakpoint encountered, which ensures that short pulses are detected. This PR introduces a new option for cheboppref, which allows turning off this new default behaviour.

A small experiment indicates that this will not cause a performance hit for piecewise problem, but on the contrary, improve timings slightly. Again, here is our favourite piecewise problem from #1599:
```
L = chebop(0,8); L.lbc = 0;
L.op = @(t,y) diff(y) + y;
t = chebfun('t',[0 8]);
% Smooth RHS
g = sin(t.^2); 
tic, 
for i =1:5
    y1 = L\g;
end
time = toc; fprintf('smooth: %4.3fs.\n', time)
% Piecewise RHS
g = sign(sin(t.^2));
tic
for i=1:5
    y2 = L\g;
end
time = toc; fprintf('piecewise: %4.2fs.\n', time)
```

On development, running 2015b, I get:
```
smooth: 0.599s.
piecewise: 4.41s.
```

On this branch, I get:
```
smooth: 0.602s.
piecewise: 4.06s.
```


@chebfun/constructODEsol.m
```diff
@@ -0,0 +1,61 @@
...
+% See also ODESET, ODE113, ODE15s, ODE45.
+
+% Copyright 2015 by The University of Oxford and The Chebfun Developers. 
+% See http://www.chebfun.org/ for Chebfun information.
+
+
+% If varargin{1} (which correspond to options) does not have a field called
+% RESETSOLVER, or options don't get passed in, this throws an error, which is
+% resolved in the catch statement to go to the default behaviour of resetting.
+try
```
nickhale: What's our policy on `try-catch` statements`?

@chebfun/constructODEsol.m
```diff
@@ -0,0 +1,61 @@
...
+    restartSolver = true;
+end
+
+% We don't want to reset the solver, or we just have one interval
+if ( ( length(tspan) == 2 ) || ~restartSolver )
+    sol = solver(odefun, tspan, uinit, varargin{:});
+    [t, y] = chebfun.odesol(sol, tspan, varargin{:});
+else
+    % Initialize a cell for storing the individual SOL pieces:
+    solCell = cell(1, length(tspan) - 1);
```
nickhale: Perhaps having an array of sols would be more elegant than having a cell array?

@chebfun/constructODEsol.m
aaustin141: As far as I'm concerned, `try` blocks are great tools for error/exception handling, but using one as a substitute for proper input-parsing logic like it's done here is not a good idea...

@chebfun/chebfun.m
```diff
@@ -599,6 +599,9 @@
         % Convert ODE solutions into CHEBFUN objects:
         [y, t] = odesol(sol, dom, opt);
         
+        % Call one of the MATLAB ODE solvers and return a CHEBFUN
```
nickhale: The order out outputs is switched here from `odesol` above. Not sure which is right.

@chebfun/constructODEsol.m
nickhale: I agree.

nickhale:
Aside from the few comments above (which I repeat below for clarity), I am happy with this. 

 * I don't think try-catch statements should be used for input parsing.
 * My preference would be for an array of structs (rather than a cell array) of sol objects.

asgeirbirkis:
Regarding array of sols vs a cell array, that was the first approach I took (i.e. I started implementing this with an array of sols). I changed the implementation to get a more elegant code by using cellfun which the cell array approach allowed, however, I think arrayfun might equally well be applied. The timings seemed to be very comparable, so happy to go for the array of sols if that does work out.

Regarding the try-catch logic, what inspired me there was this blogpost: http://undocumentedmatlab.com/blog/trapping-warnings-efficiently . I do admit though that it's not as transparent, so I'll do some timing experiments and assuming the penalty isn't outrageous, happy to change it from a try-catch to an if statement.

aaustin141:
> Regarding the try-catch logic, what inspired me there was this blogpost:  [LINK OMITTED]

That blog post is referring to using exceptions versus repeatedly checking a condition in a tight loop.  Is `constructODEsol()` called in a tight loop anywhere?  (I don't see such a place, but I only scanned quickly.)  If not, I doubt using a straightforward condition check instead would make much of a difference in performance at all.

Actually, even in the tight-loop case, it's possible for this to take a performance _hit_ using this "trick" if the condition being checked is simple.  Here's some test code, which is just a fleshed out version of the code from the post:
```matlab
function speedTest()
        Ntrials = 5;

        N  = 10000;
        A = zeros(10, 1);
        B = ones(10, 1);

        % Standard code.
        t = 0;
        tic();
        for (n = 1:1:Ntrials)
                for (loopIndex = 1:N)
                        refIndex = someCalculation();
                        if (refIndex > 0)
                                A(loopIndex) = B(refIndex);
                        else
                                A(loopIndex) = loopIndex;
                        end
                end
                t = t + toc();
        end
        disp('code 1 avg:  ')
        t / Ntrials

        % Faster code?
        t = 0;
        tic();
        for (n = 1:1:Ntrials)
                for (loopIndex = 1:N)
                        refIndex = someCalculation();
                        try
                                A(loopIndex) = B(refIndex);
                        catch
                                A(loopIndex) = loopIndex;
                        end
                end
                t = t + toc();
        end
        disp('code 2 avg:  ')
        t / Ntrials
end

function ind = someCalculation()
        ind = 1;
end
```
A typical run of this code on my machine (R2015b) produces results like the following:
```
code 1 avg:  
ans =
   0.002842200000000
code 2 avg:  
ans =
   0.030377400000000
```
The "standard" code is about 10 times faster here.  Note that my `someCalculation()` function always returns a valid index, and that's the case in which the `try-catch` code is supposed to be faster.  If I have it always return an invalid index instead, the `try-catch` code slows down by a further factor of 100 (though you would expect it to, since handling a thrown exception is expensive).  The results don't change if I remove the `someCalculation()` function and just assign `refIndex = 1` inline.

Of course, the guy writing the post isn't crazy.  This technique _can_ help sometimes.  Here's a modification which checks for the presence of a field in a struct instead:
```matlab
function speedTest2()
        Ntrials = 5;

        N  = 10000;
        A = struct('field', 0);

        % Standard code.
        t = 0;
        tic();
        for (n = 1:1:Ntrials)
                for (loopIndex = 1:N)
                        if (isfield(A, 'field'))
                                x = A.field;
                        else
                                x = 0;
                        end
                end
                t = t + toc();
        end
        disp('code 1 avg:  ')
        t / Ntrials

        % Faster code?
        t = 0;
        tic();
        for (n = 1:1:Ntrials)
                for (loopIndex = 1:N)
                        try
                                x = A.field;
                        catch
                                x = 0;
                        end
                end
                t = t + toc();
        end
        disp('code 2 avg:  ')
        t / Ntrials
end
```
For this code, I get times like these:
```
code 1 avg:  
ans =
   0.087914600000000
code 2 avg:  
ans =
   0.030052400000000
```
Now the `try-catch` code is 2-3 times faster, albeit less transparent.  Of course, this code heavily favors the `try-catch` case, since the exceptional path is never taken.  If I modify `A` such that it does not have the field to begin with, the `try-catch` code is considerably slower.

Which code is faster here depends on the frequency with which the exceptional case is likely to be hit.  Let's modify the above to
```matlab
function speedTest3()
        Ntrials = 5;

        N  = 10000;

        A = struct('field', 0);
        B = struct();

        % Standard code.
        t = 0;
        tic();
        for (n = 1:1:Ntrials)
                for (loopIndex = 1:N)
                        if (mod(loopIndex, 100) == 0)
                                S = B;
                        else
                                S = A;
                        end

                        if (isfield(S, 'field'))
                                x = S.field;
                        else
                                x = 0;
                        end
                end
                t = t + toc();
        end
        disp('code 1 avg:  ')
        t / Ntrials

        % Faster code?
        t = 0;
        tic();
        for (n = 1:1:Ntrials)
                for (loopIndex = 1:N)
                        if (mod(loopIndex, 100) == 0)
                                S = B;
                        else
                                S = A;
                        end

                        try
                                x = S.field;
                        catch
                                x = 0;
                        end
                end
                t = t + toc();
        end
        disp('code 2 avg:  ')
        t / Ntrials
end
```
The exceptional case gets hit 1% of the time.  Here are the timings:
```
code 1 avg:  
ans =
   0.101771600000000
code 2 avg:  
ans =
   0.112302400000000
```
So even an 1% exception rate is enough to put these codes on essentially equal footing.  As a control, here are the timings you get if you change the right-hand side of the checks involving `mod()` to -1, so that the exceptional case is never hit:
```
code 1 avg:  
ans =
   0.099048800000000
code 2 avg:  
ans =
   0.043447600000000
```

For more realistic codes, which case is more common can be difficult to assess.  For library codes like ours, it can depend how they get used by end users.  Even if the non-exceptional case is clearly overwhelmingly more likely, whether the use of a trick like this constitutes a _significant_ performance gain depends on how expensive the rest of the surrounding code is.  If this check is embedded in a loop for which the rest of the body takes 99% of the time, then even reducing the time of the check to 0 isn't going to make much of a difference.  Case-in-point:  I originally wrote `speedTest3()` using a subfunction to get `S` in each loop, but I found that the overhead of doing this erased any advantage of the `try-catch` code over the standard code, even when the exceptional case was never triggered.

In general, I really think it's better not to bother with micro-optimizations like this unless there's a demonstrated need for them, backed up by profiling.

asgeirbirkis:
Thanks @aaustin141 for this detailed investigation, you got me further convinced that it's not worth implementing the try-catch technique in our case (I'd already done a bit of experimentation and pushed a commit using an if statement instead). The original blogpost is from 2012, so it might also be that the if checking has become faster since the time of writing.

@chebfun/chebfun.m
asgeirbirkis: This seems to be all muddled up at the moment. The documentation at the top of `chebfun/odesol` reads
```
%   [Y, T] = ODESOL(SOL, DOM, OPT) returns also the linear CHEBFUN T on the
%   domain of Y. Note that the order of outputs is the reverse of that from
%   standard MATLAB calls to BVP4C(), ODE45(), etc.
```

However, inside the code, we've got
```
% Parse outputs:
if ( nargout > 1 )
    t = chebfun(@(t) t, y.domain);
    varargout = {t, y};
else
    varargout = {y};
end
```

So the code follows Matlab convention, while the documentation says it doesn't! I don't see a strong reason to go against the Matlab convention, especially as the code is already following it, so suggest to change the documentation, which I'd be happy to do.

asgeirbirkis:
OK @nickhale , I've addressed the issues you raised:

1. Got rid of the try-catch
2. Use array of sol structs instead of cell arrays
3. Updated documentation

@chebfun/odesol.m
```diff
@@ -12,25 +14,48 @@
...
 
 % Copyright 2015 by The University of Oxford and The Chebfun Developers. 
 % See http://www.chebfun.org/ for Chebfun information.
 
 %% Extract data from sol:
-vscale = max(abs(sol.y), [], 2); % Vertical scale (needed for RelTol)
-numCols = size(sol.y, 1);
+% Compute vertical scale (needed for RelTol)
+maxabs = @(sol) max(abs(sol.y),[],2);
+vscale = max(cell2mat(arrayfun(maxabs, sol,'uniformOutput',false)),[],2);
```
nickhale: This is very cryptic, and I'm not sure it's correct. (Seems it always returns a scalar?)

I'd prefer a loop in this case, for clarity:
```
colmax = @(sol) max(abs(sol.y), [], 2);
vscale = colmax(sol(1));                    % Initialise (vscale of 1st piece)
for k = 2:numel(sol)                        % Loop over the pieces
    vscale = max(vscale, colmax(sol(k)));
end
```

nickhale:
OK. I made one inline comment regrading `@chebfun/odesol.m` that needs looking at (I don't feel too strongly about it). Other than that, I'm happy to finally get this merged.

@chebfun/odesol.m
jrus: Why not just e.g.:
```matlab
vscale = 0;
for k = 1:numel(sol)                        % Loop over the pieces
    colmax = max(abs(sol(k).y), [], 2);
    vscale = max(vscale, colmax);
end
```

nickhale: @jrus - If you mean 
```
vscale = max(abs([sol(:).y]), [], 2);
```
then I agree.

jrus: I dunno exactly what the type is of this thing, and Matlab cell access syntax is still sorta confusing to me, I didn’t actually run that line in context, only experimented on little bits of data in the repl. You could be right: try it out. :-)

nickhale:
I made a few small changes. If @asgeirbirkis is happy with these then we can merge. (All tests pass.)

@chebfun/odesol.m
asgeirbirkis: @jrus Nice observation about the markdown Matlab syntax highlighting, wasn't aware of it, makes it look very nice, will make use of it in the future!

asgeirbirkis:
I like the new way suggested by @jrus for computing the `vscale` on `odesol`, so happy for this to be merged with the changes you made @nickhale .

PR #1613, opened 2015-09-25
fix-funqui-NaNs

aaustin141:
See commit messages for the details.  Thanks to @jaurentz for finding this.


jaurentz:
I fixed a small 0 division bug.

Issue #1610, opened 2015-09-21
failure of ATAP Ex. 8.4(c) and proposed change to `sampletest`


trefethen:
An impressive feature of Chebfun is that it does a pretty good job with functions involving 0/0 like sin(x)/x.  In exercise 8.4(c) of ATAP there's a function of this kind, log((x+3)/4)/(x-1), with a 0/0 singularity at x=1.  In v4 this worked, but now it doesn't:
```
>> chebfun(@(x) log((x+3)/4)./(x-1));
Warning: Function not resolved using 65537 pts. Have you tried 'splitting on'? 
> In chebfun.constructor>constructorNoSplit at 120
  In chebfun.constructor at 63
  In chebfun.chebfun>chebfun.chebfun at 219 
```
Yet if you construct with 20 or 200 or 2000 points, it looks fine, with a beautiful clean coefficient plateau.  @jaurentz has investigated and found that the problem is that whereas in v4 `sampletest` was basically a sanity test, checking an arbitrary point not on a grid, in v5 `sampletest` is a stringent indeed.  In particular it tests the points 1-1e-12 and -1+1e-12 with a tolerance O(1e-12).

I propose that we (@jaurentz and I) modify `sampletest` to bring it closer to what it used to be.  A practical reason is to make Chebfun again work on Exercise 8.4(c) and such examples.  But there is also a conceptual reason.  As people know, @jaurentz and I have spent a lot of the last year trying to make Chebfun construction simpler, and in particular, to put pretty much the whole construction process in a single place where we can explain it and think about it and reason about it (and discuss it in a publication, in preparation), namely `standardChop`.  The current `sampletest` violates this organization and thus is an aspect of confusing constructor complexity that still persists.  Note that the very presence of the numbers 1e-12 is a reflection of the kind of arbitrariness that we have been trying to reduce - why 1e-12?

I hope nobody will object if @jaurentz and I go ahead and prepare a modified and less stringent `sampletest`.  If anybody knows particular arguments why the v4 `sampletest` was not stringent enough, would you let us know?  Thanks.

aaustin141:
Now that #1614 has been merged, can this be closed?

nickhale:
Looks good to me.
```
>> chebfun(@(x) log((x+3)/4)./(x-1))
ans =
   chebfun column (1 smooth piece)
       interval       length     endpoint values  
[      -1,       1]       19      0.35     0.25 
```

Issue #1560, opened 2015-07-14
Singfun problem when testing orthogonality of chebpoly 


ajt60gaibb:
It seems that Chebyshev polynomials `T_k(x)` are not orthogonal in Chebfun.

```
x = chebfun( 'x' );
for k = 1:1000
    v(k+1)=sum( chebpoly(k)./ sqrt(1-x.^2) ); 
end
plot(v)
```
![orthogonality](https://cloud.githubusercontent.com/assets/1814063/8681473/8bed06ee-2a35-11e5-8a43-21fd0d2c8afe.png)

I assume this is an issue in `singfun/sum`. It looks like when `n` is of the form `1+k*32` there can be trouble. Compare 

```
find(abs(v)>1)
ans =
  Columns 1 through 11
     1    33    65    97   129   193   257   289   321   353   385
  Columns 12 through 22
   417   449   481   513   545   577   609   641   673   705   737
  Columns 23 through 30
   769   801   833   865   897   929   961   993

1+(0:31)*32
ans =
  Columns 1 through 11
     1    33    65    97   129   161   193   225   257   289   321
  Columns 12 through 22
   353   385   417   449   481   513   545   577   609   641   673
  Columns 23 through 32
   705   737   769   801   833   865   897   929   961   993
```

nickhale:
I think it's an aliasing issue rather than one of orthogonality. In particular, `singfun.constructSmoothPart()` explicitly disables `sampleTest` for some reason.

kuanxu:
@nickhale is right. It's an aliasing problem as `sampleTest` is turned off deliberately in `constructSmoothPart@singfun`. In the absence of `sampleTest`, the length of `chebpoly(k)./ sqrt(1-x.^2)` is 1 for all problematic k. Apparently, an aliasing issue.

Here, `sampleTest` is disabled simply because of, again, ill-conditioning, the origin of all `singfun` woes. Of course, I don't like it to be disabled. But I don't have a better option. If `sampleTest` is enabled by changing Line 15 in @singfun/constructSmoothPart.m to 
```
pref.techPrefs.sampleTest = 1;
```
you'll see that dozens of chebtests fail or hang forever. These are all due to rounding errors significantly amplified by the ill conditioning.

[...]

PR #1614, opened 2015-09-28
Feature simpler sampletest

jaurentz:
This pull request simplifies `sampleTest` for both `trigtechs` and `chebtechs`. See #1610.

### DIFF
--- a/@chebtech/populate.m
+++ b/@chebtech/populate.m
@@ -122,6 +122,10 @@ while ( 1 )
     vscl = max(vscl, max(abs(valuesTemp)));
     
     % Extrapolate out NaNs:
+    %
+    % TODO:  This will extrapolate out however many NaNs / Infs there are, even
+    % if that number is larger than the number of function values.  This is
+    % probably not good.
     [values, maskNaN, maskInf] = extrapolate(f, values);
 
     % Compute the Chebyshev coefficients:
--- a/@chebtech/sampleTest.m
+++ b/@chebtech/sampleTest.m
@@ -40,15 +40,8 @@ if ( isempty(vscl) )
 end
 tol = tol.*max(f.hscale*nrmf, vscl);
 
-% Choose a point to evaluate at:
-if ( n == 1 )
-    xeval = 0.61; % Pseudo-random test value
-else
-    % Test a point where the (finite difference) gradient of values is largest:
-    [ignored, index] = max(bsxfun(@rdivide, abs(diff(values)), diff(x)));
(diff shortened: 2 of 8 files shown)


nickhale:
What's the rationale behind evaluating at `xeval = 1e-5*[-1;1];` in the `chebtech` case, rather than some pseudo-random value?

asgeirbirkis:
Also, is there no way to unify the test for `chebtechs` and `trigtechs` and implement this method at the `smoothfun` level?

aaustin141:
@asgeirbirkis:  Let's leave that consideration out for the moment.  We know that lots of the code in `chebtech` and `trigtech` is duplicated and can be factored out, but I think it's best not to push for that right now.

jaurentz:
For `chebtech` I wanted 2 points that didn't show up on any of the grids and were in the least sampled part of the domain. I'm equally happy with pseudo random points though.

trefethen:
Thanks @jaurentz.  I'd be very glad if this could be merged this week, assuming no difficulties turn up, since I'm working on a paper about constructors etc.

nickhale:
I'd be in favour of pseudo random values. Might as well use the same ones as currently `trigTech`, i.e., 
```
% pseudo random sample points
xeval = [-0.357998918959666; 0.036785641195074];
```

Two other points:
 * I notice that the call `x = f.chebpts(n);` on line 19 is no longer required. As is the TODO at the start.
 * This change might remove the requirement for for `techprefs.sampleTestEps` (which I dislilked from the start). I see the desire to get this 'fixed' quickly, but I agree with @asgeirbirkis that it might be worthwhile unifying and solidifying `sampleTest` while it's in the L1 cache.

aaustin141:
Can confirm:  this appears to get rid of the need for `sampleTestTol` (see #1560) on my machine.

jaurentz:
I've update `chebtech/sampleTest` to use pseudo random values and have eliminated `sampleTestEps` from `chebtech` and `trigtech`.

nickhale:
Once #1584 is merged this could do with some tidying up, but I'm happy with it for now.

The only problem is that test 31 of `chebfun/test_norm()` fails. The test expects a `min(f, -inf)` to occur at x = inf, but it's now returning a finite (but very large) number. I suspect this test is ill-conditioned anyway, but perhaps someone else should take a look.

aaustin141:
@nickhale:  I've been seeing that test fail now for some time on my personal machine (but not on my workstation).  I think we fixed the issue somewhere in #1584, as it doesn't fail for me on that branch.

PR #1619, opened 2015-10-12
Use standardCheck() in PDE15S() and PDE23T().

nickhale:
Also add systems (i.e., `chebmatrix`) support for `pde23t()`.

All demos work, as does the extended `test/misc/test_pde15s()`.

PS. `pdeSolve()` wastes a lot of time building a `chebfun` after each successful time chunk. It might speed things up to simply store values/coefficients as we go a long, and do the `chebfun` construction at the end.


asgeirbirkis:
I think it would be great to make changes so that we'd only construct a chebfun at the end, we'd save a great amount of overheads. My only worry is what'd happen in the guimode, as we'd have to be more careful about plotting at each timestep (can't simply call plot on the chebfun, but should be able to generate the same effect by generating the plotdata from the coeffs at each timestep).

Issue #1608, opened 2015-09-17
get(fun,'lval') and get(fun,'rval')


asgeirbirkis:
This is inspired by #1607 .

By using @aaustin141 's code and turning the profiler on:
```
Ntrials = 5;
t_total = 0;
profile on
chebfun(@(x) sign(x), 'trunc', 100);
for (k = 1:1:Ntrials)
        tic();
        chebfun(@(x) sign(x), 'trunc', 100);
        t = toc()
        t_total = t_total + t;
end
profile viewer
avg_time = t_total/Ntrials
```
reveals that the most self time is spent in `chebtech2.get`, in particular in the line `case fieldnames(f)` (the design of our `get` methods is material for its own issue, and it probably would be much more efficient to do a try-catch in the method -- fieldnames is an expensive operation).

Further investigating why we call `chebtech2.get` so often, it's evidently because of lines 30-35 in `chebfun.getValuesAtBreakpoints`:
```
    vals(1,:) = get(funs{1}, 'lval');
    % Take the mean of the FUNs on either side of the break:
    for k = 2:numFuns
        vals(k,:) = (get(funs{k-1}, 'rval') + get(funs{k}, 'lval'))/2;
    end
    vals(numFuns+1,:) = get(funs{numFuns}, 'rval');
```

But this is a bit crazy, because what happens at the bottom level in `chebtech2.get` is simply
```
    case 'lval'
        out = lval(f);
    case 'rval'
        out = rval(f);
```
So isn't it a bit weird not just to have `lval` and `rval` at the higher levels, and save all these `get` calls?

nickhale:
This is weird. When I profile this (also on R2015b) I spend very little time in `get`. Most of my time is spent in `chebfun/chebcoeffs()` and in particular `chebfun.overlap()`. I can imprive this somewhat by changing lines 86-92 of `chebfun/chebcoeffs` to
```
    % Compute coefficients via inner products.
    d = f.domain;
    x = chebfun('x', d);
    if ( kind == 1 )
        w = 1./sqrt((x - d(1)).*(d(end) - x));
    elseif ( kind == 2 )
        w = sqrt((x - d(1)).*(d(end) - x));
    end
```
(Note that this only works as `chebfun/chebcoeffs' doesn't support quasimatrices anyway.)

[...]

PR #1625, opened 2015-10-20
Feature lval rval speed

nickhale:
See #1608 for discussion.


asgeirbirkis:
@nickhale I added couple of small speedup fixes in c1851d2 . Let me know if you're happy with those to be included in the PR as well (all tests still pass), and I'll go ahead and merge it.

PR #1627, opened 2015-10-26
Turn off default plotting if pde15s is called with outputs.

asgeirbirkis:
If pde15s is called without any outputs, intermediate solutions still
get plotted by default. This is similar to the default behaviour of the
MATLAB ODE solvers, which plot the marching through time if they're called
without outputs. This change is inspired by a discussion with @trefethen
and the use of pde15s in the upcoming book.


trefethen:
Thanks.

PR #1628, opened 2015-10-26
updated comments to clarify algorithm

jaurentz:
As per Asgeir's request, I've updated the documentation in `@chebtech/simplify.m` and `@trigtech/simplify.m`.

### DIFF
--- a/@chebtech/simplify.m
+++ b/@chebtech/simplify.m
@@ -25,15 +25,19 @@ if ( ~f.ishappy )
     return
 end
 
-% STANDARDCHOP requires at least 17 coefficients, so for F such that LENGTH(F) <
-% 17, the coefficients are padded using prolong. The following
-% parameters are chosen explicitly to work with STANDARDCHOP.
+% STANDARDCHOP requires at least 17 coefficients to avoid outright rejection.
+% STANDARDCHOP also employs a look ahead feature for detecting plateaus. For F
+% with insufficient length the coefficients are padded using prolong. The
+% following parameters are chosen explicitly to work with STANDARDCHOP. 
 % See STANDARDCHOP for details.
 nold = length(f);
 N = max(17, round(nold*1.25 + 5));
 f = prolong(f,N);
 
-% Grab coefficients of F.
+% After the coefficients of F have been padded with zeros an artificial plateau
+% is created using the noisy output from the FFT. The slightly noisy plateau is
+% required since STANDARDCHOP uses logarithms to detect plateaus and this has
+% undesirable effects when the plateau is made up of all zeros.
 coeffs = f.coeffs;
 [n, m] = size(coeffs);
 coeffs = chebtech2.vals2coeffs(chebtech2.coeffs2vals(coeffs));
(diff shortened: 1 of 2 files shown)


@chebtech/simplify.m
```diff
@@ -25,15 +25,19 @@
...
+% STANDARDCHOP also employs a look ahead feature for detecting plateaus. For F
+% with insufficient length the coefficients are padded using prolong. The
+% following parameters are chosen explicitly to work with STANDARDCHOP. 
 % See STANDARDCHOP for details.
 nold = length(f);
 N = max(17, round(nold*1.25 + 5));
 f = prolong(f,N);
 
-% Grab coefficients of F.
+% After the coefficients of F have been padded with zeros an artificial plateau
```
nickhale: Is this artificial plateau always required, or only when there has been zero padding of the coefficients?

jaurentz:
It's required whenever there's a zero padding which there always will be.
```matlab
N = max(17,round(1.25*nold + 5));
```

nickhale:
Is there another way to model this noise? 

I realise that in practice, with all the overheads we have elsewhere, it probably makes very little difference, but taking two FFTs here irks me!

asgeirbirkis:
I spotted those two FFTs when profiling related to PR #1625, running the code at the start of issue #1608 . The FFTs are now towards the top of the list in terms of running costs, see attached screenshot of profiler results (lines 3 and 6)
screen shot 2015-10-26 at 13 40 30
.

jaurentz:
We originally started by setting an artificial plateau based on certain parameter values. It became a bit of an engineering problem and ultimately we didn't like how complicated it had become. The goal is for the chopping procedure to simple to implement and explain. We do not claim that the FFT approach is optimal but it definitely satisfies the simplicity requirement. @trefethen and I are currently writing everything up and hopefully we will think up something more clever as we collect our thoughts.

jrus:
For a reader who doesn’t intimately know how the parts fit together, I think this comment is a bit unclear:
```matlab
% After the coefficients of F have been padded with zeros an artificial plateau
% is created using the noisy output from the FFT. The slightly noisy plateau is
% required since STANDARDCHOP uses logarithms to detect plateaus and this has
% undesirable effects when the plateau is made up of all zeros.
```

Zero padding is already a sort of “artificial plateau”, and what are “undesirable effects”?

Maybe something like:
```matlab
% The artificial plateau extension created by padding with zeros is unsuitable for
% STANDARDCHOP which uses logarithms to detect plateaus, and therefore
% breaks when the logarithm is undefined. To avoid pure zeros, perform a forward
% and inverse FFT, relying on the noise added by tiny floating point rounding errors.
```

* * *

What would happen if you padded with e.g. alternating ± `realmin`, instead of all zeros?

jrus:
One slightly related thing, but a bit more general a question: have you folks tried making sure to expand to a power of two when undertaking this kind of operation? In general the FFT is dramatically faster for power-of-two sizes compared to arbitrary sizes. Especially running an FFT on a large prime size is going to result in a considerably more expensive operation.

PR #1631, opened 2015-10-27
Fix chebop/feval to vertcat outputs which would otherwise return cell arrays

nickhale:
This allows construction of operators such as

```matlab
function foo
N = chebop(@(r,u) odefun(r,u), [-1 1]);
N.lbc = @(u) [u{1} - 1 ; u{2}];
N.rbc = @(u) [u{1} ; u{2} - 1];
N.numVars = 2; % <-- Why is this is required here?
x = chebfun('x');
u = N\0
plot(u), shg

    function op = odefun(r, u)
        for n = 1:2
            op{n} = diff(u{n}, 2) + n*u{n};
        end
    end

end
```

### DIFF
--- a/@chebop/feval.m
+++ b/@chebop/feval.m
@@ -188,7 +188,7 @@ elseif ( numberOfInputs == 2 )
 
     % Evaluate the operator!
     out = N.op(x, u);
-
+    
 else
     % The operator is specified on the form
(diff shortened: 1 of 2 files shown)


asgeirbirkis:
@nickhale I converted your example above to a new test that I just pushed, running chebtest on this branch and assuming it passes, will then merge!

PR #1632, opened 2015-10-28
Save a lot of overheads when constructing piecewise operators

asgeirbirkis:
These are the next steps to addressing the slow piecewise BVP first encountered in #1485:

```
tic
dom = [0 16];
x = chebfun('x',dom);
f = (x-floor(x))<sqrt(2)/7;
L = chebop(dom);
L.lbc = 0; L.rbc = 1;
L.op = @(x,u) diff(u,2);
tic, u = L\f;
toc
```

Before #1485 got merged, this took 12.93 seconds on my computer on development, and the merge brought it down to 4.5 seconds.

The code now takes at around 3.8 seconds on my laptop on development, but on this branch, it's down to 1.3 seconds. This is achieved by constructing the pieces of piecewise operators in a way that saves a lot of Matlab overhead, as can be seen by comparing the two profiler results before and after.

On development:
screen shot 2015-10-28 at 16 11 36

On branch
screen shot 2015-10-28 at 16 12 25

The most expensive operation now is blkdiag, which indicates that further improvements might be achieved by making more use of sparse matrices -- spying the operator that results from the blkdiag call in `chebcolloc/diff` certainly supports that idea. However, making more efficient use of sparse matrices for ODEs is a matter for another PR.
dspy

### DIFF
--- a/@chebcolloc/diff.m
+++ b/@chebcolloc/diff.m
@@ -9,35 +9,37 @@ function D = diff(disc, m)
 %  Copyright 2015 by The University of Oxford and The Chebfun Developers.
 %  See http://www.chebfun.org/ for Chebfun information.
 
-% Store information about domain and dimensions.
-d = disc.domain;
-n = disc.dimension;
-
-if m == 0
+if ( m == 0 )
     % Trivial case:
-    D = eye(sum(n));
-else
-    numIntervals = disc.numIntervals;
+    D = eye(sum(disc.dimension));
     
+else
     % Find the diagonal blocks.
-    blocks = cell(numIntervals);
+    
+    % Store information about domain and dimensions.
+    numIntervals = disc.numIntervals;   
+    n = disc.dimension;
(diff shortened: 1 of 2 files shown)


nickhale:
I've made a few small cosmetic changes, and I'm running the tests now. If these pass then I'll merge.

A small question: For a while in V4 we would cache the diffmats. Is it worth reconsidering this? Recall also that D1 is produced as a by-product of computing D2 in the chebcolloc basis.

nickhale:
@asgeirbirkis - please confirm you are happy with my changes, then I'll merge.

asgeirbirkis:
@nickhale Thanks for making the changes, they look good to me, so please merge!

Regarding caching, we could probably achieve that quite easily by declaring the `Dn` cell-array variable in `chebcolloc.diff` as a persistent variable. However, I haven't looked at all into what kind of performance effect that would have.

Issue #1249, opened 2014-11-05
Chebfun2 can overestimate numerical degrees


ajt60gaibb:
Chebfun2 can get the overestimate the numerical degree of a function.

For example,  
```
f = chebfun(@(x) 1./(1+25*x.^2)); length(f)
ans = 181
f2 = chebfun2(@(x,y) 1./(1+25*x.^2)); length(f2.rows)
ans = 257
```
That's odd, but much odder is that in the y direction, 
where there is no variation at all, we get length 17 instead of 1:
```
f2 = chebfun2(@(x,y) 1./(1+25*x.^2)); length(f2.cols)
```
This may have something to do with `epslevels` because chebfun2 relies on chebfun to reduce the length of the row and column slices.

aaustin141:
There's something other than `epslevel` that's at work here:  these oddities arise even in the world of #1584.

ajt60gaibb:
@bhashemi, do you have an opinion about this?

bhashemi:
I think the desicion which was made and is responsible for these oddities is that: "Chebfun2 should not `simplify` rows and columns at the end of its construction process" which in turn was made to keep Chebfun2s more accurate for some examples. Here is what we see if we do `simplify`:
`>> f3 = simplify(chebfun2(@(x,y) 1./(1+25*x.^2))); length(f3.rows), length(f3.cols)
ans =
   185
ans =
     1
`
Note that the smallest grid used in Chebfun2 constructor is 17*17 as set in `minSamples` in `chebfunpref.mergeTechPrefs` and without `simplify` the length will remain at-least 17 in both variables.

ajt60gaibb:
Maybe Chebfun2 should simplify somehow at the end of construction?   

I cannot remember the examples that are more accurate without simplify, but I do remember seeing that.  I am not sure if it is still true though.

trefethen:
My view in general is in favor of simplification.  Chebfun (1,2,3,...) is based on a simple story: we construct a function and then round it to 16-digits.  In my view that story should capture the truth except when there are compelling reasons to make an exception.

ajt60gaibb:
I agree. I do not believe there is a compelling reason for an exception here. Though, @bhashemi has more experience with the improved version of the Chebfun2 constructor than me. 

I cannot remember the decision: "Chebfun2 should not `simplify` rows and columns at the end of its construction process"

bhashemi:
[...]

PR #1640, opened 2015-11-05
simplify Chebfun2s in the constructor

bhashemi:
This branch adds a “simplify” step at the very end of Chebfun2 constructor
so that it does not overestimate the numerical degrees of the function.
This will address issue #1249.


bhashemi:
Some of the tests fail in ``chebfun2/diff`` on this branch. I close this branch to find out what is happening here.

PR #1644, opened 2015-11-09
fix-chebfun2-degree

bhashemi:
This PR is supposed to address #1249, i.e., to avoid overestimation in the numerical degrees of chebfun2 objects.


nickhale:
I get the following warning appearing on the new branch
```
>> git checkout development
>> chebfun2(@(x,y)-x.*sin(sqrt(abs(x)))-y.*sin(sqrt(abs(y))), [0 1 0 1])
ans =
   chebfun2 object: (1 smooth surface)
[   0,   1] x [   0,   1]        2     [1.2e-15 -0.84 -0.84 -1.7]
vertical scale = 1.7 
```

```
>> git checkout fix-chebfun2-degree
>> chebfun2(@(x,y)-x.*sin(sqrt(abs(x)))-y.*sin(sqrt(abs(y))), [0 1 0 1])
Warning: Not a low-rank function. 
> In chebfun2/constructor (line 309)
  In chebfun2 (line 57) 
ans =
   chebfun2 object: (1 smooth surface)
[   0,   1] x [   0,   1]        2     [4.8e-15 -0.84 -0.84 -1.7]
vertical scale = 1.7

**Comment by bhashemi** [comment]
For this test function, in the development branch where chopping is not applied to ``rows`` and ``cols``, ``sampleTest`` passes when the ``grid`` size is 2049 and ``tol`` is 3e-15. 
In contrast, in this branch, where we chop ``rows`` and ``cols``, we see that ``sampleTest`` is not passed with the above-mentioned ``grid`` size and ``tol``. It passes only after the ``grid`` size goes up to 4097 and ``tol`` is therefore increased to 9e-14. There are two points therefore: 

1) Simplifying ``rows`` and ``cols`` has the potential to decrease the accuracy of chebfun2s and we might want to discuss this, in principle, further.

2 ) The warning message ``Not a low-rank function`` is not informative in this case as it is issued just because the ``grid `` size was bigger than 2049. This is only a rank-2 function that is a little hard-to-resolve in x and y.

Any comments?

**Comment by jaurentz** [comment]
@trefethen, @bhashemi and I have discussed this further and decided to do the following:

1) Put `sampleTest` before `simplify` in `chebfun2` just as in 1D `chebfun`.

2) Discuss at leisure whether this is really the best ordering for these operations. This is all part of the     'standardChop' discussion, on which @trefethen and I are currently drafting a paper.

**Comment by trefethen** [comment]
This is on hold pending resolution of #1666.

**Comment by trefethen** [comment]
When all this is working, `plotcoeffs` and `explain` for a chebfun2 will look the way chebfuns should look in the post-standardChop era.  Here are two figures to illustrate.  Suppose we execute
```
f = chebfun2(@(x,y) cos(pi*x.*y)); plotcoeffs(f); close
```
In development, we get the first figure below, which is bad.  In fix-chebfun2-degree, we get the second figure below, which is good.  Similarly if you run `explain(f,4)`.
bad
good

jaurentz:
@ajt60gaibb,
The addition of a `simplify` call in the `chebfun2` constructor does give cleaner looking coefficient plots, however I am wondering if this could lead to any undesirable behavior. For example `chol` now gives factors that are slightly less accurate (though `R'*R` is still accurate). Thoughts?

ajt60gaibb:
I agree that the chebfun2 slice degrees need to be simplified.  If all the tests are passing, then that suggests `simplify` at the end is a valid option. I do not know an example where there is undesirable behavior.  There should have always been a `simplify` somewhere and I think it got forgot with the rewrite to version 5. 

For the factorization of matrices, the matrix factors are usually not as accurate as the residual so the fact that `R` is a little less accurate now is not a huge worry (for me).

PR #1659, opened 2015-11-17
Feature function iteration

### LINKED ISSUE #1437: Passing an initial guess with correct dimensions to chebop
The documentation for chebop states that if the system doesn't correctly identify which variable of an op is a parameter, you can manually specify one by passing an initial guess with the correct dimensions of the solution. However, doing so doesn't seem to work:

```
% Gives an error, as expected:
L = chebop(@(x,u,c) u, [-1 1], @(x,u,c) u(0)-c);
f = chebfun(@exp);
u = L \ [f; 0];

% Gives a different error, but should work:
L.init = [chebfun('x'); 0];
u = L \ [f; 0];
```

See also https://github.com/chebfun/chebfun/issues/1209#issuecomment-63653265.

asgeirbirkis:
This allows solving "zeroth order differential equations", e.g.
```
% Find u and v on [-1, 1] s.t. u^2 - x + sin(v) = v + exp(v) + u = sin(4x)+2
x = chebfun('x');
f = sin(4*x) + 2;
N = chebop(@(x,u,v) [u.^2 - x + sin(v); v + exp(v) + u]);
[u, v] = N\[f; f];
norm(N(u,v)-[f;f])
plot([u v])
```
giving the following result:
```
ans =
   1.6699e-13
```
uvfunctionplot

Also, this closes #1437 .

### DIFF
--- a/@chebop/chebop.m
+++ b/@chebop/chebop.m
@@ -144,7 +144,9 @@ classdef (InferiorClasses = {?double}) chebop
 % without the need to introduce extra equations into the system. Simply add the
 % unknown parameters in the list of arguments to the operator. By default, any
 % variable that is not acted on by differentiation or integration is treated as
-% a parameter, rather than a function.
+% a parameter, rather than a function. The exception is when no variable
+% appearing in a problem is differentiated or integrated, in which case, all
+% variables get treated as functions, rather than parameters.
 %
 % Example (unknown parameter in differential equation):
 %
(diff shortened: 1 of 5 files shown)


nickhale:
Neat. Seems this turned out to be pretty easy to do?

One problem, `chebop/test_linearizationDimensions` fails for me, but I think the expected behaviour has changed, so it's just a case of changing the test.

asgeirbirkis:
Ahh sorry, my bad, pushed a fix in 5ff9c9b.

Issue #1648, opened 2015-11-12
Computing PDE solns without caring how big TMAX is


trefethen:
An matter has arisen in `pde15s/chebgui` calculations that I believe is important to making Chebfun exciting and impressive for users.  (The same matter arises in @Hadrien-Montanelli's stiff PDE solvers, though this issue is officially tied to `pde15s/chebgui`.)  Users love to be able to start a signal running and watch it go for a long time -- effectively forever.  For example, it should be possible to solve the Kuramoto-Sivashinsky equation on a time interval of thousands, and just happily watch the chaotic solution evolve as you talk to your friends about the weather or the latest James Bond movie.

Currently in the default `pde15s/chebgui` modes, however, this is not possible, because solutions are stored all along.  Moroever even if you solve on a modest-length integral, at the end you have to wait often a long time for a surface plot to appear.

I hope @asgeirbirkis can rejig things so that by default at least, we don't save solutions along the way and we don't attempt a waterfall plot at the end.  

This issue is important for the V5.3 release, since one thing we'd like to publicize with the release is that our PDE solutions have speeded up.

nickhale:
I think the only way to do this is by adding an option to `pdeSet()`. If we did this, I'd be against it being the default mode, as that would be inconsistent with MATLAB's ODE solvers. Perhaps it would be OK if only `[t_init, t_final]` are passed as the time interval, but in that case we default to only 51 time steps anyway.

trefethen:
Let me counterargue!  Throughout scientific computing, it is standard to store the time-histories of ODE solutions and quite exceptional to store the time-histories of PDE solutions, because they're just too darned big.  Sure enough, storing big time-histories in Chebfun is messing up our demos consistently, taking away much of the fun of solving PDEs.  Half the PDE demos in CHEBGUI lead to an embarrassing wait for the waterfall plot at the end, and most of the time the viewer is probably doing it for the movie, not the waterfall plot anyway.  I am eager to publicize fast PDE solving with V5.3, but I won't feel comfortable publicizing this so long as this problem persists.

jrus:
If the solution evolves over time in a continuous way, wouldn’t it be possible in theory to store the time-history as a chebfun of one dimension higher? [This might take some trickery given uniformly spaced time steps, of course.]

asgeirbirkis:
OK, I've done some experiments, and here are my observations. It'd be good to reach a consensus about our plan of attack before implementing anything:

[...]

PR #1660, opened 2015-11-18
Don't specify linewidth for waterfall plots.

asgeirbirkis:
This leads to significant speed improvements.

See discussions in #1648, although this doesn't deal with the storing of PDE solutions (should be done on a separate branch).

### DIFF
--- a/@chebgui/solveGUIpde.m
+++ b/@chebgui/solveGUIpde.m
@@ -239,8 +239,7 @@ else
 end
 
 if ( ~isa(u, 'chebmatrix') )
-    waterfall(u, t, 'linewidth', defaultLineWidth)
-
+    waterfall(u, t)
 else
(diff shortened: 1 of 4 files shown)


trefethen:
@asgeirbirkis just demo'd this to me and the speedup is striking.

nickhale:
I can't reproduce this:

```
>> tic, waterfall(w, t), toc
Elapsed time is 0.023053 seconds.
>> tic, waterfall(w, t, 'linewidth', 2), toc
Elapsed time is 0.023265 seconds.
```

asgeirbirkis:
A tic, toc didn't reveal it for me either (which is why it was tricky to track down), the toc prints before the figure actually finishes rendering. @nickhale if you solve something like the Allen-Kahn in chebgui, do you not notice a change in the time it takes the waterfall plot to show up with or without the linewidth argument?

Issue #1652, opened 2015-11-16
Can we turn off aliasing?


jaurentz:
Currently after constructing a sequence of coefficients and choosing a chopping point, Chebfun aliases the sequence.  This has troubled us in writing the standardChop paper because, of course, we have to explain there why this aliasing is done.

We've done some experiments and we conjecture that it is done for no well-established reason.  We would like to propose that the aliasing step could be removed.  This would make our construction process simpler and easier to describe.  Can anyone point to some evidence that it's important to keep aliasing?

To illustrate, here are two figures.  In each case we've constructed a chebfun with and without aliasing and plotted the error at a collection of points on [-1,1].  In one case (probably typical, but we've only done half a dozen functions so far) it seems to make little difference.  In the other case, which we picked knowing that this was a function whose coefficients were subject to significant aliasing effects, we find to our surprise that aliasing seems to make the chebfun worse. Each plot below shows the error at `50*n` Chebyshev points, where `n` is the length of the function constructed using the current version of `development`. Both the aliased and non-aliased versions have the same degree.
![aliasing_exp_function](https://cloud.githubusercontent.com/assets/7694954/11179970/0e51b4ce-8c4f-11e5-898f-fa4834aaedf2.png)
![aliasing_log_function](https://cloud.githubusercontent.com/assets/7694954/11179971/0e6ee422-8c4f-11e5-87f1-14d3420d8e9e.png)

nickhale:
Yes.

trefethen:
@jaurentz and I have done some more experiments.  The trend seems to be that with a fast-converging series aliasing makes little difference, while with a slow-converging series it increases the errors by something like a factor of 2.  This is exactly what you'd expect from general theory, e.g. Chaps. 4,7,8 of <i>Approximation Theory and Approximation Practice</i>, such as the comment at the top of p. 30.

Accordingly we are now preparing a branch that turns off aliasing, with a view to making this Chebfun's default unless problems turn up.

ajt60gaibb:
I do not have much to say. 

Pedro Gonnet told us in a Chebfun meeting that aliasing improved the accuracy of the roots command by 1 or 2 digits.  I am not sure if that observation still holds.

asgeirbirkis:
I'm presuming this was closed with #1664?

PR #1664, opened 2015-11-19
Feature remove aliasing

jaurentz:
This pull request attempts to turn of aliasing in construction and prolongation. See #1652.


jaurentz:
As an update to this pull request I've also loosened the tolerances in `chebtech` and `trigtech` `sampleTest` to avoid unnecessarily large interpolation grids.

PR #1667, opened 2015-11-20
Set x and y lims in chebop/quiver, make plots much nicer.

asgeirbirkis:
While working on an example for `chebop/quiver`, I realised this would make the plots look much nicer, would be great if someone could quickly merge this minor fix (two lines of code).


asgeirbirkis:
Thanks @Hadrien-Montanelli for the quick response!

PR #1677, opened 2015-11-25
changed base loop index from 1 to 2 in standardChop

jaurentz:
This PR changes the base index in the plateau computation loop in `standardChop`.


trefethen:
This should be a no-op change that makes this code at the heart of Chebfun more logical to anyone who reads it.   I proposed the change and I support it.

PR #1678, opened 2015-11-26
Treat addition of zero functions as a special case to avoid warnings in singfun/plus.

### LINKED ISSUE #1423: sqrt(x) - 0*x fails
```
>> x = cheb.x;
>> sqrt(x) - 0*x
Warning: Non-integer difference in the exponents of the two SINGFUN objects: 
The result may not be accurate. 
> In singfun.plus at 126
  In classicfun.plus at 37
  In chebfun.plus at 107
  In chebfun.minus at 11 
Warning: Non-integer difference in the exponents of the two SINGFUN objects: 
The result may not be accurate. 
> In singfun.plus at 126
  In classicfun.plus at 37
  In chebfun.plus at 107
  In chebfun.minus at 11 
ans =
[      -1,       0]    65537    complex values 
[       0,       1]    65537     2e-14        1 
Epslevel = 1.852400e-10.  Vscale = 1.000000e+00.  Total length = 131074.
```

nickhale:
Closes #1423.

We now get:
```
>> cheb.x;
>> sqrt(x) + 0*x
ans =
       interval       length     endpoint values   endpoint exponents
[      -1,       0]        1    complex values         [0      0.5]  
[       0,       1]        1         0        1         [0.5      0]  
vertical scale =   1    Total length = 2
```

### DIFF
--- a/@singfun/plus.m
+++ b/@singfun/plus.m
@@ -137,16 +137,9 @@ else
         ['Non-integer difference in the exponents of the two SINGFUN ' ...
         'objects: The result may not be accurate.']);
     
-    % Define a function handle for the sum:
-    op = @(x) feval(f, x) + feval(g, x);
-        
-    % Take the smallest exponents to be those for the summation:
-    exps = [get(f, 'exponents'); get(g, 'exponents')];
-    exps = min(exps);
-    
     % Construct a new SINGFUN for the sum:
-    data.exponents = exps;
-    s = singfun(op, data, []);
+    op = @(x) feval(f, x) + feval(g, x);
+    s = singfun(op, [], []);
 end
 
 %%
--- a/@singfun/plus.m
+++ b/@singfun/plus.m
@@ -137,9 +137,16 @@ else
         ['Non-integer difference in the exponents of the two SINGFUN ' ...
         'objects: The result may not be accurate.']);
     
-    % Construct a new SINGFUN for the sum:
+    % Define a function handle for the sum:
     op = @(x) feval(f, x) + feval(g, x);
-    s = singfun(op, [], []);
+        
+    % Take the smallest exponents to be those for the summation:
+    exps = [get(f, 'exponents'); get(g, 'exponents')];
+    exps = min(exps);
+    
+    % Construct a new SINGFUN for the sum:
+    data.exponents = exps;
+    s = singfun(op, data, []);
 end
 
 %%
--- a/@singfun/plus.m
+++ b/@singfun/plus.m
@@ -122,6 +122,16 @@ elseif ( all(abs(round(fExps - gExps) - (fExps - gExps)) < tolExps) )
 else
     % Case 3: Nontrivial difference in the exponents of F and G. Form a new
     % function handle for the sum from F and G.
+
+    % It's worth looking for the special case when one of f or g is actually a
+    % zero funtion here. See #1423.
+    if ( iszero(f) )
+        s = g;
+        return
+    elseif ( iszero(g) )
+        s = f;
+        return
+    end
     
     warning('CHEBFUN:SINGFUN:plus:exponentDiff', ...
         ['Non-integer difference in the exponents of the two SINGFUN ' ...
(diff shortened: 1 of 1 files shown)


aaustin141:
Rather than special-casing the addition of the zero function, would it not be better just to let `singfun` auto-detect the exponents when doing these additions?  This failure surprised me because we have to call the constructor for this operation, and constructing `sqrt(x)` on both `[-1, 0]` and `[0, 1]` with auto-detection of the exponents succeeds.  I therefore concluded that the constructor must not be auto-detecting the exponents when we do this addition, and indeed, it seems that we set them using [these lines](https://github.com/chebfun/chebfun/blob/19c943b733139aa1ebbac4e520dc2fb3bbfc64dc/%40singfun/plus.m#L136-L138) (added in 1b180c71) instead.  The result these lines give is obviously wrong in this case (zero exponents) and is probably wrong in others.

Note that we used to do auto-detection in this case previously and that resuming the use of auto-detection by commenting out [this line](https://github.com/chebfun/chebfun/blob/19c943b733139aa1ebbac4e520dc2fb3bbfc64dc/%40singfun/plus.m#L141) doesn't cause any tests for `chebfun` or `singfun` to fail.  (Here "used to" means "prior to the v5 release"; see, e.g., commit d6ad2639.)

@kuanxu:  Do you have any comments regarding why those were introduced?  The commit message for 1b180c71 doesn't say much.

nickhale:
My preference is typically to avoid calling the constructor if possible, particularly where `singfun()` and exponents are involved. This is possibly not quite so important following the increased robustness from `standardChop()`, but checking for a zero function in this case is such an easy thing to do that I think it's the better solution.

aaustin141:
I generally agree that not calling the constructor when `singfun` is in the picture is advisable, but in the case where you're adding two singfuns with non-integer differences in the exponents, a construction is inevitable, isn't it?  (That is, unless you're trapping cases like these.)  If anything, addition of the zero function should be a case in which a construction is totally safe, since it doesn't change anything!

Even if we decide not to go back to doing a construction with auto-detected exponents here, there's still something wrong with [these lines](https://github.com/chebfun/chebfun/blob/19c943b733139aa1ebbac4e520dc2fb3bbfc64dc/%40singfun/plus.m#L136-L138) that needs fixing.

nickhale:
Those lines aren't necessarily wrong; if the exponents don't differ by an integer, then the construction will almost certainly be unsuccessful. Probably the only case where it will/might be successful is when one of the functions is zero, which I'd like to treat as a special case anyway.

Therefore choosing the minimum of the two exponents is probably as good a heuristic as any (although, now that I think of it, taking the mean of their fractional parts might be better..).

aaustin141:
Is there any data to support that heuristic, though?  I agree that the construction is likely to fail in most cases, but in the cases where it should succeed, I don't think that trying to heuristically guess the exponent is more likely to result in success than just trying to auto-detect it.  In the case at hand, it clearly does the wrong thing:  the minimum of 0.5 and 0 is 0, so the constructor gets passed zero exponents.  If it had been allowed to auto-detect, it would have gotten the correct answer!

nickhale:
I think my point is that it's only when you add a zero function that you can get the correct answer. In all other cases you'll have a `1 + sqrt(x)` type scenario. Searching for an exponent when you know you're not going to find the right one (because there isn't one) seems a waste of time.

aaustin141:
But it's also a waste of time (and potentially wrong) just to make one up.  I'd rather let it auto-detect and and least have a chance doing something useful.  Performance is not really a priority in `singfun`, at least as far as I'm concerned.

We can still special-case the zero function if you really think that's worthwhile, but I can't see the justification for keeping this strange heuristic.  It makes the code more mysterious, and it's not clear what we do gain from it, if anything.

On an unrelated note, I've just noticed that none of the tests in `tests/chebfun/test_plus.m` and `tests/singfun/test_plus.m` touch this case in which the exponents differ by a non-integer at all.  This is a little disturbing but not totally, as we know that it's badly behaved.

nickhale:
OK. I'd be happy to special case the addition of zero functions and let the constructor deal with everything else.

aaustin141:
I've now made the change I wanted to make.  If you're happy with it, let me know, and I'll go ahead and merge this.

nickhale:
I'm happy.

aaustin141:
I've merged this, but I wound up reverting my commit (33aebc67), since it caused a test in tests/chebfun/test_cumsum.m to fail.  The heuristic choice of exponents happens to work in that case (since the exponents are negative), but the auto-detection fails for the usual reasons.

I don't have a good way to fix the auto-detection, and this isn't a big deal, so I've just undone my changes.  I still feel that my remarks above regarding the heuristic are relevant, but I also feel that it's not worth putting too much thought into this issue at this time.

PR #1686, opened 2015-12-03
Feature +cheb choreo

Hadrien-Montanelli:
`cheb.choreo` and `cheb.choreosphere` compute choreographies in the plane and on the sphere.


aaustin141:
Other than the fact that the help texts for these files are a little sparse, this looks pretty good to me.  The help text doesn't need to be super extensive, but there are a few things it should discuss.  If I weren't already familiar with the work behind this, I would probably be confused about what I'm being asked for when prompted for "angular velocity" or what's mean when it asks if I want to see the "planets dance".  The help text should also probably mention the need to CTRL-C to stop the program.

A few other minor things I noticed are:

 - There's a discrepancy between `choreo.m` and `choreosphere.m`:  when prompting the user about showing the bodies in motion, one gives options of `(Yes=1, No=0)`, while the other displays `(Yes=1, No=Enter)`.  We should pick one or the other and use it in both circumstances.

 - The "angular velocity" prompt asks for a "non-integer" value.  Is this really necessary?  I tried using a value of 1 and everything worked just fine.  Perhaps the prompt meant to indicate that arbitrary floating-point values are acceptable, not only integers?

 - When running the movie for a planar choreography with a nonzero angular velocity, I sometimes find that the "star field" doesn't cover the entire background as the plot rotates.  This is only visible when the solution is non-circular.

nickhale:
```
>> cheb.choreo
How many bodies? (e.g. 5) 5
Angular velocity? (non-integer, e.g. 1.2) 1.2
Undefined function or variable 'imfreehand'.
Error in cheb.choreo (line 31)
  h = imfreehand; 
```

I suspect that `imfreehand()` is part of the image processing toolbox.

aaustin141:
Ugh.  You're right.  It's in `toolbox/images/imuitools/imfreehand.m` under my MATLAB install path.

I'm not aware of any replacement for this that exists within core MATLAB.  Is anyone else?  If there isn't one, do we still want to include this anyway, even though it breaks our rule for the `+cheb` folder (and Chebfun in general) against dependencies on MATLAB toolboxes?

Hadrien-Montanelli:
Thanks for your comments @aaustin141 and @nickhale. 

More fundamentally, the code relies on `fminunc` which is part of the Optimization Toolbox. I think we should allow it in the `+cheb` folder, but we have to discuss it with the rest of the team.

trefethen:
Dependencies on toolboxes are always a shame, but I think we agreed that we won't insist on eliminating them for codes in +cheb.

aaustin141:
To the contrary:  the [README.txt](https://github.com/chebfun/chebfun/blob/e5c86034830e1cbc90bd839f139205afb63bc404/%2Bcheb/README.txt) file in the `+cheb` folder states that one of the requirements for a function to be in this folder is that it "depends only on built-in MATLAB and core Chebfun."  So it would be a change of policy.  (Not saying that's a bad thing, but we should be clear about it.)

trefethen:
Ah, well spotted!  I stand corrected.   @Hadrien-Montanelli, do you think you can find a way to make choreo work pretty well without relying on toolboxes?

Hadrien-Montanelli:
No, I'm afraid `imfreehand` and `fminunc` are really needed here.

trefethen:
I've given @Hadrien-Montanelli the green light to put `choreo` in `+cheb` an an exception to our usual policy of no toolbox dependencies.  Of course, if this troubles people we can always revisit the issue.  Also I have asked him to include a simple workaround to the `imfreehand` dependency for users (like myself, as it happens) who have the Optimization toolbox but not the Image Processing toolbox.  There is no such simple workaround it seems for the `fminunc` dependency.

Hadrien-Montanelli:
@aaustin141, I've done the modifications you and Nick asked!

trefethen:
Thanks @Hadrien-Montanelli and @aaustin141.  I've tried this on my laptop (which doesn't have `imfreehand`), and it worked nicely (well, I had some trouble with 7-point choregoraphies but mainly it worked).

A small suggestion: could the comment about angular velocity be changed from "e.g. 1.2" to, say, "e.g. 0 or 1.2"?  The fact is, zero angular velocity is the case of basic interest to the world, and nonzero is gravy.  Moreover, mentioning only 1.2 might mislead a user into thinking that this number needs to be nonzero.

A more substantive suggestion: I am asked *both* to declare how many bodies there are *and* to "press <enter> for final point".  This seems redundant.  

`cheb.choreo` is very impressive.

Hadrien-Montanelli:
Thanks for the feedback @trefethen. 

_A small suggestion: could the comment about angular velocity be changed from "e.g. 1.2" to, say, "e.g. 0 or 1.2"? The fact is, zero angular velocity is the case of basic interest to the world, and nonzero is gravy. Moreover, mentioning only 1.2 might mislead a user into thinking that this number needs to be nonzero._

I've issued a pull request with this change implemented.

_A more substantive suggestion: I am asked both to declare how many bodies there are and to "press for final point". This seems redundant._

The number of points is for interpolation and is not related to the number of bodies.

PR #1688, opened 2015-12-10
Respect x-axis limits with "hold on" in trigtech/plotcoeffs

gradywright:
Fixed bug where the x-axis properties were not being respected when the plot had been held.


trefethen:
@aaustin141 and @gradywright, it's perfect now.  Thanks for this fix at great speed!

PR #1691, opened 2015-12-11
Fix chebop multiple outputs simplify

### LINKED ISSUE #1689: Can we call simplify on components of [u,v] = N\f?
Chebop solves ODEs in quasimatrix mode, so that if you have several components of the solution of a problem, they all come out with the same length:
```
L = chebop(@(t,u,v) [diff(u,2)+u; diff(v,2)+100*v],[0,10]);
L.lbc = @(u,v) [u-1; diff(u); v-1; diff(v)];
[u,v] = L\0;
length(u)
ans =   86
length(v)
ans =   86
```
Often the various components may have very different natural lengths, however:
```
length(simplify(u))
ans =  26
length(simplify(v))
ans =  86
```
Could we change things so that when users specify outputs componentwise in this `[u,v] = ...` fashion, `simplify` is applied to each output before it is returned?  (In principle I guess I'd suggest the same for the outputs of `deal`, though `deal` is hardly needed by users any more now that we have `[u,v] = ...`.)  Perhaps @asgeirbirkis would be willing to implement this if @jaurentz or others have no objection.

asgeirbirkis:
Closes #1689.

### DIFF
--- a/@chebmatrix/deal.m
+++ b/@chebmatrix/deal.m
@@ -16,6 +16,10 @@ function varargout = deal(X, simplifyFlag)
 %       N.bc = @(x,u,v) [u(0) - 1; v(0)];
 %       [u,v] = deal(N \ 0);  % solution components of a system
 %                             % of differential equations
+%
+%   However, users do not normally need this syntax as it is
+%   simpler to write  [u,v] = N \ 0 ; .
+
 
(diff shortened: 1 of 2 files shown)


trefethen:
@nickhale, are you in the mood to merge this?  If not, @asgeirbirkis, why don't you come by my office and we can see about doing it here.  (People are beginning to disperse for Christmas.)

PR #1701, opened 2016-01-04
Catch unsupported lambda*u/x syntax on RHS.

### LINKED ISSUE #1697: chebgui generates incorrect chebops for eigenvalue problem?
According to http://dx.doi.org/10.1002/cpa.3160080307, equation (1), the eigenproblem

    u" + lambda u x^{-1} = 0, u(0) = 0, u'(1) = 0,

has eigenvalues lambda_n = (sigma_n/2)^2, where sigma_n is the nth zero of the zeroth Bessel function J_0. (The boundary conditions imposed in (1) are written differently, but the eigenfunctions do indeed satisfy that Dirichlet boundary condition on the left.) Thus, I think the eigenvalues should be

    [1.4457964907366962, 7.6178155859155217, 18.721751697673795, 34.760071106614959, 55.733075904408544, 81.640838233082121, 112.48338212950888, 148.2607174139888, 188.97284869598326, 234.61977836892356, ...]

However, when I implement this in chebgui, I get

    lambda =
    
    2.4674
    22.2066
    61.6850
    120.9027
    199.8595
    298.5555

Looking at the generated chebeig.m, one part stands out:

    % Assign the equation to two chebops N and B suchthat N(u) = lambda*B(u).
    N = chebop(@(x,u) diff(u,2)+0./x, dom);
    B = chebop(@(x,u) -u, dom);

which is a different eigenproblem. (The division by x is acting on zero, not on -u.)

Have I made a mistake, or is this a bug in chebgui?

Thanks,

Patrick

asgeirbirkis:
Moving the division by x automatically to the LHS would require fairly substantial work and rewrite of code. The approach I'm taking here instead is to look for such division (or an extra multiplication, e.g. `lambda*(u+u')*sin(x)`) and throw an error to the user, cf. attached screenshot.
screen shot 2016-01-04 at 21 33 13
 

Closes #1697.


nickhale:
I changed the example (and format) of the error message, since the example given would break in any case.

If @asgeirbirkis is OK with this, it can be merged.

asgeirbirkis:
Fine with me, please go ahead and merge!

aaustin141:
This looks OK to me.  Shall I go ahead and merge this?

nickhale:
Yes. Sorry, I forgot about it.

Issue #1707, opened 2016-01-11
Accuracy of `jacpts` for very large n


ajt60gaibb:
James Bremer, Associate Professor at UC Davis, just told me about the following inaccuracies in `jacpts`: 

```
[x, w] =jacpts(10^6, -1/2,0);
sum(w) - 2*sqrt(2)
ans =
  -5.2684e-09
```

I do not think this is rounding error due to the `sum` command. More likely, `jacpts` should have done one extra Newton iteration.

nickhale:
I think this is more complicated than simply requiring an additional Newton iteration. My guess is that the problem lies in the computing the final `cos(alpha)` term in the interior formula (which needs to be done accurately). It doesn't look like the faked (Hi-Lo) extended precision computation is working in some cases.

A more convenient example for verifying this is
```
>> [x, w] = jacpts(114, -1/2,0); sum(w) - 2*sqrt(2), k = k+1;
ans =
    -1.603162047558726e-13
```

nickhale:
I think this is more complicated than simply requiring an additional iteration. Rather, I suspect the problem is in computing the final `cos(alpha)` values in the interior formulae, which needs to be done very accurately. We do this by faking an extended precision computation with a high-low factorisation, but it seems maybe this is failing in some cases.

A more convenient example for experimenting with is
```
>> [x, w] = jacpts(114, -1/2,0); sum(w) - 2*sqrt(2), k = k+1;
ans =
    -1.603162047558726e-13
```

LourencoPeixoto:
I agree with Nick. The question is not to require an extra iteration. However I'm not sure that the problem is on computation of `cos(alpha)`. I'll investigate it.

nickhale:
Closed by 95d2739 in #1702.

```
>> [x, w] = jacpts(n, a, b); sum(w) - 2*sqrt(2)
ans =
     5.329070518200751e-15
```

... this took me all morning to find. =(

ajt60gaibb:
Sweet!

nickhale:
Closed by #1702.

LourencoPeixoto:
Great!

But the problem still remains in some cases. Check this example:
```
>> [x, w] = jacpts(1e5, 1,1);
>> sum(w)

ans =

   1.333333333031809
```
The exact sum is ```1.333333333333333```.

[...]

Issue #1645, opened 2015-11-12
Refactoring of ***PTS() methods


nickhale:
With the introduction of `ultrapts()` there's now considerable code overlap between the `legpts()`, `jacpts()`, and `ultrapts()` methods. I suggest refactoring these into a `quadpts()` method. 

To be clear, the `legpts()`, etc, methods would still exist, but simply be wrappers for a call to `quadpts()`.

Edit: Alternatively we bundle all the code in to `jacpts()`, since the others are a special case of this anyway. However, `quadpts()` might eventually allow for other non-Jacobi nodes and weights.

LourencoPeixoto:
@nickhale , I proposed some modifications in JACPT and ULTRAPTS to they call the other particular cases. See #1650, and #1651.

PR #1711, opened 2016-01-14
Fix jacpts

LourencoPeixoto:
Compute the constant out the front in a more stable way using the ratio of gamma functions. See #1707.

### DIFF
--- a/jacpts.m
+++ b/jacpts.m
@@ -161,7 +161,7 @@ end
 
 % Compute the constant for the weights:
 if ( ~strcmpi(method,'GW') )
-    if ( strcmpi(method,'ASY') )
+    if ( n >= 100 )
         cte1 = ratiogamma(n+a+1,b);
         cte2 = ratiogamma(n+1,b);
         C = 2^(a+b+1)*(cte2/cte1);
--- a/jacpts.m
+++ b/jacpts.m
@@ -69,17 +69,17 @@ end
 % Check inputs:
 if ( nargin > 3 )
     if ( nargin == 5 )
-        % Calling sequence = LEGPTS(N, INTERVAL, METHOD)
+        % Calling sequence = JACPTS(N, INTERVAL, METHOD)
         interval = int;
         method = meth;
         method_set = 1;
     elseif ( nargin == 4 )
         if ( ischar(int) )
-            % Calling sequence = LEGPTS(N, METHOD)
+            % Calling sequence = JACPTS(N, METHOD)
             method = int;
             method_set = true;
         else
-            % Calling sequence = LEGPTS(N, INTERVAL)
+            % Calling sequence = JACPTS(N, INTERVAL)
             interval = int;
         end
     end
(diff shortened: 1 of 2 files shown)


aaustin141:
This change is simple enough that I'm tempted to merge it, but after reading through the discussion on #1707, I'm not left with the sense that we've not quite reached consensus on what to do here yet.  This implements the "ratio of gamma functions" approach to computing this constant, but as you point out in [your comment](https://github.com/chebfun/chebfun/issues/1707#issuecomment-172024658) on that thread, this can be problematic for small n, and this pull request doesn't do anything about that.

Also, there should probably be a test or two added to `tests/misc/test_jacpts.m` to check for this issue since it's so subtle.  (#1702 should have done this as well but didn't.)

nickhale:
For small `n` the direct calculation of this constant (via `gamma()`) should cause no trouble.

LourencoPeixoto:
I've updated the file as we point out in #1707. Now is fine to me.

Also, I'm going add some tests to ```tests/misc/test_jacpts.m```.

nickhale:
I think you need the conditional to be on the size of `n`, not on the method used?

LourencoPeixoto:
Oh, indeed! I'll fix now!

LourencoPeixoto:
I think that there are some typing errors on comments. Also, I believe that 

```w(end) = .5*w(end);```

 is missing on the line 897. See 08f7873.

**Comment by aaustin141** [comment]
@LourencoPeixoto:  This is looking better; thanks for your work!  Can you add one more test that checks the validity of the change you proposed to the line you mentioned?

**Comment by LourencoPeixoto** [comment]
@aaustin141: Actually the code

`w(end) = .5*w(end);`

on the line 897 that I proposed is not affecting the `jacpts` commmand now because the function `asy2_higherterms` is not calling that. I proposed this change just in case. Therefore the test doesn't make sense unless we increase the number of output arguments in `asy2_higherterms`... Recall that only four output arguments are necessary and sufficient, namely `tB1`, `A2`, `tB2` and `A3`. I believe that @nickhale or @ajt60gaibb could say if this change is really fine.

On the other hand, we also could delete all the lines that not are being called:

```
% A3p:
A3p = D*A3;
A3p = A3p - A3p(1);
A3p_t = A3p./t;
% Extrapolate point at t = 0:
w = pi/2-t(2:end);
w(2:2:end) = -w(2:2:end);
w(end) = .5*w(end);
A3p_t(1) = sum(w.*A3p_t(2:end))/sum(w);

% B3:
tB3 = -.5*A3p - (.5+alph)*(C*A3p_t) + .5*C*(f.*A3);
B3 = tB3./t;
% Extrapolate point at t = 0
B3(1) = sum(w.*B3(2:end))/sum(w);

% A4:
K = C*(f.*tB3);
A4 = .5*(D*tB3) - (.5+alph)*B3 - .5*K;
A4 = A4 - A4(1);

% Make function for output:
tB1 = @(theta) bary(theta, tB1, t, v);
A2 = @(theta) bary(theta, A2, t, v);
tB2 = @(theta) bary(theta, tB2, t, v);
A3 = @(theta) bary(theta, A3, t, v);
tB3 = @(theta) bary(theta, tB3, t, v);
A4 = @(theta) bary(theta, A4, t, v);
```

 @nickhale, @ajt60gaibb: any suggestion?

nickhale:
As far as @LourencoPeixoto is correct, and it should be `w(end) = .5*w(end);`. (Of course, what we should have done was to use 1st-kind points, in which case this extrapolation would not have been necessary!)

However, I also agree that for clarity we may as well remove the code which is never called.

LourencoPeixoto:
Ok, @nickhale. I have removed the code which is never called. See 319dde9.

aaustin141:
All right---this is looking good to me now.  @nickhale:  do you have any further comments on this, or am I OK to merge it?

nickhale:
This looks OK.

There's still a *lot* of duplication between `legpts()`, jacpts()`, and `ultrapts()` (as pointed out in #1645), but that's for another day.

LourencoPeixoto:
Regarding duplication, are #1650 and #1651 not sufficient for now?

Of course, with the updated `ultrapts()` in #1725, we have new overlaps that could be arranged.

PR #1715, opened 2016-01-18
More graceful bailing out of Newton iteration when linear solves go awry

asgeirbirkis:
If when solving the linear system for the Newton update did not converge, we end up with a gibberish update, from which there is little chance to converge afterward. This will cause the solution process to halt (we're solving something that's way too noisy, with a large discretization size), so it's better to bail out of the Newton iteration. This PR allows the Newton algorithm to exit more gracefully out of such situations.

### DIFF
--- a/@chebop/dampingErrorBased.m
+++ b/@chebop/dampingErrorBased.m
@@ -126,7 +126,17 @@ while ( ~accept )
     
     % Compute a simplified Newton step using the current derivative of the
     % operator, but with a new right-hand side.
-    [deltaBar, disc] = linsolve(L, deResFunTrial, disc, vscale(u), pref);
+    [deltaBar, disc, converged] = linsolve(L, deResFunTrial, disc, ...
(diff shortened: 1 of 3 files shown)


aaustin141:
This looks OK to me.  Just a few questions:
 - Can you give an example that illustrates the behavior (or, better yet, add one to the tests)?
 - We return this `converged` output from `linsolve()`, but nothing's using it.  Is this necessary?

asgeirbirkis:
Thanks @aaustin141  for taking a look. To answer your questions:

1. Yes, I'll try to create a test out of this, so please don't merge this PR until that gets done.
2. The output `converged` from `linop/linsolve` is actually used in lines 95-98 in `chebop/solvebvpNonlinear` of this branch, and similar lines in `chebop/dampingErrorBased`.

aaustin141:
OK, great!  Sorry about the second question---I don't know how I missed that!

Issue #1509, opened 2015-05-27
ultraS representations from diffmat() and additional outputs.


nickhale:
As far as I can tell, it isn't possible to obtain ultraS differentiation matrices from `diffmat()`. Is there a reason for this?

Furthermore, it would be useful if the `diffmat()` function returned (optional) additional outputs depending on the type of discretisation requested.

For example:
 (a) The rectangular projection matrix if cheb1 or cheb2 collocation is used.
 (b) The conversion matrices for ultraS representations.

My motivation for the this is mostly the latter, since at the moment I usually have to resort to writing five or six lines of code to generate ultraS diffmats and their conversion matrices by hand. Since Chebfun is probably **the** place to play with ultraS representations^\* it would be nice if this was made more accessible.

^\* In addition to ApproxFun, of course.

Do @ajt60gaibb or @kuanxu have any thoughts?

kuanxu:
@nickhale I totally agree that we should have additional outputs, for both (a) and (b). Both of them are easy to implement, as we have the code in the bitbucket repo for our paper, right?

ajt60gaibb:
What is wrong with  `ultraS.diffmat` and `ultraS.convertmat`?

nickhale:
Because it's not _immediately_ obvious which conversion matrix corresponds to which differentiation matrix, particularly since the help text of `ultras.convertmat` is still wrong:
```
function S = convertmat(n, K1, K2)
%CONVERTMAT  Conversion matrix used in the ultraspherical spectral method.
%   S = CONVERTMAT(N, K1, K2 ) is a private method for constructing conversion
%   matrices. It returns the N-by-N matrix realization of the conversion
%   operator. The matrix S maps N coefficients in a C^{(K1)} basis to N
%   coefficients in a C^{K2} basis. Here, C^{(K)} is the ultraspherical
%   polynomial basis with parameter K.
```
```
>> full(ultraS.convertmat(5,0,0))
ans =
    1.0000         0   -0.5000         0         0
         0    0.5000         0   -0.5000         0
         0         0    0.5000         0   -0.5000
         0         0         0    0.5000         0
         0         0         0         0    0.5000
```

 * (a) The method doesn't appear to be private.
 * (b) If `K1` = `K2` I should expect to get an identity matrix, no?

Also `[D, S] = diffmat(5, 2, [0, 1], 'ultras');` seems nicer than
```
D = 4*ultraS.diffmat(5, 2);
[...]

PR #1729, opened 2016-01-27
Fix jac2heb

nickhale:
Deal with the special cases of `alpha=beta=-0.5` and `alpha=beta=0.5`, which are simply diagonal and banded scalings, respectively.

### DIFF
--- a/cheb2jac.m
+++ b/cheb2jac.m
@@ -25,7 +25,33 @@ if ( alph == 0 && bet == 0 )
     return
 end
 
-c_jac = cheb2jac_direct(c_cheb, alph, bet);
+if ( alph == -.5 && bet == -.5 )
+    % Special case: alpha = beta = -.5 (scaled chebpoly1)
+    [n, m] = size(c_cheb);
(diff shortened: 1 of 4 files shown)


nickhale:
*bump*

aaustin141:
This looks OK to me except in one, mostly unrelated aspect, and that's that the help text for `ultraS.convertmat()` is confusing.  It says that `ultraS.convertmat(n, k1, k2)` gives the matrix for converting `n` coefficients in a `C^{(k1)}` basis to `n` coefficients in a `C^{(k2)}` basis.  Reading this, it sounds to me like the `ultraS.convertmat(n, 0, 0)` lines should return the `n`-by-`n` identity matrix, but in fact, they return the matrix used to convert coefficients for first-kind Chebyshev polynomials to second-kind ones.  That's confusing.

I'll merge this anyway, but we really should fix the help text for that function.

nickhale:
Thanks @aaustin141 . 

I mentioned the problem with the `convertmat()` documentation previously: #1509 (see my second comment).

PR #1742, opened 2016-02-02
Reset color order to original (not to 1) in CHEBFUN/PLOT

### LINKED ISSUE #1741: Wrong line colors from plot() under "hold on".
Consider the following plot of `x`, `x^2`, and `x^3`, created using plain MATLAB:
```matlab
x = linspace(-1, 1, 32.');

figure(1);
clf();
hold on;
plot(x, x);
plot(x, x.^2);
plot(x, x.^3);
hold off;
```
builtin

I would expect the Chebfun plot generated by the following code to look identical, but it's not.  The color of the line for `x^3` is wrong:
```matlab
x = chebfun(@(x) x);

figure(2);
clf();
hold on;
plot(x)
plot(x.^2)
plot(x.^3)
hold off;
```
chebfun

I'm running this on MATLAB R2015b.  Note that if I create the Chebfun plot using a quasimatrix via `plot([x x.^2 x.^3])`, the colors come out fine.

nickhale:
Closes #1741.

(Whilst looking at this, it's not clear to me what the colour rules are when plotting delta functions.)

### DIFF
--- a/@chebfun/plot.m
+++ b/@chebfun/plot.m
@@ -101,6 +101,10 @@ end
 
 % Store the hold state of the current axis:
 holdState = ishold;
+% Acquire initial color cycle if running R2014b+.
+if ( ~verLessThan('matlab', '8.4') )
+    originalColorOrder = get(gca, 'ColorOrderIndex');
+end
 
 % Store the current X and Y-limits, and whether ylim is in manual or auto mode.
 if ( holdState )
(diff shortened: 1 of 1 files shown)


aaustin141:
This fixes the problem from #1741 nicely, but now the colors that appear on the legend for a quasimatrix plot may be incorrect.  Try sticking the following lines in a function or script:
```matlab
x = chebfun(@(x) x);
plot([x x.^2 x.^3]);
legend('x', 'x^2', 'x^3');
```
Now, switch to this branch, clear classes, etc., and run this file twice.  On the first run, I get
run1
which is OK, but on the second, it becomes
run2
which is wrong.

aaustin141:
That seems to have done the trick.

PR #1746, opened 2016-02-03
feature-trig-lebesgue

### LINKED ISSUE #1717: A Lebesgue command for trigonometric interpolation
A `lebesgue` command for trigonometric interpolation would be very useful.  The syntax could even be 
```
[L, c] = lebesgue(linspace(-pi,pi,n), 'trig')
```
as defined in https://en.wikipedia.org/wiki/Trigonometric_interpolation.

aaustin141:
Closes #1717.

### DIFF
--- a/lebesgue.m
+++ b/lebesgue.m
@@ -84,17 +84,18 @@ elseif ( nargin == 4 ) % LEBESGUE(X, A, B, 'trig')
         error('CHEBFUN:lebesgue:parseInputs:badArg3', 'Invalid argument.');
     end
 else
-    error('CHEBFUN:lebesgue:tooManyArgs', 'Too many input arguments.');
+    error('CHEBFUN:lebesgue:parseInputs:tooManyArgs', ...
+          'Too many input arguments.');
 end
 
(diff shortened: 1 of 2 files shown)


ajt60gaibb:
Thanks @aaustin141, its pretty code and convenient to use.  

Do you think line 177 should also print a warning?  We currently have
```
oddNumber = 11; 
[leb,const] = lebesgue(linspace(-1,1,oddNumber),'trig')
Error using lebesgue>trigLebesgue (line 182)
LEBESGUE for trigonometric interpolation requires an odd-length grid.

Error in lebesgue (line 38)
    L = trigLebesgue(x, d);
```
I do not feel strongly about it.

aaustin141:
Good point.  I think a warning on [line 177](https://github.com/chebfun/chebfun/blob/9dbdd3080aa259d31cc974b843801b53621bf656/lebesgue.m#L177) is probably overkill here, since it would then appear whenever you did, e.g., `lebesgue(linspace(-pi, pi, 16).', [-pi pi], 'trig')`, and that's kind of annoying.  Instead, I've modified the error message to hint that periodic endpoints may be a problem.

PR #1749, opened 2016-02-05
Feature spherefun 

gradywright:
Spherefun is ready for code review.


bhashemi:
@gradywright, @ajt60gaibb, @heatherw3521 -- @nickhale has recently modified the way we parse the inputs in the `chebfun2` constructor (#1765). Should we do the same for the `spherefun` constructor?

gradywright:
Yes, I think this is a good idea and something I've been meaning to look at.

ajt60gaibb:
@bhashemi, can we merge the pull request and then sort out the input parsing after that?   Is that reasonable?  I added an issue #1783.

bhashemi:
@ajt60gaibb , yes; that's fine. I restart reviewing the code today and will apply some trivial modifications according to coding style.

ajt60gaibb:
Great. Thank you!

bhashemi:
@ajt60gaibb , @gradywright , @heatherw3521 , I have two comments at the moment about `spherefun/constructor`:
1) I think the variable names 'evp' and 'evm' in the constructor are not informative. Is it possible for you to change these names?
2) In Chebfun2 constructor, a tolerance is given to `simplify` for columns and rows. See e.g., #1777 .  I suspect a similar thing is needed in `spherefun/constructor`. I am not suggesting this to be done right now, but it might be good to explore this at some point. This can make the code look more unified that what it is now.

gradywright:
In response to (1), yes I agree that 'evm' and 'evp' could be more descriptive.  If we follow the notation in the spherefun paper, these correspond to $m^{-}$ and $m^{+}$, so we could call them 'mminus' and 'mplus'.  Or more descriptive names would be 'minusEigenvalueM' and 'plusEigenvalueM'.  I'm happy to make these changes.

In regards to (2), I believe @ajt60gaibb added an issue in tracker (#1785).

Thanks @bhashemi!

bhashemi:
This is (probably) a final set of comments/questions I have about the `spherefun` branch. I divide them into parts A and B. My feeling is that Part A needs to be done before merging, but you may want to delay Part B. The code generally looks very good and as you'll see most of my comments are related to the help texts. 

Part A:
1) In `spherefun/norm`, the "otherwise" case of the "swich" command contains an incomplete line for an error message.

2) In the subroutine `sphereQR` of `svd`, there is a call to `legpts`, but the comment says "Trigpts".

3) The help text of `spherefun/quiver` is a bit confusing. It says that it creates a quiver plot of _a_ SPHEREFUN, but there are three inputs which I think should all be spherefuns.

4) The help text of `spherefunv/smooth` is not clear at least to me, as e.g. it mentions both the heat and Helmholtz equations. Hadrien is reviewing parts of spherefun related to differentiation and the whole `spherefunv`. So, I was not supposed to comment about this code... oops!

5) The help text of the subroutine `mySphHarm` in `sphharm`, seems to be duplicated from the help text of the main code `sphharm`. It also contains communications between developers which we might want to refine. 

6) Line 16 of `spherefun/vertcat`, allows only for 3 `spherefun` objects to be concatenated and causes an error message if two inputs are given. The help-text on the other hand, says it's OK to have two inputs. I see that `spherefunv` actually allows only for 3 `spherefun` objects at the moment which seems not to be consistent with `chefbun2v`.

7) Is it possible to double-check whether the help text of `sphf2cartf` says what it is really doing?

8) Do we use `sph2torus.m` anywhere?

9) Some of the files do not have any copyright statement, some of them say 2015, some say 2016.

Part B:
1) Currently, the plus/minus structure of a `spherefun` is used in `sum2`, but not in `sum`. Wouldn't it be neat to take advantage of this structure also in `sum(f,2)`?

2) If I type e.g., 
`>> help spherefun/max`

then I see `SEPARABLEAPPROX`. I am not sure whether the appearance of the word `SEPARABLEAPPROX` in a help text is consistent with what is decided earlier.

gradywright:
Thanks @bhashemi!  

I have addressed 1-3 and 5 of Part A.  

I am a bit confused about 4 from Part A.  I could not find a `spherefunv/smooth` method and the `spherefun/smooth` method does not mention the Poisson equation.

We will get to the rest of Part A very soon.

bhashemi:
Sorry @gradywright . For Q.4 in Part A, I meant `spherefun/smooth`, not 'spherefunv`, where at Line 7, the _heat_ equation is mentioned.

gradywright:
Thanks for the clarification @bhashemi.  I updated the comment in `spherefun/smooth` to better explain why we call the Helmholtz solver.  

So, 1-5 of Part A of your comments have now been addressed.

ajt60gaibb:
@bhashemi, thank you for doing such a good job.  

Part A:
6)  I have changed the help text for `spherefun/vertcat`.  Is there a use for a `spherefunv` object with two components? 
7) I have updated the help text for `spherefun/sphf2cartf` so it is clearer.  
8) I have remove the command `sph2torus`.  It was used in the constructor at one point, but that was removed. Thanks. 
9) I assume this will get fixed by a universal sweep of all the Chebfun copyright statements at some point. 

Part B: 
1) I agree. 
2) There are probably many commands that fall into this category. I have added an issue: #1786

bhashemi:
Considering Q.6 in Part A above, I won't insist allowing `spherefunv` objects consisting of two `spherefun` objects, as it seems not to be useful. Thanks @ajt60gaibb and @gradywright for addressing all my comments/questions.

Hadrien-Montanelli:
Just a quick comment: most files don't respect the Chefbun coding style. (See chebfun_files/Chebfun5/codingstyle/codingStyle.pdf.) I'm happy to modify the spherefunv files and the files related to differentiation (Poission, Helmotz, diff, grad, gradient, ...) but I'm not going to do it for all the files. @bhashemi -- what did you do about this?

ajt60gaibb:
What coding styles did we miss exactly?

Hadrien-Montanelli:
(Could you also please add the Copyright in each file?)

Hadrien-Montanelli:
Example in Poisson.m: 
1. Line 2: No space between 'Poisson' and '%'
2. Line 4: We (and MATLAB) usually write FUNC(A, B, C) instead of FUNC( A, B, C )
3. Copyright missing

Example in Helmotz.m:
Line 65: 'length(C) + mod(length(C),2)' instead of 'length(C)+mod(length(C),2)'
Line 81: '.5i*spdiags(ones(n,1)*[-1,1], [-1 1], n, n) instead of '.5i*spdiags(ones(n,1)*[-1,1],[-1 1],n,n)'

Sorry if I'm being annoying and picky about it!

ajt60gaibb:
I have fixed all the copyright statements in `@spherefun` and `@spherefunv` commands. 

I am not sure what to say about (1) and (2)...

ajt60gaibb:
@Hadrien-Montanelli, instead of finding this picky, I take it as an amazing fact!!     We have `108` files and `213` commits on this pull request.   We are down to whitespace and indents.    Your comment is more of a compliment. We are basically ready. :+1:

Hadrien-Montanelli:
Unfortunately, I have nontrivial comments/questions too :). 
In Possion.m, to construct the Mccossin/Msin2 matrices, could you use trigpsec/diffmat? 
Also, in the paper page 21, is the matrix Dm correct? In the code, you're using trigspec/diffmat which generates a diagonal matrix with entries ```(1i)*[0, -m/2+1:1:m/2-1]```. (Note the zero in the first entry.)

Hadrien-Montanelli:
(Following my last comment, because of the zero in the first entry, the second-order diff. matrix ```D^{(2)}``` is not the square of the first-order diff. matrix, i.e., ```D^{(2)} ~= D^2```.)

ajt60gaibb:
@Hadrien-Montanelli, yes, we had a discussion by email on the 18th of July 2015 about this.   I was asking why there was a zero in the `(1,1)` entry of `trigspec.diffmat(4,1)`.  

Afterwards, I concluded that there is a zero `(1,1)` entry because the highest mode, i.e., `cos(m/2*x)`, has a derivative that is zero on the grid.   Since Grady and I were thinking totally in coefficient space (as opposed to value space) in the spherefun paper and `d/dx e^(1i*k*x) = 1i*k e^(1i*k*x)`, we decided that the zero `(1,1)` entry is unnecessary.  It also improved the clarify of the paper.

The `Dm` in the paper is not a mistake. It was intentional. 

As for the code, we originally made the `diffmat` ourselves.  We will go back to that.  Of course, this change barely makes any difference (only up to aliasing errors). Thank you for pointing this out.  Since we were not sure how `trigspec` was treating even and odd, we originally just made all the operators ourselves.  This has the add benefit of making the code faster for Navier-Stokes simulations, where the `Poisson` code is called thousands of times.

Hadrien-Montanelli:
Ok, thanks for the clarification!

Hadrien-Montanelli:
I think that your code Poisson should use trigpsec/diffmat and trigpsec/mulmat. Do you really think that calling trigspec/diffmat makes the code slower? If not, I'm happy to add a flag in trigspec/diffmat to zero (or not) the first coefficient in the even case.

Hadrien-Montanelli:
(Trivial comment reading the paper: after equation 5.6, I guess you mean F is the matrix of Fourier coeffs multiplied by M_sin^2. )

Hadrien-Montanelli:
(Isn't F mxn, as opposed to nxn?)

ajt60gaibb:
Yes, thank you for finding that typo.  Fixed in the revision (submitting in the next few days). 

OK. The `spherefun/Poisson` and  `spherefun/Helmholtz` code can call `trigspec.diffmat` code if there is a flag in `trigspec/diffmat`.  Great.

ajt60gaibb:
At this point, for non-critical issues, I am adding issues to be dealt with after this large pull request.   I 
just add #1798 regarding the comment that `spherefun/Poisson` and `spherefun/Helmholtz` should use `trigspec` instead of making their own Fourier operators. 

Is there anything critical left?   Of course, there is still @Hadrien-Montanelli's coding style issues (I have fixed some).  I will continue to fix them as I find them.  Does that kind of thing stop a pull request being issued these days?

Hadrien-Montanelli:
I'm still reviewing my part, @bhashemi and I haven't looked at the tests yet, give us a couple of more days :).

Hadrien-Montanelli:
More comments: (Please pull my changes first.)

1. In spherefun/diff: line 149, what exactly do you mean?

2. In spherefun/gradient: can we delete the commented lines? 

3. In spherefun/laplacian: can you please add an help text? Also, it seems that you've tested several methods and chose to use laplacianE. Can we get rid of the other functions? Can you also please explain what laplacianE does? (Just a few words.) 

4. In spherefun/quiver: there are several warnings about the use of TriRep. Also, is it possible to get rid of the commented lines at the end of BISECTTRI? Can you also please follow Chebfun coding style in the help texts of TRISECTTRI, BISECTTRI and TRISECTEDGE (and CIRCUMENTERSPHTRI)?

gradywright:
Okay @Hadrien-Montanelli we completed the requested changes 1-4 from your previous post.

gradywright:
It looks like all the tests in `spherefun` crash because we determine tolerances for the tests with a line of code that includes
```
chebfunpref().techPrefs.eps
```
Should this be changed to?
```
chebfunpref().techPrefs.chebfuneps
```
If so, do you want us to make the change?

bhashemi:
@Hadrien-Montanelli kindly reminded me that we need to review `tests` also. Here is a question in that regard. In
A) `test_constructor( )`, `test_diff( )`, `test_laplacian( )` codes of `tests/spherefun`
and in
B) `test_constructor( )` code of `tests/spherefunv`

do we need to keep comments in the two subroutines `SampleError(h, g)` and `getPoints(m, n)`?

bhashemi:
@gradywright , I replaced 'eps' with `chebfun2eps` (as opposed to `chebfuneps`) which I assume to be closer to the default tolerance we want to use in `spherefun` and `spherefunv`. Is that OK? 
I think we need to do the same in the beginning of `spherefun/constructor` when several variables including the `tol` are initialized. I guess this would be done later as part of addressing #1783

gradywright:
Thanks @bhashemi.  I removed the unnecessary comments from the tests you mentioned.  

Also, replacing `eps` with `chebfun2eps` works good.

Hadrien-Montanelli:
I've reviewed the SPHEREFUNV files and pushed some coding style/documentation improvements. A few comments:

1. In spherefunv/cross: do you want to implement the TODO?

2. In spherefunv/divergence: do you want to implement the TODO?

3. Can you also double check the Copyright in the SPHEREFUN and SPHEREFUNV files? It indicates 2015 in (at least) SPHEREFUN/PLOTCOEFFS2, SPHEREFUN/RESTRICT, SPHEREFUN/ROOTS, SPHEREFUN/SPHEREFUN and SPHEREFUN/SVD. It's missing in (at least) SPHEREFUNV/NORMAL.

4. Can you have a look at SPHEREFUNV/MLDIVIDE and MRDIVIDE, and check that you agree with the new help text? 

5. I've just tried the example you give in the paper:

```f = spherefun(@(x,y,z) cos(1 + 2*pi*(x+y) + 5*sin(pi*z)))```

works fine but 

```f = spherefun(@(la,th) cos(la + 2*pi*(cos(la).*sin(th) + sin(la).*sin(th)) + 5*sin(pi*cos(th))))```

gives a warning. 

6. How can I extract the first entry of a SPHEREFUNV object? 

```f = spherefun(@(x,y,z) cos(1 + 2*pi*(x+y) + 5*sin(pi*z)));
F = [f; f; f];
F(1)```

doesn't work. (That's the syntax we use for CHEBFUN2V.)

7.  Can you please follow Chebfun coding style in the help texts of the functions in SPHEREFUNV/QUIVER?

**Comment by ajt60gaibb** [comment]
`1 and 2:`  There are lots of todos in `spherefun`. Let's leave them for now.  These two todos are minor things.   

`3:`  Fixed `spherefunv/normal`, thanks.  I assumed that `2015` will be changed to `2016` by #1737. 

`4:` I modified the help text of `mldivide` so that I am happy. 

`5:` Yes, `f = spherefun(@(la,th) cos(la + 2*pi*(cos(la).*sin(th) + sin(la).*sin(th)) + 5*sin(pi*cos(th))))` was a typo in revision `1` that has already been fixed to `f = spherefun(@(la,th) cos(1 + 2*pi*(cos(la).*sin(th) + sin(la).*sin(th)) + 5*sin(pi*cos(th))))`. (Note that `la` has become `1`.)

This works for me: 
```
f = spherefun(@(x,y,z) cos(1 + 2*pi*(x+y) + 5*sin(pi*z))); 
F = [f; f; f]; 
F(1)
ans =
   spherefun object
       domain        rank    vertical scale
     unit sphere      23            1
```
Just about to update the help text for `spherefunv/quiver` too.   

Thanks, @Hadrien-Montanelli.

Hadrien-Montanelli:
Thanks, looks good. I'm going to have a meeting with @bhashemi to talk about tests and a few files we haven't looked at. Almost there!

ajt60gaibb:
Thank you.  An acknowledgment in the spherefun paper to @Hadrien-Montanelli and @bhashemi has been added.

gradywright:
The help text for `spherefunv/quiver` has been updated.

Hadrien-Montanelli:
A few comments about the tests:

1. It seems that most `separableApprox` methods are tested at the `chebfun2` level. We believe that (some of) these methods should be tested at the `spherefun` level too.

2. Several `spherefun` methods don't have tests. Same holds for `spherefunv`. 

We understand that having tests for every single file requires a lot of work, but there are for the moment only 4 `spherefunv` tests (for 31 methods) and 14 `spherefun` tests (for >120 methods).

Hadrien-Montanelli:
I've got a few questions about the plotting.

``` f = spherefun(@(x,y,z) cos(1 + 2*pi*(x+y) + 5*sin(pi*z)));
plot(f)```

1. Would it look better if the x-axis and the y-axis contained the -1 and +1 ticks?

2. Should we use `box on`? 

I've spent some time on thinking how to plot stuff on the sphere -- please have a look at http://epubs.siam.org/doi/pdf/10.1137/15M1024652 (page 246, figure 6).

Hadrien-Montanelli:
A few more comments:

1. Could you please add an example in the help text of `spherefun/Helmholtz`?

2. In their help text, many methods (curl, div, grad, ...) don't point to the correct methods, i.e., they point to the MATLAB methods, as opposed to the `spherefun` or `spherefunv` methods. You can type, e.g., "See also SPHEREFUNV/DIV" as opposed to "See also DIV".

ajt60gaibb:
Maybe you want to add an issue for `spherefun/plot` then?  The missing ticks are a (semi-)bug in MATLAB, though.

Hadrien-Montanelli:
I think the plotting has to be fixed before we merge this branch. I'm sure it's an easy fix. The missing tick is not a bug -- look at the x-limit and y-limit in `gca`. Since it's always the unit sphere, you can overwrite them to include the end points.

gradywright:
If you think it's an easy fix @Hadrien-Montanelli, then by all means make it.  Thanks.

Hadrien-Montanelli:
Done -- what do you think about `box on`?

Hadrien-Montanelli:
(We don't use it in `chebfun2`.)

gradywright:
I prefer to leave that as a user option, but I don't feel strongly about it.  Note that if you type `sphere` in `matlab` it brings up a figure of a sphere without `box on`.  Thanks for updating the code @Hadrien-Montanelli

Hadrien-Montanelli:
OK @gradywright!

gradywright:
@Hadrien-Montanelli and @bhashemi:

We have included an example in the help text of `spherefun/helmholtz`.

We have updated the SEE ALSO help text in the various `spherefunv` methods to point to the correct `spherefun` or `spherefunv` methods.

Finally, we have added 14 new tests for `spherefun` (doubling the previous number) and 11 new tests for `spherefunv`.  We added an issue (#1806) about strengthening these test suites further in the future.

Hadrien-Montanelli:
@gradywright, @ajt60gaibb, @heatherw3521 -- congratulations! We've finished the review process and ran the tests successfully. You did a great job, the spherefun code is impressive.

gradywright:
Thank you @bhashemi and @Hadrien-Montanelli!

ajt60gaibb:
Thank you!!

heatherw3521:
This is great news! Thanks so much @bhashemi  and @Hadrien-Montanelli !

Issue #1606, opened 2015-09-13
shape of output when evaluating transposed quasimatrix / array-valued chebfun at an array of points


jrus:
I’m pretty confused by the shape of the output when I evaluate an array-valued chebfun at an array of points.

For example:
```matlab
>> f = chebfun('x');
>> f_array = [f f];
>> ft_array = f_array';
>> f_array([0 1])
ans =
     0     1     0     1

>> ft_array([0 1])
ans =
     0
     1
     0
     1

>> f_array([0; 1])
ans =
     0     0
     1     1

>> ft_array([0; 1])
ans =
     0     1
     0     1
```

Shouldn’t `f_array(x) == ft_array(x')'`?

Is there documentation somewhere explaining the reason it works like this?

jrus:
I’ve been reading the code more carefully, tracing the data flow of feval calls from chebfun -> bndfun -> onefun, and I think quite a bit of overhead can be cut by reducing the number of places where arrays get reshaped.

Currently, chebtech.clenshaw/trigtech.horner are at a low enough level to not care about evaluation-point vector shape, but just assume they get a column vector in (and reshape other arrays to be column vectors), and output an array with a column for each column of the coefficient array passed to them.

I think it would be beneficial to also keep {chebtech.feval/trigtech.feval, bndfun.feval/unbndfun.feval, singfun.feval, deltafun.feval} out of the business of worrying about the shape of `x`, and just take in column vectors and spit out one column per column of an array-valued fun. This all happens at the level below which there is any distinction between column and row chebfuns.

Then chebfun.feval – the high-level end-user entry point for evaluating a chebfun – which knows whether its own shape is k-by-∞ or ∞-by-k, and is designed to handle an arbitrary shape matrix of evaluation points, can be the sole place where input data is reshaped into one big column, passed down into the lower levels of the system for evaluation, and then reshaped back into whatever shape is appropriate at the end.

The code for actually doing the reshaping can also be simplified somewhat to be both easier to follow and faster to execute.

Any thoughts?

jrus:
Here’s the simplest way I could come up with to reshape input/output in chebfun.feval. This should work with quasimatrices, array-valued chebfuns, either column chebfuns and row chebfuns, and evaluation points `x` of arbitrary dimension:

```matlab
% Let’s assume the input evaluation points `x` is of shape `[m n p1 p2 ...]`,
% where the product `p1*p2*p3*... = p`. First we save these values:

[...]

PR #1751, opened 2016-02-05
Feature improved qr

nickhale:
There are two components to this PR:
 1) Computing in coefficient space at the `chebtech` level (speak to @jaurentz about this and see #1542 )
 2) For `chebfun` objects of multiple pieces, computing the QR factorization of each piece independently and then recombining the result.

These two can be separated if necessary.

Edit: Thanks to @jaurentz and @aaustin141 for useful discussions on this.

### DIFF
--- a/@chebfun/qr.m
+++ b/@chebfun/qr.m
@@ -61,22 +61,24 @@ elseif ( isArrayValued )
     for k = 1:numFuns
         [Q{k}, Rhat{k}] = qr(A.funs{k});
     end
-
-    % Step 2:
-    [Qhat, R] = qr(cell2mat(Rhat));     % Compute [Qhat, R] = qr(Rhat). 
-    R = R(1:numCols,:);                 % Extract require entries (top square).
-    m = cellfun(@(v) size(v, 1), Rhat); % m(k) = length of A.FUN{k}.
-    Qhat = mat2cell(Qhat(:,1:numCols), m, numCols); % Only need first column.
-
-    % Step 2b: Ensure diagonal has positive sign. (A = QR = (Q*S)*(S*R))
-    s = sign(diag(R)); s(~s) = 1; 
-    S = spdiags(s, 0, numCols, numCols); 
-    R = S*R; % (Qhat --> Qhat*S is performed below)
     
-    % Step 3: Fold Qhat (more precisely, Qhat*S) back in to Q.
-    for k = 1:numFuns
-        Q{k} = Q{k}*(Qhat{k,1}*S);
-    end
+    % Step 2: Compute [Qhat, R] = qr(Rhat),
+    [Qhat, R] = qr(cell2mat(Rhat));
+    R = R(1:numCols,:);       % Extract first block row.
+    Qhat = Qhat(:,1:numCols); % Extract first block column.
+
+    % Step 2b: Ensure the diagonal is non-negative. (A = QR = (Q*S)*(S*R))
+    s = sign(diag(R)); s(~s) = 1;
+    S = spdiags(s, 0, numCols, numCols);
+    R = S*R;
+    Qhat = Qhat*S;
+
+    % Step 2c: Separate the segments of Qhat back into a cell.
+    m = cellfun(@(v) size(v, 1), Rhat); % m(k) = length of A.FUN{k}.
+    Qhat = mat2cell(Qhat, m, numCols);
+    
+    % Step 3: Fold Qhat back in to Q.
+    Q = cellfun(@mtimes, Q, Qhat, 'UniformOutput', false);
     
     % Construct a new CHEBFUN from the computed FUNS:
     Q = chebfun(Q);
--- a/@chebfun/qr.m
+++ b/@chebfun/qr.m
@@ -70,8 +70,8 @@ elseif ( isArrayValued )
     % Step 2b: Ensure the diagonal is non-negative. (A = QR = (Q*S)*(S*R))
     s = sign(diag(R)); s(~s) = 1;
     S = spdiags(s, 0, numCols, numCols);
-    R = S*R;
     Qhat = Qhat*S;
+    R = S*R;
 
     % Step 2c: Separate the segments of Qhat back into a cell.
     m = cellfun(@(v) size(v, 1), Rhat); % m(k) = length of A.FUN{k}.
--- a/@chebfun/qr.m
+++ b/@chebfun/qr.m
@@ -9,40 +9,80 @@ function [Q, R] = qr(A, econ)
 % Copyright 2015 by The University of Oxford and The Chebfun Developers.
 % See http://www.chebfun.org/ for Chebfun information.
 
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-% Developer note:
-%  If A contains only a single FUN, then FUN/QR is used directly. If A has
-%  multiple pieces but each of these are simple CHEBTECH objects, then
-%  QRSIMPLE() is called. This violates OOP principles, but is _much_ more
-%  efficient.
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
-
-% Check inputs
+% Check inputs:
 if ( (nargin == 2) && (econ ~= 0) )
     error('CHEBFUN:CHEBFUN:qr:twoargs',...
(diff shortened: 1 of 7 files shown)


jrus:
> ```matlab
% TODO: This case is not tested!
%  DEVELOPER NOTE:
%   Currently (5th Feb 2016) the only way this is reachable is in the
%   case of a quasimatrix consisting of SINGFUN or DELTAFUN objects,
%   neither of which return anything sensible when we attempt to compute
%   a QR factorization.
% 
% Legendre-Vandermonde matrix:
L = legpoly(0:numCols-1, domain(A), 'norm', 1);
% Convert so that L is also quasimatrix:
L = cheb2quasi(L);
% Call abstract QR:
[Q, R] = abstractQR(A, L, @innerProduct, @normest);
```

Does this comment mean that this section of code only ever returns non-sensible output?

* * *

> ```matlab
function [Q, R] = qr(A, econ)
```

What’s the point of the second “econ” parameter, which is required to be zero or missing and doesn’t do anything?

* * *

> ```matlab
% Compute [Qhat, Rhat] = qr(Q):
m = cellfun(@(v) size(v, 1), R);
[Qhat, Rhat] = qr(cell2mat(R));
Rhat = Rhat(1:numCols,:); % Extract require entries.
Qhat = mat2cell(Qhat(:,1:numCols), m, numCols);
```

Can you folks add some comments in the code explaining what Qhat and Rhat are?

**Comment by jrus** [comment]
Actually, now that I’m looking at this closer, I think in general the comments and control flow can be tightened up. There are a few important typos in the comments, for instance, the comment `% Compute [Qhat, Rhat] = qr(Q):` should be `%... qr(R):` and `% If f has only one column we simply scale it.` should be `% If A has...`. I also think several comments would benefit from being written as full sentences.

I think the overall flow gets a bit nicer after adding some `return` statements to reduce the level of nesting.

**Comment by jrus** [comment]
Okay, that gets most of it. While the comment helps I still think `s = sign(diag(R)); s(~s) = 1; S = spdiags(s, 0, numCols, numCols);` is harder to follow than the alternative which multiplies appropriate rows/columns by -1, but anyway...

Even if you want the multiplication by `S`, you can do it all at once like `Qhat = Qhat*S;` before turning `Qhat` back into a cell, instead of doing the multiplication separately on each piece. I think it also keeps things clearer to put both multiplications by `S` in the same place right after `S` is defined.

The comment `% Only need first column.` doesn’t make sense to me in context.

I’m not sure if the change of names of `Rhat` ↔ `R` is really an improvement over the previous version (all these names are much too terse for me as a programmer), but it’s okay.

* * *

I think the logic is easier to follow and reason about organized like:

```matlab
    numFuns = numel(A.funs);

    % Step 1: Perform QR on each piece.
    Q = cell(numFuns, 1); Rhat = Q;
    for k = 1:numFuns
        [Q{k}, Rhat{k}] = qr(A.funs{k});
    end

    % Step 2: Compute [Qhat, R] = qr(Rhat),
    % then chop off their irrelevant right columns/bottom rows, respectively.
    [Qhat, R] = qr(cell2mat(Rhat));
    R = R(1:numCols,:);
    Qhat = Qhat(:,1:numCols);

    % Step 2b: Ensure the diagonal is non-negative. (A = QR = (Q*S)*(S*R))
    s = sign(diag(R));
    s(~s) = 1;
    S = spdiags(s, 0, numCols, numCols);
    R = S*R;
    Qhat = Qhat*S;
    
    % Step 2c: Separate the segments of Qhat back into a cell.
    m = cellfun(@(v) size(v, 1), Rhat); % m(k) = length of A.FUN{k}.
    Qhat = mat2cell(Qhat, m, numCols);
    
    % Step 3: Fold Qhat back into Q for each segment.
    for k = 1:numFuns
        Q{k} = Q{k}*(Qhat{k,1}*S);
    end
    
    % Construct a new CHEBFUN from the computed FUNS:
    Q = chebfun(Q);
```

jrus:
:+1:

I didn’t realize `cellfun` could return a cell like that instead of an array.

Would the same work earlier, like:
```matlab
[Q, Rhat] = cellfun(@(fun) qr(fun), A.funs, 'UniformOutput', false);
```

nickhale:
Code layout and documentation aside, there seems to be a fairly significant performance hit from the move to solving via the coefficients in `chebtech/qr()`.

Performing the following test 
```
A = chebfun(rand(1000, 100), [-1 1]); 
tic, [Q, R] = qr(A); toc
A(0,:) = A(0,:);
tic, [Q, R] = qr(A); toc
```
(which may or may not be representative of an operation we are interested in performing) I find:
```
>> git checkout development
Switched to branch 'development'
Elapsed time is 0.424995 seconds.
Elapsed time is 1.406427 seconds.
Elapsed time is 0.093466 seconds.
Elapsed time is 1.086327 seconds.
>> git checkout feature-improved-qr
Elapsed time is 1.512205 seconds.
Elapsed time is 1.948390 seconds.
Elapsed time is 1.020234 seconds.
Elapsed time is 2.001105 seconds.
>> git checkout feature-improved-qr2
Elapsed time is 0.232609 seconds.
Elapsed time is 1.152606 seconds.
Elapsed time is 0.106940 seconds.
Elapsed time is 1.168003 seconds.
```
where (feature-improved-qr2)[https://github.com/chebfun/chebfun/tree/feature-improved-qr2] reverts to the values-based computation in `chebtech/qr()`. (Recall this takes advantage of some caching, hence the speed up on consecutive runs.)

Summary: The changes in this PR to `chebfun/qr()` give a significant improvement in code (in particular encapsulation) with a negligible performance hit. The changes to `chebtech/qr()` also simplify the code, but with a significant performance hit.

Unless there are strong arguments against it, I suggest we revert to the values-based computation in `chebtech/qr()`. (Recall that `chebfun2/plus()` etc relies heavily on `chebfun/qr()`.)

nickhale:
In theory we could do this:
```
[Q, Rhat] = cellfun(@(fun) qr(fun), A.funs, 'UniformOutput', false);
```
(I didn't realise until now it was possible to get two outputs from `cellfun()`.)

In practice it seems that the orientation of `A.funs` is unreliable (sometimes it is a row vector, other times a column vector). I'm looking to see how easy this will be to fix up.

nickhale:
It seems that enforcing that the `.funs` field of a `chebfun` is stored vertically is nontrivial -- in fact, a number of existing tests check for the fact that it is stored horizontally! I think this is wrong, since we typically think of the vertical direction as being associated with the continuous dimension. I will file this as a separate low priority bug/feature request.

jrus:
I’m sure it’s fine either way. My interest is mostly idle curiosity.

About array-valued chebfun orientation generally, I think there’s some substantial code cleanup and bugfixing that can be done in several parts of Chebfun. As on example, here’s an issue about some related weirdness: https://github.com/chebfun/chebfun/issues/1606

W/r/t performance, is that something that appears to have some hotspots that can be optimized, or does the coefficients version look inherently slower?

ajt60gaibb:
I suggested 1) in #1542 and we left the idea because it was slower.  What changed? 

2) is the skinny QR algorithm (as described by Demmel and probably others).  I like that idea too.

nickhale:
Sorry, yes, I should have referenced #1542 earlier. (I used pretty much the same code as from there.)

@jaurentz convinced me it was better to work in coefficient space, but I think the performance hit is too great. I have reverted to values in ec017b4.

ajt60gaibb:
OK. So the situation is still the same as mentioned in #1542.  Thanks.

aaustin141:
I was going to merge this, but the tests `chebfun2v/test_curl.m` `chebfun2v/test_threecomponents.m` `chebfun2v/test_twocomponents.m` crashed.  The failures come when computing the norm of an identically-zero rank-1 `chebfun2`, and they stem from the fact that `qr(chebfun(0))` now results in a crash:
```
>> qr(chebfun(0))
Error using chebfun/rdivide>columnRdivide (line 138)
Division by zero.
Error in ./ (line 72)
        h = columnRdivide(f, g, pref);
Error in chebfun/qr (line 29)
    Q = A./R;
```
Previously, we got this:
```
>> qr(chebfun(0))
ans =
[      -1,       1]        1       NaN      NaN
vertical scale =   0
```
This result is pretty much nonsense, but `chebfun2.svd()` (called by `chebfun2.norm()`) seems to be able to work with it.

Any thoughts?

nickhale:
I think this is the right thing to do:
```
>> [Q, R] = qr(chebfun(0))
Q =
[      -1,       1]        1      0.71     0.71 
vertical scale = 0.71 
R =
     0
>> Q'*Q
ans =
   1.000000000000000
```
Compare with:
```
>> [Q, R] = qr(0)
Q =
     1
R =
     0
```

aaustin141:
Sounds sensible to me.

PR #1755, opened 2016-02-10
Feature promote

nickhale:
Allow the solution of problems like
```
N = chebop(@(u) diff(u,2) + sum(u), [-1 1], 0);
u = N\1;
```
and
```
N = chebop(@(u) diff(u,2) + u.*sum(u), [-1 1], 0);
u = N\1;
```


Hadrien-Montanelli:
@nickhale, I ran the tests and `chebop/test_autoVectorize` failed.

nickhale:
This actually has nothing to do with `autovectorize`, it's to do with non-linear boundary conditions for a system of equations. (Is this really the only place where this get tested!?)

What happens is that the boundary conditions are initialised with an `operatorBlock`, and when we multiply by a functional block, this gets promoted (which is not the desired behaviour here).

Here's a simple example which breaks:
```
N = chebop(@(x,u,v) [diff(u) ; diff(v)]);
N.bc = @(x,u, v) [u(0)-1; u(0).*v(.5)];
size(matrix(linearize(N), 3))
u = N\1
norm(N(u) - 1, inf)
```

nickhale:
I found a fix for the example above. The problem was actually that these lines:
```
f.jacobian = operatorBlock.mult(f.func, f.domain)*g.jacobian + ...
                    operatorBlock.mult(g.func, g.domain)*f.jacobian;
```
of `adchebfun/mtimes()` produce an operatorBlock even for scalar.*scalar because `operatorBlock.mult(f.func)` always returns an operatorBlock, even when `f.func` is a scalar. My fix for this (which I think is OK?) is for `operatorBlock.mult(u)` to simply return `diag(u)` when `u` is a vector of numeric values.

This fixes the above problem, and doesn't appear to break anything else. A couple more tests have been added.

Hadrien-Montanelli:
Great, thanks @nickhale -- let me have a look at it.

Issue #888, opened 2014-06-17
Method declaration in CLASSDEF files


mhsnjvd:
Should we declare ALL methods of a class in the CLASSDEF file? Some classes do this, and some don't and it would be good to have consistency. 

However, at this stage, this is an extremely low priority thing.

PR #1758, opened 2016-02-10
fix-remove-classdef-method-decls

### LINKED ISSUE #1601: Remove template 'headers' from CLASSDEF files.
When @aaustin141 and I were removing epslevels, a major source of frustration was the templating in the CLASSDEF files for each of the classes.

Since it's not clear that we gain anything from having these here (the descriptions you see in `>> doc chebfun` appear to be sourced from the m-files themselves, rather than the descriptions in the CLASSDEF), I propose they be removed.

aaustin141:
Closes #1601 and #888 and saves nearly 1300 lines of code.


nickhale:
@aaustin141 - I notice that `checkSingTypes` in the `singfun` classdef file is never used. (I may have commented on this before.)

Can it be removed?

nickhale:
@aaustin141 - Another point: obviously the static methods need to be declared in the classdef files, but they're still templated even when implemented in other .m files. 

Is it worth considering replacing these occurrences with (admittedly) nasty `[varargout{1:nargin}] = staticMethod(varargin)` calls? (I'm not necessarily advocating this, just asking.)

kuanxu:
@nickhale ``checkSingTypes`` can be safely removed.

aaustin141:
I've now removed `checkSingType()`.  It always feels great to get rid of dead code!

> Is it worth considering replacing these occurrences with (admittedly) nasty `[varargout{1:nargin}] = staticMethod(varargin)` calls? (I'm not necessarily advocating this, just asking.)

It might be.  One possible disadvantage to doing this is outlined in [this comment](https://github.com/chebfun/chebfun/issues/1601#issuecomment-139135169) from #1601, but I just spent a few minutes experimenting, and it doesn't seem to be an issue.  Overall, I don't feel as inclined to make this change, since the result is kind of ugly, and since there are far fewer static methods, but I could imagine that if I were to get tripped up by issues related to these things again in the future, I'd consider coming back and doing it.

Curiously if you declare a method in the classdef file as `varargout = method(varargin)`, you get errors about "too many arguments" if the M-file prototype has, say, 2 arguments and you call it with 3.  In this case, MATLAB seems to have resolved things in favor of the M-file prototype, whereas in the past, it seems to have resolved things in favor of the classdef prototype.  (For instance, if the classdef file prototype specifies 2 arguments and the M-file one gives 3, then you get a "too many arguments" error when you call the method with 3 arguments.)  Maybe it resolves in favor of whichever prototype is "more restrictive"?

PR #1759, opened 2016-02-10
Fix help text for ultraS.convertmat().

aaustin141:
See comments on #1729 and #1509.  This is documentation that occurs in a part of the system that I don't completely understand, so I'm issuing a pull request for it.  It might actually be a better idea to tweak `ultraS.convertmat()` to do what its help text originally said, as that interface would be less confusing.


nickhale:
I agree that in the long term it would be better to adjust the functionality of this function, but in the short term at least making the documentation match reality is still an improvement.

PR #1760, opened 2016-02-11
feature-more-accurate-leg2cheb

nickhale:
Use Reinsch's modification of the standard Legendre recurrence relation for more accurate computation of the 'direct' components in these transforms. (As proposed by @MikaelSlevinsky in his recent [paper](http://arxiv.org/abs/1602.02618).)


MikaelSlevinsky:
Wow, that was **fast**. P.S. I computed your test value on line 46 in extended precision:
```julia
julia> using ApproxFun

julia> N = big(1000)

julia> c_leg = 1./collect(N:-1:1).^2; c_leg[2:2:end] = -c_leg[2:2:end]

julia> f = Fun(c_leg,Legendre(Interval{BigFloat}()))

julia> g = Fun(x->f(x),Chebyshev(Interval{BigFloat}()),1000)

julia> g.coefficients[559]
6.379508600676002013455006832858066795950416371825174435785085364843969616272414e-04
```
I don't know if that will allow the tolerance to be reduced. I get a little over 14 digits:
```julia
julia> using FastTransforms

julia> c_cheb = cjt(map(Float64,c_leg),0.,0.);

julia> Float64(g.coefficients[559]./c_cheb[559]-1)
9.465349032387583e-15

```

ajt60gaibb:
Great.

ajt60gaibb:
You may want to remove the `foo.m` file.

PR #1763, opened 2016-02-15
Provide code for computing roots of Bessel functions at user level.

nickhale:
Make the code for computing Bessel roots available at the user level.

This also removes duplication, since `jacpts` and `ultrapts` (and to some extent `legpts`) previously implemented this same code locally.


nickhale:
I should note that this code was mostly written by @LourencoPeixoto .

ajt60gaibb:
I would prefer `besselrts`.

trefethen:
I prefer `besselroots` to `besselrts`.

LourencoPeixoto:
@nickhale, I'd like to recall that, except for `V=0` and `N<=20`, this code gives the **approximations** for the zeros. Furthermore I think it would be suitable to add/rewrite some comments like these in the beginning of the code:

```
J = BESSELPTS(V, N) returns the approximations for the first N roots the Bessel function J_V(X) according to the parameter V:

V = 0 ---> exact values in double precision for N<=20 (Wolfram Alpha), and very accurate approximations for N>20 (McMahon's expansion);
-1 <= V <= 5, V~=0 ---> the 6 first zeros to at least 12 decimal figures (Piessens's Chebyshev series approximations), and very accurate approximations for the others (McMahon's expansion);
V > 5 ---> moderate approximations for the 6 first zeros, and good approximations for the others (McMahon's expansion).
```

Issue #1764, opened 2016-02-15
Adjusting `eps` in Chebfun, Chebfun2, and Chebfun3


trefethen:
Traditionally -- i.e. in Chebfun[1] -- we have discouraged users from adjusting `eps` (except when dealing with noisy functions) on the grounds that there is not so much to be gained.  See Section 8.8 of the Guide at http://www.chebfun.org/docs/guide/.

In 2D and even more in 3D, it seems there will be much more to be gained by weakening the tolerance.  Indeed it may happen that we decide to use a default of, say 1e-8, in Chebfun3 -- lots to be thought about here!

Two practical matters come up immediately, more basic than changing the default.  First: both Chebfun2 and Chebfun3 need to accept a specification of `eps` in a call like `f = chebfun2(@(x,y) g(x,y),'eps',1e-8)`.  Action item 1: make this happen in Chebfun2.

Second: both Chebfun2 and Chebfun3 need to make sure that not only each 1D chebfun, but also the overall low-rank compression process respects the current value of Chebfun `eps`.  Currently Chebfun3 (not yet on GitHub... soon we hope!) does this but Chebfun2 does not.  Action item 2: make this happen in Chebfun2.

Here is a first experiment illustrating the modest gain from weakening `eps` in 1D.  The lengths
shorten as `eps` gets looser, and there is some time saved, but it's not a killer.

```
disp('   eps     time    length')
ff = @(x) tanh(100*(x-.3));
for chebeps = 10.^(-16:4:-4)
  chebfunpref.setDefaults('eps',chebeps)
  f = chebfun(ff);
  tic, f = chebfun(ff); t = toc;
  r = length(f);
  fprintf('%8.1e  %6.4f   %4d\n', chebeps,t,r)
end 

   eps     time    length
 1.0e-16  0.0251   2079
 1.0e-12  0.0186   1749
 1.0e-08  0.0188   1099
 1.0e-04  0.0135    377
```

In 2D, at present the improvement in lengths is there but there's no improvement in ranks and timings  because Action item 2 is not yet taken care of.  See the numbers below.  Once Action item 2 is taken care of, both lengths and ranks will diminish as `eps` is loosened and this will add up to a significant savings and speedup.  In Chebfun3 preliminary experiments suggest it will be dramatic indeed.

```
disp('   eps     time      rank m-length n-length')
ff = @(x,y) tanh(10*(x+y-.3));
for chebeps = 10.^(-16:4:-4)
[...]

Issue #1745, opened 2016-02-02
Calling constructor on a trigfun


gradywright:
Suppose we do
```
f = chebfun(@(x) cos(pi*x),'trig')

f =

   chebfun column (1 smooth piece)
       interval       length     endpoint values trig
[      -1,       1]        3        -1       -1 
vertical scale =   1 
```
Then for some reason we do
```
g = chebfun(f)

g =

   chebfun column (1 smooth piece)
       interval       length     endpoint values  
[      -1,       1]       29        -1       -1 
vertical scale =   1 
```
Now `g` is not a trigfun.  I seem to recall that the default behavior used to be to make `g` a trigfun since `f` is a trigfun, that is the constructor respects the type of the tech.  If memory serves me right @mhsnjvd made this change some time ago.

Looking through the code in a bit more detail it looks like if we call `chebfun(f)` where `f` is already a chebfun then we go through the whole construction process again.  Is this the desired behavior?

mhsnjvd:
I recall the opposite, i.e. when you call ```chebfun(f)``` where ```f``` is a trigfun, it constructs a chebfun from scratch using the default chebtech and this was desired because it gave us an easy way to switch from trig to cheb whenever operations on a ```trigfun``` give rise to discontinuities or destroy periodicity. For example, for ```abs(f)``` where ```f``` is a ```trigfun```.

aaustin141:
> Looking through the code in a bit more detail it looks like if we call chebfun(f) where f is already a chebfun then we go through the whole construction process again. Is this the desired behavior?

See issue #77.  Yes, this is what we want to do.  At the very least, there are several examples that make use of this functionality, and I'm pretty sure it's in ATAP as well.

aaustin141:
Can we close this issue?

gradywright:
Thanks for the responses @mhsnjvd and @aaustin141.  I will close the issue since this is the desired behavior.  

However, note that redoing the construction process and not respecting the tech type slows down `minandmax2@separableAppprox` in `spherefun`.  We might be able to work around this in another way.

PR #1765, opened 2016-02-15
Major overhaul of input parsing in chebfun2.

nickhale:
Major overhaul of `chebfun2` constructor input parsing. All the tests pass (remarkably..)

As a consequence we can now do this (see #1764).
```
>> chebfun2(@(x,y) 1./(1+25*(x.^2+y.^2)))
ans =
   chebfun2 object (1 smooth surface)
[  -1,   1] x [  -1,   1]       19     [0.02 0.02 0.02 0.02]
vertical scale =   1 
>> chebfun2(@(x,y) 1./(1+25*(x.^2+y.^2)), 'eps', 1e-8)
ans =
   chebfun2 object (1 smooth surface)
[  -1,   1] x [  -1,   1]        9     [0.02 0.02 0.02 0.02]
vertical scale =   1 
```

### DIFF
--- a/@chebfun2/constructor.m
+++ b/@chebfun2/constructor.m
@@ -109,12 +109,12 @@ while ( ~isHappy && ~failure )
     % Check if the column and row slices are resolved.
     colData.hscale = norm(dom(3:4), inf);
     colData.vscale = vscale;
-    colChebtech = tech.make(sum(colVals,2), colData);
-    resolvedCols = happinessCheck(colChebtech, [], sum(colVals, 2), colData, pref);
+    colTech = tech.make(sum(colVals,2), colData);
+    resolvedCols = happinessCheck(colTech, [], sum(colVals, 2), colData, pref);
     rowData.hscale = norm(dom(1:2), inf);
     rowData.vscale = vscale;
-    rowChebtech = tech.make(sum(rowVals.',2), rowData);
-    resolvedRows = happinessCheck(rowChebtech, [], sum(rowVals.', 2), rowData, pref);
+    rowTech = tech.make(sum(rowVals.',2), rowData);
+    resolvedRows = happinessCheck(rowTech, [], sum(rowVals.', 2), rowData, pref);
     isHappy = resolvedRows & resolvedCols;
     
     % If the function is zero, set midpoint of domain as pivot location.
@@ -127,6 +127,7 @@ while ( ~isHappy && ~failure )
     end
     
     %% %%% PHASE 2: %%%
+    pref.eps = tol;
     % Now resolve along the column and row slices:
     n = grid;  m = grid;
     while ( ~isHappy && ~failure  )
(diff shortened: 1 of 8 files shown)


ajt60gaibb:
That's glorious.  Thanks.

bhashemi:
I think in line 224 of `@chebfun2/constructor.m` in this branch, we need to replace
``
g = simplify( g, pref.eps );
``
by
``
g = simplify( g, tol );
``

@chebfun2/constructor.m
```diff
@@ -684,3 +523,171 @@
...
+    error('CHEBFUN2:DOMAIN:INFINITE', ...
+        'Chebfun2 cannot approximation functions on infinite domains.');
+end
+
+% Preferences structure given?
+isPref = find(cellfun(@(p) isa(p, 'chebfunpref'), varargin));
+if ( any(isPref) )
+    pref = varargin{isPref};
+    varargin(isPref) = [];
+elseif ( isa(dom, 'chebfunpref') )
```
aaustin141: I think this `elseif` clause can be removed:  execution never reaches here when we call the constructor in the case it says it's supposed to handle.

@chebfun2/constructor.m
```diff
@@ -684,3 +523,171 @@
...
+                    abs(round(d(1)) - d(1))< eps && ...
+                    abs(round(d(2)) - d(2))< eps)
+                [xx, yy] = chebfun2.chebpts2(d(1), d(2));
+                op = op(xx, yy);
+            else
+                error('CHEBFUN:CHEBFUN2:constructor:domain2', ...
+                    'Domain not valid or fully determined.');
+            end
+        end
+    elseif ( numel(d) == 1 )
```
aaustin141: This case does not get hit anywhere in the tests.

@chebfun2/constructor.m
```diff
@@ -684,3 +523,171 @@
...
+end
+
+isTrig = find(cellfun(@(p) any(strcmpi(p, {'trig', 'periodic'})), varargin));
+if ( isTrig )
+    varargin(isTrig) = [];
+    pref.tech = @trigtech;
+end
+
+isEpsGiven = find(cellfun(@(p) strcmpi(p, 'eps'), varargin));
+if ( isEpsGiven )
```
aaustin141: This case does not get hit anywhere in the tests either.

aaustin141:
OK.  I've fixed the issues that I pointed out previously and am generally very happy with this.  This is significantly better than what was there before.

One issue that made evaluating this pull request difficult is that the documentation for the `chebfun2` constructor needs some work.  For instance, the `'padua'` option isn't even listed; I had to look at the tests to figure out how to use it.  If you use it improperly, the error message is less than helpful.  In general, I think the checking of error conditions in the `chebfun2` input parser could be substantially improved, but it's hard to do this without a clear sense of what should and should not work.

Nevertheless, as this definitely improves on the original and doesn't, as far as I can tell, break anything or make anything worse, I'm happy to put it through once someone confirms or denies @bhashemi's comment on the call to `simplify()`.  (I think he's right.)

gradywright:
The problem with doing
```
Cvals = chebfun(C, m).values; 
Rvals = chebfun(R, n).values; 
```
Is that if `C` and `R` are trigfuns then they get converted to chebfuns when you call the constructor (see issue  #1745), so you have not really gained any speed up over just calling `feval` on `C` and `R`.

By having a `sample` method at the `tech` level, the individual techs can generate values at a set of points that are the most computationally efficient to compute.  However, it looks like the current implementations for `sample` in `chebtech1`, `chebtech2`, and `trigtech` are not taking advantage of any of these speed ups as they are just calling `feval`.  I think using `alias` and `coeffs2vals` provides a more efficient implementation.

nickhale:
@aaustin141 - I think I tried using `tol`, rather than `pref.eps` and it broke a number of tests.

If we do decide to use `tol`, then I think we should set `pref.eps = tol` before the start of stage 2.

aaustin141:
@nickhale:  Your latest commit comments out the `simplify()` line entirely, and that causes the second-to-last test in `tests/chebfun2/test_constructor.m` to fail for me.  If I comment it back in and do the simplification with either `tol` or `pref.eps`, I get failures in `tests/chebfun2/test_guide.m` and `tests/chebfun2/test_plus.m`.  If I just do `g = simplify(g)`, as was there originally, everything passes.

I suggest that we revert that line to `g = simplify(g)` and save these issues for another ticket.

nickhale:
Sorry - that was some left over debugging.

Issue #1233, opened 2014-10-23
chebfun2v.roots depends inappropriately on persistently-stored tech preference settings.


aaustin141:
If I do

```
>> chebfunpref.setDefaults('factory');
>> p = chebfun2(@(x, y) x - y);
>> q = chebfun2(@(x, y) x + y);
>> roots([p ; q])
ans =
   1.0e-17 *
  -0.241731300270792  -0.268895580714789
```

everything is fine.  If I now do

```
>> chebfunpref.setDefaults('tech', @trigtech);
>> roots([p ; q])
```

MATLAB sits and spins.  (Presumably, this will terminate eventually, possibly with an error message, but I sent it Ctrl+C after it sat there for 60 seconds.)

The reason is basically [this line](https://github.com/chebfun/chebfun/blob/347b601f132690f8b4541312240ebcb07ea6c409/%40chebfun2v/roots.m#L1346).  `chebfun2v.roots` should not be checking the persistently-stored default tech preference to determine the techs of its operands.  Fixing this will require some thinking about the software architecture.

ajt60gaibb:
@aaustin141, do you know how to fix this now that we have #1768?

Issue #1081, opened 2014-07-24
Issue with use of chebpolval2 and 'periodic' chebfun2


gradywright:
The `@chebpolyval2` method is used in the `max`, `median`, `sampleTest`, and `vscale` methods for evaluating a `chebfun2` at a tensor product Chebyshev grid.   If one constructs the `chebfun2` with the 'periodic' option, then a conversion is required from Fourier to Chebyshev every time any of these methods is called, which can significantly slow down any operation (such as a simple call to `display`).  There is no reason that a tensor product 'Fourier' grid (i.e. equally spaced) could not be used in the 'periodic' case.

I propose we deprecate `chebpolyval2` and create the new methods `@chebfun2/sample`, `@chebfun/sample`, `@bndfun/sample`,  `@chebtech1/sample`,  `@chebtech2/sample`, and `@fourtech/sample`.  Then instead of calling `chebpolyval2` to get the values on a tensor product grid, one calls `@chebfun2\sample`, this method then calls the `@chebfun/sample` method on the rows and columns, which propagates the calling sequence down to the appropriate tech level.  The sample function at the tech level takes as input the number of points `m` to evaluate the tech at on its 'natural' grid.  Piecewise chebfuns are evaluated on a `m`-point natural grid corresponding to each piece.  

The purpose of the `sample` methods are to allow the underlying function to be evaluated much more efficiently than by calling `feval`.

gradywright:
Closed by #1768.

Issue #246, opened 2014-04-20
Should we remove f.values from chebtech?


ajt60gaibb:
In the feature_remove_vals branch (latest commit: f3e94d5ee7) I have removed the f.values property of a chebtech. So far all the chebtech tests pass (except one isnan test).  

I've put in a getvscl() function to calculate the vscale when it is required and the command times.m now uses coefficients with the Toeplitz-plus-Hankel-plus-rank-one matrix multiply. 

Once all the Chebfun tests pass we can start comparing against what we currently have.

PR #1768, opened 2016-02-18
Fix chebfun2/sample

### LINKED ISSUE #1766: Error using separableApprox/sqrt
[MeanCurvatureFlow.txt](https://github.com/chebfun/chebfun/files/131435/MeanCurvatureFlow.txt)
When running the following code, there is a change of sign error while taking a dot product, but the value of the dot product is positive throughout.

ajt60gaibb:
`chebfun2/sample` was assuming that the underlying techs were chebtechs.  Closes #1766.


@chebfun2/sample.m
```diff
@@ -17,16 +18,27 @@
...
+end
+if ( nargin > 2 ) 
+    n = varargin{2};
+end
+
+% Use CDR decomposition so we can keep it in low rank form: 
+[C, D, R] = cdr( f ); 
+Cvals = chebfun(C, m).values; 
+Rvals = chebfun(R, n).values; 
+
```
aaustin141: Directly grabbing the `values` from a `chebfun` is a no-no:  it breaks encapsulation.  As the documentation for `chebfun.get()` states, the only reason we support this in the first place is for backward-compatibility reasons, and I'd like to discourage its use internally as well as externally.  Is there a reason not to just call `feval()` here, like we discussed in #1766?

ajt60gaibb:
Oh, ok.  Maybe you would like to put in feval() then...  I never really understood what is and isn't encapsulation.

aaustin141:
All right, I've "fixed" this.  I ended up not using the `feval()` approach and instead went for adding a `sample()` function to `chebtech1

gradywright:
@aaustin141 This is something I've been wanting for a long time.  Is it only `chebtech1` that has this functionality?  I would also like to add this to `trigtech`.

gradywright:
See #1081.

aaustin141:
@gradywright:  No, it's not just `chebtech1`.  My previous comment (which has now been edited) accidentally got posted in a half-written state because I fat-fingered something on my keyboard.

ajt60gaibb:
I am not sure why @aaustin141's solution is good practice, while drilling down with `.values` on a chebfun was a total no-no. At some stage one needs to break encapsulation (if that's the right word) and assume an underlying grid.  But, I am happy with any solution that works. 

There are quite a few places where `separableApprox` could use these sample functions. For example, in `minandmax2` (instead of using Chebyshev grids when the function is complicated). I think 2D and 3D parts of the code need to sometimes know about the underlying grids.

gradywright:
I think there is another reason that @aaustin141's solution is better than drilling down to the values.  The issue with doing
```
[C,D,R] = cdr(f);
Cvals = chebfun(C, m).values; 
Rvals = chebfun(R, n).values; 
```
Is that it doesn't exploit any fast sampling at the `tech` level.  If `C` or `R` are `trigfuns` then calling the above calling sequence above will just evaluate them at Chebyshev points and no fast transform is used.  

By having a `sample` function at the `tech` level, the `techs` can exploit any fast evaluation methods.  For example evaluating at equally spaced points for `trigtechs`, or 2nd kind Chebyshev points for `chebtech2`.

Unfortunately, the implementations by @aaustin141 in [71f6e9e](https://github.com/chebfun/chebfun/commit/71f6e9e789fd09545a7f6823b63fbe869d742602) do not exploit any fast algorithms at the `tech` level.  For example, `chebtech2/sample` has the implementation
```
points = chebtech2.chebpts(n);
values = feval(f, points);
```
It would be more efficient to use `alias` and `coeffs2vals`.

aaustin141:
> I am not sure why @aaustin141's solution is good practice, while drilling down with .values on a chebfun was a total no-no.

Leaving the broader question of whether a "values" should be part of the external interface of a `chebfun` aside, the main reason why accessing this information through a method with a well-defined interface is better than just grabbing the property directly is that it avoids directly coupling `chebfun2` to how the techs work _internally_.  In general, you can't be assured that the tech at the bottom layer even has a `values` property to begin with, and even if it does, you can't assume that it means the same thing for all techs or that it's stored in the way you expect.

To be more concrete, suppose that tomorrow we decided to transpose the way the values are handled inside `chebtech`.  If `chebfun2` grabs the values directly, that change will break `chebfun2`.  If `chebfun2` accesses the values only via the special-purpose method, the change won't break `chebfun2` because the special-purpose method's implementation will be changed to keep it functioning in the same way from an external perspective.  As a result, we wanted to make a change to `chebtech`, and the only thing we had to change was `chebtech`.  That, in general is a good thing.

> At some stage one needs to break encapsulation (if that's the right word) and assume an underlying grid.

I disagree with this statement.  What about a `chebfun` based on a tech that uses mesh-free approximation methods?  For that matter, what about current `chebtech`, which is entirely based on coefficients since #246 and #312 were closed?  In fact, I even remember at one point discussing with @nickhale the idea of doing away with `chebtech1` and `chebtech2` and just having the values be a construction-time preference, since the grids essentially don't exist after construction-time anymore anyways.  Finally, there is the basic principle that, from the outside, `chebfun`s are supposed to act like "continuous" objects, and having "values" be part of the `chebfun` interface strikes me as undesirable from that perspective.

I think the statement you're really trying to make here is that `chebfun2` _specifically_ needs to assume an underlying grid.  If that's the case, then it is my view that the right place to deal with this is really inside `chebfun2`, not in `chebfun`.  `chebfun2` already gets grid information at construction time.  The problem is that it eventually discards it.  If it really needs that information at a later date, it should store it in some appropriate form, perhaps a handle to the constructor for the relevant tech.  Techs suitable for use with `chebfun2` will need to satisfy the requirements of a (currently nonexistent) abstract class that specifies exactly what that means.  Part of that may be having such `sample()` function like the one being discussed (in order to do the required evaluations efficiently), but now that would exist only in the lower layers, where behaving discretely is a little more acceptable.

There are other signs that this is the "proper" approach to dealing with this, as this is the same fundamental issue that underlies #1233 and #1772.

Now, I recognize that a change along those lines is likely to require more thought/work than we want to expend on it at the moment (cf. @gradywright's [comment](https://github.com/chebfun/chebfun/pull/1772#issuecomment-186497460) on #1772).  The way that I've proposed to do it is a compromise that will at least allow us to fix a few things in the short-term.  Long-term, though, something along the above lines may be better.  That is, it might be better for us just to acknowledge that `chebfun2` cannot truly be completely tech-agnostic and let it deal with that in its own way rather than pollute the interfaces of the parts of the system that can.

nickhale:
At the risk of wading in on this argument without fully understanding what's going on, @aaustin141's  approach (once @gradywright's [comment](https://github.com/chebfun/chebfun/pull/1768#issuecomment-187875174) has been addressed) approach seems elegant. (Although the comments discussing what an "appropriately-chosen grid" is are sufficiently wishy-washy as to be almost meaningless. It might be better to add an honest developer note saying precisely why this method is required.)

I also agree with his comments regarding encapsulation.

ajt60gaibb:
I am a little lost in the ocean of arguments here.  I clearly do not fully understand `encapsulation` so I am tapping out.  I am happy as long as whatever is in the code returns the correct answer and @aaustin141 is OK with the solution.

aaustin141:
@gradywright:  See my new commits.  I've now changed the implementations to use `alias()` and `coeffs2vals()` as you suggest.

@nickhale:  See [these lines](https://github.com/chebfun/chebfun/blob/fix-chebfun2-sqrt/%40chebfun/sample.m#L15-L28) in `@chebfun/sample.m` for the requested developer note.  I agree that the comments I've written about what an "appropriately-chosen grid" is are mostly meaningless.

gradywright:
@aaustin141 This looks great, thanks!  Once we have this merged I can update `Spherefun' and `Diskfun` to use the new `sample` method.

nickhale:
One small matter:

It seems that in (currently) the only use of this functionality, the points are not required, only the values. It might therefore make sense to return the values as the first output argument and only compute the points if required?

aaustin141:
@nickhale:  Done.  I'm not sure why I didn't return the outputs in the other order in the first place.

PR #1770, opened 2016-02-18
removed "(1 smooth surface)" from @chebfun2/disp.m

trefethen:
This does what it says on the tin: removes that pesky expression "1 smooth surface" from every display of a chebfun2.   (This is my first pull request!)

### DIFF
--- a/@chebfun2/disp.m
+++ b/@chebfun2/disp.m
@@ -26,7 +26,7 @@ vals = vals(:);
 vscl = vscale(F);                         % vertical scale
 
 % Display the information: 
-disp('   chebfun2 object (1 smooth surface)')
+disp('   chebfun2 object')
 fprintf('       domain                 rank       corner values\n');
 if ( isreal(vals) )
     fprintf('[%4.2g,%4.2g] x [%4.2g,%4.2g]   %6i     [%4.2g %4.2g %4.2g %4.2g]\n', ...


ajt60gaibb:
OK. Should the analogous phrase be removed from `chebfun/disp`?  See, for example:

```
f = chebfun('x')
f =
[      -1,       1]        2        -1        1 
vertical scale =   1 
```

trefethen:
No, in the 1D case the "1 smooth piece" message has content: sometimes there are multiple pieces.
```
>> f = chebfun('abs(x)','splitting','on')
f =
[      -1, 1.2e-17]        2         1        0 
[ 1.2e-17,       1]        2         0        1 
vertical scale =   1    Total length = 4
```
So in 1D there is genuine information involved, whereas in 2D I believe there is no such information since *all* functions are smooth.  Please correct me if I'm overlooking something.

ajt60gaibb:
OK.  After so many years of having `(1 smooth surface)` in the display, it looks empty without it.  But, I agree, it should be removed until there is a piecewise chebfun2.

bhashemi:
A similar issue exists in ``spherefun`` which prints ``spherefun object (1 smooth surface) ``...

gradywright:
Thanks @bhashemi.  I made the change on the `feature-spherefun-new` branch.

PR #1771, opened 2016-02-18
Fix chebfun2 grids

ajt60gaibb:
If `minSample` in the preference structure is not a power of 2 plus one (for Chebyshev) or not a power of 2 (for Trig), then we round it down to the smallest number that is.    

This is fixing a strange bug when in `periodic` mode.


aaustin141:
I'm going to hold off merging this until I've looked at #1765.  In fact, it might be best just to scrap this and redo the changes once that's done (should be pretty easy).

ajt60gaibb:
Yeah, this pull request is only a two line change.

aaustin141:
Closed in favor of #1773.  (Post-#1765, this branch now has conflicts.)

Issue #1573, opened 2015-08-06
Remove chebfun2v dependence from the abstract separableApprox class


gradywright:
With the merge of #1471, we now have an abstract class `separableApprox` which technologies similar to `chebfun2` can inherit from.   However, there are several places where `separableApprox` inherently relies on `chebfun2v`.  These dependencies should be removed and a new class hierarchy should be developed for vector-valued functions similar to `separableApprox`.  Below is a list of specific dependencies:
1.  How do we handle line 121 of `separableApprox/feval` when the input is a chebfun2v object?
2.  `grad` and `gradient` return chebfun2v objects.  These functions may depend heavily on the topology of the underlying `separableApprox` domain.
3.  Need to figure out a general way for `separableApprox/mtimes` to handle the case where the input is a chebfun2v object.
4. `separableApprox/quiver`, `separableApprox/quiver3.m`, and `separableApprox/roots` make heavy use of `chebfun2v`.
5. The help text and code in `separableApprox/subsref` make explicit use of `chebfun2v`.
6.  `separableApprox/vertcat` explicitly constructs chebfun2v objects.  Can this be made more general?

PR #1772, opened 2016-02-18
Fix chebfun2v/dot: Respect underlying tech

ajt60gaibb:
The `chebfun2v/dot` command now respects the underlying tech.  

Now we have
```
chebfunpref.setDefaults('tech',@trigtech)
x = chebfun2(@(u,v) (3+cos(v)).*cos(u), [-pi pi -pi pi]);
y = chebfun2(@(u,v) (3+cos(v)).*sin(u), [-pi pi -pi pi]);
z = chebfun2(@(u,v) sin(v), [-pi pi -pi pi]);
S = [x; y; z];

E           =   dot(S,S);
[C,D,R] = cdr( E ) 
C =
       interval       length     endpoint values  trig
[    -3.1,     3.1]       10     -0.15    -0.15 
vertical scale = 0.59 
D =
  68.252862777426245
R =
       interval       length     endpoint values  trig
[    -3.1,     3.1]        4      -0.4     -0.4 
vertical scale = 0.4 
```
showing that periodic output was calculated from periodic input.

### DIFF
--- a/@chebfun2v/dot.m
+++ b/@chebfun2v/dot.m
@@ -1,4 +1,4 @@
-function f = dot( F, G )
+function H = dot( F, G )
 %DOT   Vector dot product.
 %   DOT(F, G) returns the dot product of the CHEBFUN2V objects F and G. DOT(F,
 %   G) is the same as F'*G.
@@ -8,13 +8,20 @@ function f = dot( F, G )
 % Copyright 2015 by The University of Oxford and The Chebfun Developers.
 % See http://www.chebfun.org/ for Chebfun information. 
 
-if ( isempty( F ) || isempty( G ) ) 
-    f = chebfun2();
+if ( isempty( F ) ) 
+    H = F;
+    return
+end
+
+if ( isempty( G ) ) 
+    H = G;
     return
 end
 
 nF = F.nComponents; 
 nG = G.nComponents; 
+
+% Check that F and G have the same number of components:
 if ( nG ~= nF ) 
     error('CHEBFUN:CHEBFUN2V:dot:components', ...
         'CHEBFUN2V object should have the same number of components.');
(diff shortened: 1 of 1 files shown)


aaustin141:
This doesn't seem to work for me.   Try running the following code:
```matlab
chebfunpref.setDefaults('factory');

f = chebfun2(@(x, y) sin(pi*x).*cos(pi*y), 'trig');
class(f.cols.funs{1}.onefun)

F = [f ; f ; f];
F1 = F(1);
class(F1.cols.funs{1}.onefun)

u = dot(F, F);
class(u.cols.funs{1}.onefun)
```
This prints
```
ans =
trigtech
ans =
trigtech
ans =
chebtech2
```
for me, both on this branch and on development.

ajt60gaibb:
@aaustin141, I did not intend it to work in your case because you have not set the global preference using `chebfunpref.setDefaults('tech',@trigtech)`. 

To get your example to deliver a `trigtech`, we will have to decide how to pass the periodic information around and into `compose`. This requires a much larger and more careful change to the code. Consider, for example, 
```
f = chebfun2(@(x,y) cos(pi*x),'trig'); 
g = f.*f; 
class(g.cols.funs{1}.onefun)
ans =
chebtech2
```

gradywright:
I agree with @ajt60gaibb that to get the example to work will require a _significant_ change to the code.

gradywright:
I agree with @ajt60gaibb that to get the example to work will require a _significant_ change to the code.

gradywright:
I agree with @ajt60gaibb that to get the example to work will require a _significant_ change to the code.

aaustin141:
If that's the case, then this pull request actually fixes nothing because things have always worked when the default `tech` was set to `trigtech`.  The problem was that it doesn't work when that's _not_ the case.

ajt60gaibb:
The code now works when the components are any `separableApprox` objects, not just `chebfun2` objects.

aaustin141:
Fair enough.  I'd forgotten that `chebfun2v` could work with general `separableApprox` objects.  Cross-ref. #1573.

aaustin141:
Sorry---I botched this.  I'll re-do this merge shortly.

Issue #1776, opened 2016-02-29
Strange chopping in chebfun2 constructor


jaurentz:
Constructing the following `chebfun2` leads to a strange coefficient plot:
```matlab
>> f = chebfun2(@(x,y) exp(sin(pi*(x+3*y)))); plotcoeffs(f)
```
![strangecoeffs](https://cloud.githubusercontent.com/assets/7694954/13397200/cf94e478-def0-11e5-9973-31edee551661.png)
This should be no problem for `standardChop`.

aaustin141:
Closed in #1777.

PR #1777, opened 2016-02-29
Fix chebfun2 simplify tol

jaurentz:
This pull request addresses the problem in #1776. The key problem, as pointed out by @bhashemi, is that the `tol` computed during phase 1 of the `chebfun2` constructor is not being passed on to the `simplify` call at the end of phase 2. This can result in a tighter simplification tolerance than what was ultimately used for construction.


aaustin141:
This change looks innocuous enough, but the last commit highlights a few things that I don't understand.  Is the tolerance `tol` supposed to be absolute or relative?  That is, is it supposed to incorporate the `vscale` or not?

One would think this question would be answered by [this line](https://github.com/chebfun/chebfun/blob/29b2386e1018928b32ae315b4f3cb16a93692c2d/%40chebfun2/constructor.m#L526) in `getTol()`, which clearly builds the `vscale` into `tol`.  That also seems to be consistent with its usage in, say, the `completeACA()` subfunction and in `separableApprox.sampleTest()`.  But then, I see things like [this line](https://github.com/chebfun/chebfun/blob/29b2386e1018928b32ae315b4f3cb16a93692c2d/%40chebfun2/constructor.m#L97) in which it's being used as if that's not supposed to be the case.  I also see `tol` being inserted into `pref.eps` in [this line](https://github.com/chebfun/chebfun/blob/29b2386e1018928b32ae315b4f3cb16a93692c2d/%40chebfun2/constructor.m#L130), but `pref.eps` is supposed to be a relative tolerance, not an absolute one.

What's going on here?  Are these latter lines just wrong?  @ajt60gaibb, can you comment on this?

ajt60gaibb:
@aaustin141, I have not been keeping up with the chopping/simplify changes.   I think @bhashemi and @jaurentz know far more about the tolerances in the chebfun2 constructor than I do at this point.

aaustin141:
Very well, then.  @bhashemi, do you have any comments to make about the role of `tol` here and, in particular, whether it is supposed to relative or absolute?

jaurentz:
I've updated the subroutine `getTol` to output both a vscale and hscale-invariant `relTol` and the original output which is now called `absTol`. There are no fancy tricks here, just a refactoring of the variables for clarity. Now whenever `getTol` is called the internal `pref.eps` is set to equal `relTol`. This means that both `happinessCheck` and `simplify` use `relTol` and the rest of the code uses `absTol` as it did before.

bhashemi:
@aaustin141 , considering these last changes, is this now clear enough?

bhashemi:
I think this PR needs more work before being merged, as #1776 still persists!

jaurentz:
Added the magic `grid^(2/3)` factor to the computed `relTol`. This fixes the problem in #1776.

PR #1779, opened 2016-03-03
Fix symmetric test in chol() command.

### LINKED ISSUE #1780: separableApprox/chol is broken
This used to work:
```
A = chebfun2(@(x,y) cos(10*x.*y) + y + x.^2 + sin(10*x.*y));
R = chol( A' * A );
norm( A'*A - R'*R ) 
ans =
   0.002494201668744
```

It is because the `CDR` factorization is not symmetric for `A'*A`: 
```
A = chebfun2(@(x,y) cos(10*x.*y) + y + x.^2 + sin(10*x.*y));
[C, D, R] = cdr( A'*A );
norm( C - R ) 
ans =
   0.046157076339960
```

ajt60gaibb:
The symmetric test should check the symmetry of the function on its domain, not outside.   

This also closes #1780.

### DIFF
--- a/@separableApprox/chol.m
+++ b/@separableApprox/chol.m
@@ -74,6 +74,7 @@ if ( nargout < 2 && ~isempty( posdef ) && ~posdef )
 end
 
 % Get the CDR decomposition (already computed by constructor):
+f = chebfun2(@(x,y) feval( f, x, y ), f.domain);
 [C, D, R] = cdr( f );
 
 % Return an error if the function is not symmetric: 
--- a/@separableApprox/chol.m
+++ b/@separableApprox/chol.m
@@ -80,8 +80,8 @@ end
 dom = f.domain; 
 r = 0.0192475;   % arbitrary point in [-1,1] 
 s = -.34756987;  % arbitrary point in [-1,1]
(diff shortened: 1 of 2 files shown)


@separableApprox/chol.m
```diff
@@ -74,14 +74,15 @@
 end
 
 % Get the CDR decomposition (already computed by constructor):
+f = chebfun2(@(x,y) feval( f, x, y ), f.domain);
```
aaustin141: Just one question:  why is this line necessary?  commenting it out has no effect.

aaustin141:
Hmmm... there seems to have been some strange hiccup.  @ajt60gaibb:  I have a notification e-mail from you indicating that you posted a comment here asking if the pull request was rejected.  For some reason, I don't see it showing up on the GitHub interface at all.  No, it was not rejected.  I merged it, but there was some other kind of hiccup, and for some reason GitHub marked the pull request as "closed" instead of "merged".

This has happened once or twice in the past, and I was able to fix it to get the proper resolution tag by simply reopening the pull request, after which it would immediately re-close itself as "merged".  But that seems not to have worked this time around...

Either way, I see ecf88388 as being on development both locally and on GitHub, so the merge _did_ go through.

PR #1795, opened 2016-03-28
Adjoints for linear differential operators

jaurentz:
This pull request allows for the automatic computation of adjoints of linear differential operators. Integral and other exotic operators are not supported. This also only supports periodic or endpoint boundary conditions.


jaurentz:
I think it would be best if #1788 was merged before this one.

trefethen:
@jaurentz, I'm teaching the Exploring ODEs course at NYU and in today's lecture I would have enjoyed showing the students adjoints, but discovered this is still in a branch.  Can you give us a status update?  Is this something that should be mergeable into development?  What's your understanding of the state of this branch and its need for review?   - Thanks for any comments.

jaurentz:
The last person I talked with about code review was @Hadrien-Montanelli who appears to still be assigned to the task. As far as I know that only issue is for someone to review it.

Hadrien-Montanelli:
@jaurentz Since the method chebop/getBCType is only used in chebop/adjoint, maybe it should be moved at the bottom of the chebop/adjoint file?

Hadrien-Montanelli:
@jaurentz Could you please add a couple of tests at the `chebop/adjoint` level?

Hadrien-Montanelli:
All the tests pass - happy to merge it once there are more tests

jaurentz:
Added a few more tests and hid some of the low level files inside the main adjoint routines.

Hadrien-Montanelli:
@jaurentz brilliant - running the tests

PR #1797, opened 2016-03-29
fix-mixed-tech-innerProduct

aaustin141:
This is a re-do of #1787 that moves the tech type-casting in `innerProdcut()` to `chebfun`, where it belongs, instead of `chebtech` and `trigtech`.  We do all other casting for binary operations with mixed tech types in `chebfun`, so this should be done there as well.  Also adds some tests.

### DIFF
--- a/@chebfun/innerProduct.m
+++ b/@chebfun/innerProduct.m
@@ -39,6 +39,14 @@ out = zeros(numColsF, numColsG);
 
 if ( numel(f) == 1 && numel(g) == 1 )
     % Array-valued CHEBFUN case:
+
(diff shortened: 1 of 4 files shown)


jaurentz:
Thanks for the clean up @aaustin141.

jaurentz:
In `chebfun/test_trigcasting` I get a failure at `pass(17)`. It appears to be a domain misalignment problem. This kind of test seems too demanding to me. Maybe we should check that the domains are close and then check a single point in each subinterval?

jaurentz:
Hmm.... this has nothing to do with your pull request, more likely a bug with the Matlab 2016a pull request. Sorry for the chatter.

PR #1804, opened 2016-03-31
Add a basic ULTRACOEFFS() code.

nickhale:
Nothing fancy. Just a wrapper for `jaccoeffs()`, but it's useful to me and saves having to manually adjust the coefficients each time.

Includes a basic test.


nickhale:
I added a note that one should not expect great accuracy for lambda >> 1. Does that suffice?

PR #1809, opened 2016-04-04
JACPTS() calls ULTRAPTS() for ALPHA=BETA and ALPHA > -1

LourencoPeixoto:
1. `JACPTS()` calls `ULTRAPTS()` if` ALPHA = BETA` for `ALPHA > -1` and `ALPHA ~= -.5, 0, .5`;

2. Add some comments about range of `LAMBDA`.


ajt60gaibb:
@LourencoPeixoto, this looks good to me.

PR #1812, opened 2016-04-06
Fix clenshaw docs

nickhale:


bhashemi:
@nickhale I am interested to learn what we gained by reordering coefficients in Clenshaw. Would you please give me a hint?

nickhale:
The is just a change in some commented code. It wasn't updated when the order of the way we store Chebyshev coefficients was changed some time ago.

bhashemi:
OK. Thanks.

Issue #1814, opened 2016-04-14
Enforcing even/odd symmetries in trigfun (and Chebfun)


trefethen:
We already force the trigfun of a real function to be real (this required taking action, as the FFT introduces complex errors).  We don't, however, force the trigfun or the chebfun of an even or odd function to preserve that symmetry.

In the case of chebfuns, you have to look hard to find the loss of symmetry.  The first integer k for which cos(k*x) doesn't come out even is 1471 (at least on @jaurentz's Mac).  Similarly the first integer k for which sin(k*x) doesn't come out odd is 1429.

With trigfuns, however, the anomalies appear immediately.  cos(pi*x) is not even, and sin(pi*x) is not odd.  This is a "cosmetic" problem but a rather glaring one.  We propose to fix it for trigfuns, and while we're at it, likely do the same for chebfuns even though in that case the problem arises more rarely.

Here is our plan: (a) understand where the losses of symmetry are introduced; (b) either fix this, or work around it by making Chebfun test the input samples for symmetry and when it finds it, impose the appropriate symmetries on the output.  To be concrete: sampling cos(k*x) at chebpts(n) gives a perfectly even vector regardless of k and n.  The problem is that somewhere along the line, Chebfun loses that symmetry.

We'd be glad for comments on this project -- e.g. from @gradywright, who was probably the one who enforced reality for trigfuns of real functions.

One reason this is on our minds is that we've been working on tidying up the anomalies in `plotcoeffs` when it deals with coefficients that are exactly zero.  For example, the code
```
subplot(2,1,1), plotcoeffs(chebfun('cos(1470*x)'),'.')
subplot(2,1,2), plotcoeffs(chebfun('cos(1471*x)'),'.')
```
at present produces the figures below.  Our plotcoeffs work will improve the plotting of the first figure; the present issue is concerned with eliminating the disparity between the first and second figures.
![cosfig](https://cloud.githubusercontent.com/assets/3601488/14522904/e282e444-0228-11e6-84f6-1b407bdf6fa7.png)

jaurentz:
It appears that the problem is that Matlab's `fft` does not preserve symmetry. See the following example:
```matlab
>> x = linspace(-1,1,100)'; x = (x-x(end:-1:1))/2; % enforce symmetry
>> vals = sin(pi*x); norm(vals+vals(end:-1:1)) % test symmetry (should be zero)
ans =
     0
[...]

PR #1818, opened 2016-04-18
Fix trigpts symmetry

jaurentz:
This pull request enforces symmetry in `tripts`, `trigtech/vals2coeffs` and `trigtech/coeffs2vals`. See #1814.

### DIFF
--- a/@trigtech/coeffs2vals.m
+++ b/@trigtech/coeffs2vals.m
@@ -38,7 +38,18 @@ else
 end
 coeffs = bsxfun(@times, coeffs, even_odd_fix);
 
+% test for symmetry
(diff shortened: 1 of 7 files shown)


jrus:
What’s the performance penalty of these checks and corrections? It seems like there’s a fair amount of overhead even when there’s no symmetry found.

aaustin141:
The overhead doesn't seem to be too bad, at least if we measure in the (admittedly flawed) "chebtest metric".  On current development (a07d5a9f), which includes these changes as well as those from #1822, I find that `chebtest('chebfun')` takes about 112 s to run on my workstation.  On 438aa659, before these changes, I find that it takes about 107 s to run, so that's a slowdown of about 4-5%.

PR #1835, opened 2016-04-22
Fix chebfun2 roots

ajt60gaibb:
The following was failing: 
```
f = chebfun2( @(x,y) 1/2-y);
r = roots( f );
```
because of an internal bug when `f` is of rank 1.  Fixed now.

### DIFF
--- a/@separableApprox/roots.m
+++ b/@separableApprox/roots.m
@@ -42,11 +42,19 @@ if ( nargin == 1 )
         
         cols = f.cols;
         rows = f.rows;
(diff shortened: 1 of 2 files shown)


gradywright:
Looks like I merged this pull request a bit too soon.  If one does 
```
f = chebfun2( @(x,y) 1/2-y, [-2 2 -1 1] );
r = roots( f );
plot(real(r),imag(r))
```
The zero contour goes from `[-1,1]` in `x` instead of `[-2 2]`, i.e. r does not have the correct domain.

PR #1854, opened 2016-04-30
Implement wave number scaling for trigtechs.

### LINKED ISSUE #1853: bug in label of trigfun `plotcoeffs`
When `plotcoeffs` is applied to a trigfun, the $x$ axis is labeled "wave number".  However, this is not the wave number as is universally understood.  For example, $\sin(x)$ should have wave number 1 on any domain.  In Chebfun, this only happens if the domain is $[-\pi,\pi]$.  Thus e.g. the commands
```
f = chebfun(@(x) sin(x),[-10*pi,10*pi],'trig'); plotcoeffs(f)
```
give the erroneous plot
foo

This would cease to be an outright error if the $x$ axis were relabled to something like "normalized wave number".  However, please let's not do that.  I hope instead we can maintain the standard convention and keep the label but change the numbers: the numbers currently plotted need to be multiplied by $2\pi$ divided by the length of the domain.

nickhale:
Closes #1853.

### DIFF
--- a/@chebfun/plotcoeffs.m
+++ b/@chebfun/plotcoeffs.m
@@ -28,7 +28,7 @@ if ( isempty(f) )
     return
 end
 
-dom = f.domain([1,end]);
+dom = f(1).domain([1,end]);
 
 % We can only plot the coefficients of one CHEBFUN at a time:
 if ( any(cellfun(@(f) isa(f, 'chebfun'), varargin)) )
--- a/@chebfun/plotcoeffs.m
+++ b/@chebfun/plotcoeffs.m
@@ -28,6 +28,8 @@ if ( isempty(f) )
     return
 end
 
+dom = f.domain([1,end]);
+
 % We can only plot the coefficients of one CHEBFUN at a time:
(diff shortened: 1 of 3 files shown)


trefethen:
@nickhale, can you explain?  In the comment by @gradywright in issue #1853, it seemed there was a difficulty to implementing this.  Have you worked around this, and can you comment?  Thanks.

gradywright:
@nickhale I think that `chebtech/plotcoeffs` has to also deal with the new `domain` property.  The following causes and error:
```
>> f = chebfun(@(x) cos(100*x));
>> plotcoeffs(f)
```

gradywright:
@nickhale There seems to be an issue with `plotcoeffs` of a quasimatrix:
```
>> Q = cheb2quasi(chebfun(@(x) [sin(x) cos(x) exp(x)]));
>> plotcoeffs(Q)
Expected one output from a curly brace or dot indexing expression, but there were 3 results.

Error in chebfun/plotcoeffs (line 31)
dom = f.domain([1,end]);
```
I don't know enough about the internal workings of quasimatrices to come up with an appropriate fix.

nickhale:
You need f(1).domain, I think.

PR #1860, opened 2016-05-02
Fix chebfun2 and spherefun sample method

### LINKED ISSUE #1808: Calling chebfun constructor in `chebfun2/sample`
Currently in `Chebfun2/sample`, we do the following:
```
[C, D, R] = cdr( f ); 
Cvals = sample(chebfun(C), m);
Rvals = sample(chebfun(R), n);
```
My point is only about the parts `chebfun(C)` and `chebfun(R)` in the last two lines above. `C` and `R` are already quasimatrices. Is it needed to call the 1D chebfun constructor again on `C` and `R`? The point is that our goal here is just to sample `C` and `R`. Do we gain anything by calling the constructor one more time? This troubles me particularly when I make a cross-section of a Chebfun3 (so a Chebfun2). Why not just say the following?
```
[C, D, R] = cdr( f ); 
Cvals = sample(C, m);
Rvals = sample(R, n);
```
This change solves my issue in Chebfun3 and also none of the whole chebfun tests fail with this. Following #1807, is this related to trigfun-chebfun conversions? @ajt60gaibb,  would you (and possibly others) please give me a hint on this?

gradywright:
This PR includes the following:

First, `chebfun2/sample` now calls the `sample` function of the techs that makes up the the rows and columns instead of always sampling on a tensor product Chebyshev grid.  This dramatically improves the speed of some operations.  Here's an example where two Chebfun2 objects are constructed with the `trig` flag:
```
>> f = chebfun2(@(x,y) tanh(10*(1+cos(5*pi*(x-0.2)).*sin(7*pi*y))),'trig');
>> g = chebfun2(@(x,y) cos(20*pi*cos(pi*x).*sin(pi*y)),'trig');
```
Adding these together and displaying the result on the development branch takes the following amount of time:
```
>> tic, h = f+g, toc 
h =
   chebfun2 object
[  -1,   1] x [  -1,   1]       62     [   2    2    2    2]
vertical scale =   2 
Elapsed time is 4.836512 seconds.
```
Here are the timings on this branch:
```
>> tic, h = f+g, toc
h =
   chebfun2 object
[  -1,   1] x [  -1,   1]       62     [   2    2    2    2]
vertical scale =   2 
Elapsed time is 0.652679 seconds.
```

Second, the `spherefun/sample` method now calls the `sample` method of `trigtech` instead of breaking encapsulation.

Closes #1808.

### DIFF
--- a/@chebfun2/sample.m
+++ b/@chebfun2/sample.m
@@ -30,8 +30,8 @@ end
 
 % Use CDR decomposition so we can keep it in low rank form: 
 [C, D, R] = cdr( f ); 
-Cvals = sample(C, m);
-Rvals = sample(R, n);
+Cvals = sample(C, n);
+Rvals = sample(R, m);
 
 % Evaluate: 
 if ( nargout <= 1 )
--- a/@chebfun2/sample.m
+++ b/@chebfun2/sample.m
@@ -30,8 +30,8 @@ end
 
 % Use CDR decomposition so we can keep it in low rank form: 
 [C, D, R] = cdr( f ); 
-Cvals = sample(chebfun(C), m);
-Rvals = sample(chebfun(R), n);
+Cvals = sample(C, m);
+Rvals = sample(R, n);
 
 % Evaluate: 
 if ( nargout <= 1 )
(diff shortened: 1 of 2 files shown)


ajt60gaibb:
Thanks, @gradywright !

Issue #1441, opened 2015-04-02
QR factorization of trigfun based quasimatrix.


jaurentz:
The QR factorization of a trigfun based quasimatrix typically works as expected. One exception is the case where the quasimatrix is nearly rank one. 
```matlab
>> [Q,R] = qr(chebfun(@(x) [cos(pi*x),cos(pi*x)+1e-12],'trig')); norm(Q'*Q-eye(2))

ans =

   1.5410e-09
```
We see that a fairly small perturbation can return a Q that might be a bit far from orthogonal. The phenomenon gets worse as the perturbation gets smaller.
```matlab
>> [Q,R] = qr(chebfun(@(x) [cos(pi*x),cos(pi*x)+1e-15],'trig')); norm(Q'*Q-eye(2))

ans =

   3.4400e-03
```
Shouldn't a QR factorization routine always return a quasimatrix that is orthogonal to machine precision?

jaurentz:
In the branch feature-prescribed-eps there is a rank deficient case that is currently disabled in 
tests/trigtech/test_qr.m. Once the bug is fixed, this test should be reinstated.

jaurentz:
A similar issue has arisen both on Kuan and Behnam's laptops in `chebtech/qr.m` as illustrated by epslevel-dependent failures in `tests/chebtech/test_qr.m` in branch `feature-prescribed-eps`.

jaurentz:
Here is a similar example of the failure using the `chebtech1` class and Householder QR.
```matlab
>> f = chebtech1.make(@(x) [x x x]); 
>> [Q,R] = qr(f,[],'householder');   
>> norm(innerProduct(Q,Q)-eye(3),inf)

ans =

    0.1978
```
The `trigtech` failure above is also due to the Householder QR used in the `abstractQR.m` routine. For speed reasons, the continuous QR factorization is computed in value space. The danger with this is that if not enough points are chosen the inner product is computed inaccurately. This is exactly what happens in the rank deficient case. After the first Householder reflector is applied, the remaining columns are high degree polynomials on the order of machine precision. Since they have high degree, the value space-based inner product is incorrect, resulting in a Q that is not orthogonal.

asgeirbirkis:
In merging #1474, two tests in `chebtech/test_qr` and `trigtech/test_qr` got disabled because of this issue. Once we close this ticket, we should reinstate those tests.

PR #1862, opened 2016-05-03
Implement built-in qr for trigtech

### LINKED ISSUE #1847: trigtech qr fails for zero quasimatrix
The following example illustrates the issue:
```
>> g = trigtech(@(x)[0*cos(x) 0*sin(x)]);
>> f = qr(g);
>> f.coeffs
ans =
      NaN + 0.0000i      NaN + 0.0000i
      NaN +    NaNi      NaN +    NaNi
      NaN + 0.0000i      NaN + 0.0000i
      NaN +    NaNi      NaN +    NaNi
```
Here's what happens for the same example in `chebtech2`:
```
>> g = chebtech2(@(x)[0*cos(x) 0*sin(x)]);
>> f = qr(g);
>> f.coeffs
ans =
    0.5000    0.5000
   -0.8660    0.8660
```

gradywright:
This PR contains a `built-in` version of `qr` for `trigtechs` similar to the `built-in` version for `chebtechs`, and it makes this the default version that is used.  This new version is considerably faster than the `householder` version, which uses `abstractQR`.  This is especially nice for Spherefun objects since many of the methods in Spherefun make heavy use of `trigtech/qr`.

Here's an example of the speed-up:
```
>> f = spherefun(@(x,y,z) cos(100*pi*x));
>> g = spherefun(@(x,y,z) cos(100*pi*y));
```
Time for `f+g` on this branch:
```
>> tic, h = f+g; toc
Elapsed time is 0.591149 seconds.
```
Time for `f+g` on development:
```
>> tic, h = f+g; toc
Elapsed time is 2.056394 seconds.
```

This PR also closes #1847.

### DIFF
--- a/@trigtech/qr.m
+++ b/@trigtech/qr.m
@@ -127,7 +127,7 @@ function [f, R, E] = qr_builtin(f, outputFlag)
 % Quadratures are being done using trapezoidal rule. We must enforce that 
 % f.coeffs has at least as many rows as columns
 [nf, mf] = size(f);
-n = max(nf, mf);
+n = max(nf, mf);        % We want Q to have mf columns. 
 isReal = all(f.isReal);
 f = prolong(f, n);
 if (isReal)
--- a/@trigtech/qr.m
+++ b/@trigtech/qr.m
@@ -58,9 +58,6 @@ if ( size(f, 2) == 1 )
     return
 end
 
-% Simplify so that we don't do any extra work: (QR is O(m*n^2)? :/ )
-f = simplify(f);
-
 % Decide which algorithm to use:
 if ( strcmpi(methodFlag, 'householder') )
     % Call Trefethen's Householder implementation:
--- a/@trigtech/qr.m
+++ b/@trigtech/qr.m
@@ -130,6 +130,10 @@ function [f, R, E] = qr_builtin(f, outputFlag)
 n = max(nf, mf);
 isReal = all(f.isReal);
 f = prolong(f, n);
+if (isReal)
+    % Force values to be real if f is real (avoid complex rounding errors in R)
+    f.values = real(f.values);
+end
 
 if ( nargout == 3 )
     [Q, R, E] = qr(f.values, 0);
@@ -162,10 +166,6 @@ f.coeffs = f.vals2coeffs(Q);
 % If any columns of f were not real, we cannot guarantee that the columns
 % of Q should remain real.
 f.isReal(:) = isReal;
-% However, if f is real, we want to remove complex rounding errors in R.
-if (isReal)
-    R = real(R);
-end
 % Prune the unneeded coefficients.
 f = prolong(f, nf);
 
--- a/@trigtech/qr.m
+++ b/@trigtech/qr.m
@@ -124,12 +124,12 @@ end
 end
 
 function [f, R, E] = qr_builtin(f, outputFlag)
-% Quadratures are being done using trapezoidal rule, so we need to double
-% n to ensure that all integrals involving products p_n(z)*p_m(z), where
-% z = exp(i*x), are done exactly.  Additionally, we must enforce that 
+% Quadratures are being done using trapezoidal rule. We must enforce that 
 % f.coeffs has at least as many rows as columns
 [nf, mf] = size(f);
-n = 2*max(nf, mf);
+n = max(nf, mf);
+% Increasing n by one seems to make the results more accurate.
+n = n + 1;
 f = prolong(f, n);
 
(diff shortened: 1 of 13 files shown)


nickhale:
Since the weights are constant here, do they really need to be applied inside the QR? (i.e., can't you just sort out the scaling afterwards?)

Also it appears that you do the same thing on either side of the `if ( n <= 4000 )` branch.

How about this? https://github.com/chebfun/chebfun/tree/fix-trigtech-qr-nh

gradywright:
Thanks @nickhale, this is much cleaner than my version.  I wasn't exactly sure what to do with the weight, so I just left it in there.  I will merge your changes into this branch if that's okay with you.

The TODO in your code is handle in the last line when `prolong` is called.  We (I) actually forgot to do this in the `householder` version in my original commit of `trigtech`.

nickhale:
```n = 2*max(nf, mf);```
Is the doubling necessary here? Isn't the trapezium rule is 'Gaussian' for trigonometric polynomials?

**Comment by gradywright** [comment]
@nickhale, of course!  I guess I was mixing up degree and length in my head.  Thanks, now the code is even faster.

**Comment by nickhale** [comment]
The problem is that this fails for me if I remove that factor of two:
```
>> f = spherefun(@(x,y,z) sin(pi*x.*y));
>> norm(f+f+f-3*f)
ans =
     3.561713951650347e-05
```

I can't see why. 

This seems to fix it:
```
n = max(nf+1, mf);
```
but I also can't see why..

**Comment by nickhale** [comment]
PS. It's possible the Householder code also doesn't need to double n. 

Also, I  have a suspicion that `prolong` is destroying the orthogonality.

**Comment by nickhale** [comment]
`test_composition_operators` was failing for me. 

Part of the reason for this was that although f was real (f.values were complex after simplifying, but f.isReal is true) R had some complex rounding errors. After forcing R to be real in this case, the test passes, although someone should investigate why have such small complex rounding errors can make such a big difference?)

I now find that I don't have to increase n by one. Do you find the same @gradywright ?

**Comment by ajt60gaibb** [comment]
Is the factor of `2` necessary because the quadrature rule is not Gaussian with the `sin` weight?  I am the one that put a factor of `2` there...

**Comment by gradywright** [comment]
@ajt60gaibb Your comment concerns `spherefun/svd`.  We should not worry about this in the present PR.

**Comment by ajt60gaibb** [comment]
All the tests pass, which is all I can say about this PR.  #Nervous.

File: @trigtech/qr.m [hunk 1]
```diff
@@ -95,3 +122,52 @@
...
 
 end
+
+function [f, R, E] = qr_builtin(f, outputFlag)
+% Quadratures are being done using trapezoidal rule. We must enforce that 
+% f.coeffs has at least as many rows as columns
+[nf, mf] = size(f);
+n = max(nf, mf);
+isReal = all(f.isReal);
+f = prolong(f, n);
```
ajt60gaibb: Why do we prolong to `n = max(nf,mf)`?

gradywright:
Along a similar line of questioning, why do we do
```
% We must enforce that f.coeffs has at least as many rows as columns:
[n, m] = size(f);
if ( n < m )
    f = prolong(f, m);
    n = m;
end
```
in `chebtech/qr`?  I added `n = max(nf,mf)` do mimic this behavior.

ajt60gaibb:
Since this is just being overly cautious, shall we just merge this PR?  This is possible a very small optimization that can be done at a later date.  If you agree @gradywright, then I will merge.

nickhale:
Wait one.

gradywright:
This PR also closes #1441.  👍

nickhale:
```
>> [Q, R] = qr([1 2 3], 0)
Q =
     1
R =
     1     2     3
```
^^ This is why we do `n = max(nf, mf);`. We want Q to have mf columns.

nickhale:
Can I propose removing the `simplify` call at the start of `qr`? Most of the time is spent there now.

nickhale:
Off-topic, I know, but
```
>> test_gaussfilt
Undefined function 'gaussfilt' for input arguments of type 'spherefun'.
Error in test_gaussfilt (line 8)
g = gaussfilt(f); 
```
?

ajt60gaibb:
The `simplify` call is in both `trigtech/qr` and `chebtech/qr`.   It can be removed if you wish.

ajt60gaibb:
@gradywright and I are happy for @nickhale to merge this exhausting PR.

gradywright:
Thanks @nickhale and @ajt60gaibb for all your help on this one.

PR #1866, opened 2016-05-05
Improve spherefun svd

ajt60gaibb:
@gradywright's improvement. 

There is no need to set `n = max(2*length(A), 30)` in `spherefun/svd` for the quadrature rule to calculate inner-products. Instead, one can use `n = length(A) + 9`.  This small changes slightly improves the efficiency of `spherefun/svd`. For example, 

```
% development branch:
f = spherefun(rand(100)); 
tic, v = svd( f ); toc 
Elapsed time is 0.326853 seconds.

% improve-spherefun-svd branch:
f = spherefun(rand(100)); 
tic, v = svd( f ); toc 
Elapsed time is 0.266533 seconds.
```

### DIFF
--- a/@spherefun/svd.m
+++ b/@spherefun/svd.m
@@ -55,7 +55,9 @@ end
 
 
 function [Q, R] = sphereQR( A )
-% Fast version of the abstractQR code, specifically for the sphere. 
+% A faster version of the abstractQR code made specifically for the sphere.
+% 
+% See abstractQR.
 
 n = length(A);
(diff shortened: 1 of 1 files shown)


gradywright:
The results are a bit more impressive when the length of the columns is larger, for example using `f = spherefun(rand(400,100))`, we have
```
% development branch:
>> tic, v = svd( f ); toc
Elapsed time is 0.995572 seconds.
% improve-spherefun-svd branch:
>> tic, v = svd( f ); toc
Elapsed time is 0.368517 seconds.
```

PR #1867, opened 2016-05-05
Fix separableApprox/norm.

### LINKED ISSUE #1720: Distinction between infinity norm and maximum norm in chebfun2
We know that, as continuous analogues of matrix norms, the 1-norm and the infinity-norm of a `chebfun2` object $f$, each correspond to integration in one of the two dimensions. But, this is not the way these are implemented in `separableApprox/norm`. In particular, currently the infinity norm is mixed with the maximum (Chebyshev) norm, as saying
`
norm(f,'inf')
`
runs the same algorithm as saying
`
norm(f,'max')
`
I suggest that we do the following:
1- Correct the code for the infinity norm.
2- Add the capability of computing the 1-norm to chebfun2.

### LINKED ISSUE #1757: 2-norm and Frobenius norm of a chebfun2
I agree that the Frobenius norm of $f(x,y)$ is

```
sqrt(sum2(f.^2)) = sqrt( sum( svd( f ).^2 ) )
```

Also, I don't complain setting the Frobenius norm as defualt in `chebfun2/norm`. But,  I believe we should at-least allow a user to get the correct 2-norm if she/he explicitly inputs `norm(f,2)`, where by correct 2-norm I mean the largest singular value. Currently, chebfun2 considers `norm(f,2)` to be the same as `norm(f,'fro')`.

### LINKED ISSUE #1865: bug in computing operator norm of separableApprox objects
The code looks like the following when computing operator norm of a separableApprox object:

```
 case {'op', 'operator'}
            [C, D, R] = cdr( f ); 
            L = C * D * R; 
            s = svd( L ); 
            normF = s(1);  
```

I think that the 3rd line above should be replaced with  
`
L = C * D * R.'; 
`
or
`
L = C * D * R'; 
`

ajt60gaibb:
Closes #1865.  Closes #1757. Closes #1720.


bhashemi:
I am happy with the changes on this branch unless there is now an inconsistency in `@chebmatrix/norm` which causes the following test
```
spinscheme/test_startMultistep.m
```
to fail on this branch. Specifically, in Line 45 of `@chebmatrix/norm`, we porbably need to remove the 2-norm (it is being treated the same as the Frobenius norm) similar to what we do in the new `@separableApprox/norm` from this pull request. Any thoughts @Hadrien-Montanelli or anybody else who knows chebmatrices better than me?

Hadrien-Montanelli:
@ajt60gaibb Conflicts here too!

Hadrien-Montanelli:
@ajt60gaibb thanks!

Hadrien-Montanelli:
@ajt60gaibb The `spinscheme/test_startMultistep` test fails because it uses `chebmatrix/norm`, which by default uses the 2-norm on the blocks. Could you please change this?

ajt60gaibb:
I do not know the details of `spinscheme` or `chebmatrix`. I am not sure what change you have in mind. (How should it work if not on the 2-norm on the blocks?)

PR #1903, opened 2016-05-30
Feature chebfun2 sumdisk

klauszuoxinwang:
I have finished a first version of the sumdisk routine. This code evaluate the integral of a chebfun2 over a unit disk. The integral is approximated using a truncated 2D Chebyshev polynomial with a certain maximal degree that can resolve the function to machine precision. Mikael Slevinsky 's formula is used to accelerated the computation. This code only works for a chebfun 2 represented by a chebyshev pol7nomial

@Hadrien-Montanelli  you mentioned that there is a way to check if the chebfun 2 is represented by cheb polys or trig polys. I haven't find the information about this. Could you tell me how to do that? Also what is the difference between the 'master' branch and the 'development' branch?

@trefethen Do you think it looks ok? If it is necessary for me to declare 'sumdisk' as method of the chebfun2 class? Or I just need to put the m-file in the chebfun2 folder?


klauszuoxinwang:
@Hadrien-Montanelli  What is the difference between the 'master' branch and the 'development' branch?

klauszuoxinwang:
@trefethen  Dear Nick, the sumdisk now use a fast way to do all the computation, and it works for chebfun2 and trigfun2 defined on any rectangular domain. But there are several lines you put there that I don't understand 

% % If nrows and ncols are big enough, you can do the following
% % Of course you'll hto modify this a little in case they're not big enough:
% 
% I = pi*C(1,1) - (pi/2)*(C(1,3)+C(3,1);
% for i = 3:2:min(ncols,nrows)
%     I = I + xxx*C(i,i);
%     etc.

PR #1910, opened 2016-06-04
added 3 "peg" examples to cheb.gallery2

trefethen:
Following discussions with @heatherw3521 about Diskfun, I've modified `cheb.gallery2` to include three new examples called roundpeg (which has rank 45), squarepeg (which has rank 1 since it's aligned with the axes) and tiltedpeg (which has rank 100).  @heatherw3521, I've taken the liberty of assigning this to you -- ok I hope?


heatherw3521:
I had to change the tolerances in the singfun plus and times tests to get things to pass. Let me know if this is problematic.

PR #1913, opened 2016-06-07
Feature chebop string input

asgeirbirkis:
This is something that @trefethen and I started discussing a long time ago. The idea is to allow string input for constructing chebops (we support strings for chebfuns), similar to what is done in chebgui. Due to Matlab reserving `'` for characters/strings, which would require specifying derivatives as `' u'' '` for first order derivatives, `' u'''' '` for second order etc., @MikaelSlevinsky suggested the use of backticks (`` ` ``) for derivatives. Thus, a first order derivative is specified as `` 'u` ' ``, second order as ``` 'u`` ' ``` etc. 

If this gets merged, it'll be very easy to specify ODEs, such as the IVP u'' + sin(u) = 0, u(0) = 0, u(1) = 1, as 

```
>> N = chebop(' u`` + sin(u) ',[0 5]);
>> N.lbc = [0; 1]
N =
   Nonlinear operator:
      u |--> diff(u,2)+sin(u)
      [0,5]
   with
    left boundary condition(s):
      u  = 0
      u' = 1
```


asgeirbirkis:
Always forgot to say thanks @Hadrien-Montanelli for merging!

Issue #1924, opened 2016-06-19
chebfun3v lacks an integral2 command


bhashemi:
`Chebfun3` has an `integral2` command, but `chebfun3v` does not. We need to add one.

bhashemi:
This is addressed by 8626218.

PR #1925, opened 2016-06-19
Feature chebfun3v/integral2

oliviersete:
Adds an integral2 command to compute the flux of a chebfun3v object through a 2D-surface.
This addresses #1924 .


bhashemi:
@oliviersete would you please check if you are happy with the obvious changes I made?

oliviersete:
Thanks @bhashemi  for the changes and comments.
For some reason F(2) does not work, so I changed it to F.components{2}.
I also improved the error identifiers.

Issue #1943, opened 2016-07-06
Add divergence for chebfun3v with 2 components


oliviersete:

oliviersete:
Closed by #1945 .

PR #1945, opened 2016-07-06
Allow chebfun3v objects with 2 components in chebfun3v/divergence.

oliviersete:
Addresses #1943 .


Hadrien-Montanelli:
@oliviersete could you please add a test? (There was a bug in your code.)

oliviersete:
Thanks @Hadrien-Montanelli for spotting the bug.  I added a test now.

PR #1946, opened 2016-07-07
Feature lagpts using Riemann-Hilbert asymptotics

popsomer:
lagpts.m now uses Newton-Raphson iterations on asymptotic expansions of Laguerre polynomials for n > 148. These asymptotics are non-uniform and heuristics separate three regions. Higher-order terms are hard-coded, but a (longer) code can be provided to get an arbitrary number of terms. There are also (more accurate but longer) expressions without series expansions near z=0 or 1 for R1 and R2 in the functions asyAiry and asyBessel.

Although the asymptotics are not accurate up to machine precision, a comparison with respect to a high precision computation shows that the small nodes are a bit more accurate and the large nodes are less accurate than the ones from the Glaser-Liu-Rokhlin algorithm. This fast algorithm might suffer from an O(n) error by using the recurrence relation in the first 20 nodes. 

Using method='RHW' means one only wishes to compute nodes and weights where the latter do not underflow. The corresponding complexity is then O(sqrt(n) ), while the recurrence relation limits the GLR complexity to O(n). But even when one computes all nodes (method='RH'), one should see a speed-up with respect to GLR.

Because of http://dlmf.nist.gov/18.7#E20, this can be extended to hermpts.m. This has the advantage that the Bessel expansion coincides with the (cheaper) expansion in the bulk, and the Airy expansion will not be necessary from moderate n.


ajt60gaibb:
Wow! This is *really* impressive. Congratulations, Peter. 

For me this is faster than GLR for `n>=200` and the code is beautifully written and commented. It feels like the end of an era. We now have all the standard Gauss quadrature rules with some sort of asymptotics.  Thank you for making `lagpts` happen. Bravo. 

Here, are some small remarks.  

1) Do you think we should remove the annoying warning?  
```
Warning: Weights are < realmin from k about 1692. 
> In lagpts>alg_rh (line 339)
  In lagpts (line 103) 
```
A user may find it confusing.  Having a comment in the help text and keeping the method `RHW` for sublinear complexity is a good idea.  

2) Should we remove the code for the `GLR` method?  We did this for `legpts` and `jacpts`, but not `hermpts`.     

3) Could you easily do `lagpts` with `alpha\neq 0`? http://dlmf.nist.gov/18.3.  If so, please add. If it takes more than a few hours, then it's probably not worth it.  

4) I see you have a publication in preparation. Please send me a draft when appropriate. I would enjoy reading and commenting.

ajt60gaibb:
I have a suggestion.    

Once your publication appears in print (maybe 18 months from now), I encourage you to get in contact with the DLMF editors and see if you can improve `18.15(iv)` http://dlmf.nist.gov/18.15#iv.  I have found the DLMF editors to be quite receptive to improvements.

LourencoPeixoto:
Congratulations, Peter. In fact this work is very exciting!

Thinking about @ajt60gaibb's remark 4, maybe these initial guesses (Tricomi, 1947) could be fine:
`x_{n,k} = j_{\alpha,k}^2/(4*C)*(1 + (2*(\alpha^2-1) + j_{\alpha,k}^2) /(48*C^2) ) + O(n^{-5})`

where `j_{\alpha,k}` is the `k`th positive zero of the Bessel function and `C=n+(\alpha+1)/2`.

popsomer:
Thank you for your interest: the commit "Rem warn GLR, gen symb alpha, gen alpha qm m with conv and init guesses" contains the following changes

1) Warning removed

2) GLR temporarily removed to compensate the increase in code length a bit

3) I added two ways to get nonzero alpha: calling lagpts(n,[0,inf],'RH',alpha) will use explicit formulas for the higher order terms, while lagpts(n,[0,inf],'RH',[alpha, 1, 1]) will compute convolutions for them. The latter requires about 4 times less code and allows computing an arbitrary number of terms. It also enables a more general weight x^(aqm(1))*exp(aqm(2)*x^aqm(3) ), although the current version skips some zeros, probably because I use finite differences. But one can also argue that some expressions in the former can be removed, so I wonder which approach you prefer?

4) I added the more accurate initial guess suggested by Lourenco, and method='RHi' or 'RHWi' uses it throughout the interval and another initial guess in terms of Airy zeros from the references in the code. Both are not very accurate in the interior of the interval, leading to convergence to another zero. It could be interesting to add more terms in the initial guesses or other types of them, but the time needed to compute lagpts(1e3,'RHWi') does not seem to be lower than lagpts(1e3,'RHW'), so I think the reduction in number of iterations does not fully compensate the computation of the initial guesses.

ajt60gaibb:
1) Thanks. 
2) Thanks. 
3) That's great. I am surprised you managed to get explicit expressions in terms of `alpha`. Nice. It would be nice to keep the syntax as simple as possible: Should `lagpts(n,[0,inf],'RH',alpha)` be equivalent to `lagpts(n, alpha)` for large `n`. The final code should not skip zeros for small `alpha` (`alpha<5`, say), get close to 15-digits, and be as fast as possible.  I probably prefer the former for Chebfun.  For the Julia package, where we hope to get everything working with `BigFloat`s I hope we can keep both versions as the higher order terms will be required.  
4) If that is the case, then we should stick with the simple initial guesses. We are not allowed to miss any nodes.

popsomer:
a) No zeros are skipped for low alpha and I added a warning concerning alpha. When alpha^2/n is too large, the matrix R(z) in the last stage of the Riemann--Hilbert problem is not close to the identity, so the analysis breaks down and many zeros are skipped. However, increasing n should counter that so that there only is a problem when there are nodes above the MRS number 4*n.

b) More terms were added for alpha=0 and the bounds multiplying eps were lowered, giving about 14 digits in most cases. But I don't know whether it will be possible to get 15 digits for all nodes and weights, for all n.

c) I combined the initial guesses in terms of Bessel and Airy zeros with extrapolation by a sixth (instead of second) degree polynomial, using heuristics to choose the most accurate one. This does not miss any nodes and is so accurate that we only need 1 or 2 Newton iterations.

d) Although the number of iterations is already quite low, maybe it would be interesting to add the (unpublished) extension of Newton's method presented by Javier Segura at SIAM AN16 here or at other chebfun routines? I remember there was some simple modification in the denominator to improve the quadratic convergence of the standard Newton's method to cubic convergence.

e) Explicit formulas in terms of alpha are indeed faster; due to their length, I did not follow chebfun's guideline of a maximum of 80 characters per line everywhere. I hope the documentation of the code is OK. Is the syntax now more simple?

ajt60gaibb:
a) We all expect problems for `alpha` sufficiently large. I had forgotten about the Mhaskar–Rakhmanov–Saff (MRS) number -- great comment.

b) OK. We can live with `14`. 👍 

c) Sounds good.  What exactly are you extrapolating?  Are you using the second-order differential equation satisfied by the Laguerre polynomial? 

d) Yes, this could be a marginal improvement to our algorithms. I suspect for the range `1000<=n<=5000`.  

e) I am taking a look now...

ajt60gaibb:
This looks great to me so I am going to merge now.   I have some minor help text typos that I will fix immediately afterwards (I do not have access to update your repo).   

Thanks, @popsomer.  We have conquered all the standard Gauss quadrature rules!

popsomer:
b) In the attempt to get more digits, I did lower the constants multiplying eps and the maximum number of iterations, which does give a higher risk to get the 'CHEBFUN:lagpts:converge' error. So, a user getting this error might want to increase the eps*40, 50*eps and 5 (>=/== l)

c) No, I draw a (Lagrange) interpolating polynomial through x(k-1:-1:k-7) and evaluate at k: this gives the cheap code at the start of the loop.

I 'll do the changes to the Julia code and try to make you a nice draft...

PR #1947, opened 2016-07-18
fix-pdeSolve-parsing

nickhale:
Fix parsing of multiple options to pdeSolve.

### DIFF
--- a/@chebfun/pde15s.m
+++ b/@chebfun/pde15s.m
@@ -7,8 +7,8 @@ function varargout = pde15s(varargin)
 %   PDEFUN(UU, t, x) with the initial condition U0 and boundary conditions BC
 %   over the time interval T.
 %
-%   [TT, UU] = PDE15s(PDEFUN, T, U0, BC, OPTS, N) returns also the time chunks
-%   TT. If T is a two-vector these will be given by linspace(T(1), T(2), 51).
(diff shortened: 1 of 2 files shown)


asgeirbirkis:
Is my recollection right that the only way now to pass options to `pdeSolve` is to use the struct returned from `pde15s`, in which case, this is fixing the parsing to reflect the help text of [pde15s](https://github.com/chebfun/chebfun/blob/decb718cacd4f6fb0ca4417d2b0a94a78b399305/%40chebfun/pde15s.m) ?

Actually, looking at that help text, [L10](https://github.com/chebfun/chebfun/blob/decb718cacd4f6fb0ca4417d2b0a94a78b399305/%40chebfun/pde15s.m#L10) includes passing `OPTS` to the method, while it's not really discussed until [L70](https://github.com/chebfun/chebfun/blob/decb718cacd4f6fb0ca4417d2b0a94a78b399305/%40chebfun/pde15s.m#L70). Do you want to make the help text consistent as a part of this PR, I can then merge it?

nickhale:
I think you're right and that forcing options to be passed in the `pdeopts` struct was better.

I think I've adjusted the help documentation as you suggested. 

This pull request is now just a very small doc fix.

PR #1948, opened 2016-07-20
Fix chebcoeffs helptext

trefethen:
This pull request mainly adjusts help text in `@chebfun/plotcoeffs` in response to questions raised by @klauszuoxinwang.  The old and new help texts are reproduced below.  In addition, a number of spurious appearances of `FOURCOEFFS` in help texts have been changed to `TRIGCOEFFFS`.

Old help text:
```
 chebcoeffs   Chebyshev polynomial coefficients of a CHEBFUN.
    A = chebcoeffs(F, N) returns the first N Chebyshev coefficients of F is
    the column vector A such that F = A(1) T_0(x) +  A(2) T_1(x) + ... + 
    A(N) T_(N-1)(x), where T_M(x) denotes the M-th Chebyshev polynomial of the
    first kind.
 
    If F has a 'finite' Chebyshev expansion (i.e., it is a smooth CHEBFUN with
    no breakpoints or endpoint singularities and is based on a CHEBTECH), then
    chebcoeffs(F) is equivalent to chebcoeffs(F, LENGTH(F)). This syntax should
    be used with caution, and passing N is preferred.
 
    If F is array-valued with M columns, then A is an NxM matrix.
 
    C = chebcoeffs(F, N, 'kind', 2) returns the vector of coefficients of F
    such that F = C(1) + C(2) U_1(x) + ... + C(N) U_(N-1)(x), where U_M(x)
    denotes the M-th Chebyshev polynomial of the second kind.
 
  See also legcoeffs, FOURCOEFFS.
```
New help text:
```
 chebcoeffs   Chebyshev coefficients of a CHEBFUN.
    A = chebcoeffs(F) returns the Chebyshev coefficients of F assuming 
    it is a global (not piecewise) chebfun.  This is column vector of
    coefficients such that  F = A(1) T_0(x) + ... + A(N) T_(N-1)(x),
    where N is the length of F.
 
    If F is an array-valued chebfun, then A is a matrix with the
    same number of columns as F.
 
    If the domain of F is [a,b] rather than [-1,1], then the 
    coefficients are those of F transplanted to [-1,1].
 
    If F is a piecewise chebfun, you can extract the Chebyshev
    coefficients of the pieces with, e.g., A = chebcoeffs(F.FUNS{1}).
 
    Alternatively, A = chebcoeffs(F, N) returns the first N Chebyshev
    coefficients of a piecewise chebfun F even though F is not
    represented by a global Chebyshev expansion.  Chebfun does this
    by evaluating appropriate integrals. 
 
    A = chebcoeffs(F, 'kind', 2) or A = chebcoeffs(F, N, 'kind', 2)
    return vectors or matrices corresponding to expansions 
    F = A(1) U_0(x) + ... + A(N) U_(N-1)(x) in Chebyshev polynomials
    of the second kind.
 
    Examples:
     x = chebfun('x');
     chebcoeffs(exp(x))
     chebcoeffs(abs(x),10)
 
  See also legcoeffs, trigcoeffs.
```

### DIFF
--- a/@chebfun/chebcoeffs.m
+++ b/@chebfun/chebcoeffs.m
@@ -1,5 +1,8 @@
 function out = chebcoeffs(f, varargin)
 %CHEBCOEFFS   Chebyshev polynomial coefficients of a CHEBFUN.
+%   A = CHEBCOEFFS(F) returns the Chebyshev coefficients of F assuming
+%   it is a global (not piecewise) chebfun.
+
 %   A = CHEBCOEFFS(F, N) returns the first N Chebyshev coefficients of F is
 %   the column vector A such that F = A(1) T_0(x) +  A(2) T_1(x) + ... + 
 %   A(N) T_(N-1)(x), where T_M(x) denotes the M-th Chebyshev polynomial of the
--- a/@chebfun/chebcoeffs.m
+++ b/@chebfun/chebcoeffs.m
@@ -1,25 +1,35 @@
 function out = chebcoeffs(f, varargin)
-%CHEBCOEFFS   Chebyshev polynomial coefficients of a CHEBFUN.
-%   A = CHEBCOEFFS(F) returns the Chebyshev coefficients of F assuming
-%   it is a global (not piecewise) chebfun.
-
-%   A = CHEBCOEFFS(F, N) returns the first N Chebyshev coefficients of F is
-%   the column vector A such that F = A(1) T_0(x) +  A(2) T_1(x) + ... + 
-%   A(N) T_(N-1)(x), where T_M(x) denotes the M-th Chebyshev polynomial of the
-%   first kind.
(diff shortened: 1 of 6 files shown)


aaustin141:
The only complaint I have about this is that recommends accessing `F.FUNS{1}`.  That breaks encapsulation:  clients of `chebfun` should not be concerned with the fact that the funs are stored in a cell array.  Suggesting `get(f, 'coeffs')` would be better.  Perhaps even better would be to introduce a new syntax for `chebcoeffs' to handle this use case.

trefethen:
Thanks @aaustin141.  I've modified that bit of the text as you suggested:
```
If F is a piecewise chebfun, you can extract the Chebyshev
coefficients of the pieces with GET(F, 'COEFFS').
```
If you're happy now, please go ahead and merge.

PR #1951, opened 2016-08-01
Feature diskfun

heatherw3521:


ajt60gaibb:
👍 (There's no like button on GitHub.)

trefethen:
Very exciting!  Bravo!  The first thing I tried made me feel good:
```
>> sum2(diskfun(@(x,y) cos(13*x+exp(y))))
ans =  -0.012683198081825

>> cheb.xy, sumdisk(cos(13*x+exp(y)))
ans =   -0.012683198081825
```
However, I notice various sniggles when trying `help` this and that, so some editing is needed.  Here are some examples:

`help diskfun`: (1) must make it clear that you specify `'polar'` for the polar interpretation (or is it `'cart'` for the cartesian one?).  (2) Is it really true that a matrix of numbers is interpreted as polar rather than cartesian?  (3) The paper cited has a typo in its name.

`help sum2`: Some comment lines about splitting into even and odd parts are appearing erroneously in the help text.

`help vscale`: "vertial" should be "vertical"

`help rank`: (1) TOL/N should be TOL*N.  (2) "the approximant" should be deleted.

`help coeffs2`: (1) "modes" is the wrong word.  (2) "bases" should be "basis".

`help simplify`: "compressed" should be "compresses"

`help compose`: the texts for `compose(F,OP)` and `compose(F,OP,G)` are the same.

`help grad`: "numericalgradient" needs a space

`help svd`: (1) should repeat twice "returns the singular values of F".  (2) "quasi-matrix" should be "quasimatrix"

`help minandmax2`: (1) "value" should be "values", (2) "position" should be "positions"

gradywright:
The help text for `flipud` and `fliplr` needs to be updated and made specific to the disk.  Also,  a test should be added to make sure these methods and `flipdim` are working correctly.

gradywright:
The help text of `@diskfun/harmonic` method refers to `@spherefun/harmonic`, but this method does not exist.  In Spherefun we have the method `@spherefun/sphharm` for generating spherical harmonics.  I'm not opposed to creating a `@spherefun/harmonic` method that calls `@spherefun/sphharm`, but until that is done we should change the reference.

ajt60gaibb:
As a general rule, I would prefer all the outputs of the Diskfun commands to be independent of the underlying approximation technique that is employed. I could only find a handful of culprits: 

`trace`: Could this be an integral over a radial slice? At the moment it is some unweighted integral over the diagonal in `(th,r)`, which does not mean very much to the disk. Also, I think `trace(f) = sum(diag(f))`... currently it does not. 

`min` and `max`: Should these commands be returning zero curves in `(x,y)`, not `(th,r)`? 

`transpose` and `ctranspose`: These current reverse the roles of `r` and `th`, not reflect along the line `y=x`.  

I also found a bug in `diag`: 
```
f = diskfun(@(x,y) 1+0*x);
>> diag(f)
Conversion to logical from cell is not
possible.
Error in diskfun/diag (line 23)
if varargin
```

trefethen:
Dear everybody,

@heatherw3521 was present by skype in our Chebfun meeting at Oxford this afternoon and showed the very nice current state of Diskfun.  We propose to ask for code review to begin shortly.

The main thing to be resolved first is (a) getting the policy right about polar vs. cartesian coordinates and (b) making sure the help text and Guide chapter match this policy.

*Default coordinate system*.  One issue is: should the default coordinates be polar or cartesian?  The consensus is cartesian.  In favor of polar would be the fact that it's the natural coordinate system for the disk.  However, there are more arguments in favor of cartesian.  One is that because of the singularity at r=0, the variables r and theta, for example, can't be represented by diskfuns.  Another is that again because of that singularity, derivatives with respect to r and theta will constantly get you into trouble.  A third is a matter of consistency with Chebfun2.  

*Polar vs. Cartesian modes: stored as a preference or stored in each diskfun?*  If cartesian is the default, we still want users to be able to work with polar.  Currently there's a field `f.coords` attached to each diskfun which says that it's either to be interpreted as polar or as cartesian.  This makes me uneasy; I think polar vs. cartesian interpretations are a matter of user interface and shouldn't be attached to diskfuns.  Thus I would favor having cartesian interpretations by default, with this overrideable by a preference; and there would be an easy user interface to that preference, like `diskcoords polar` (or just `coords polar` or even `polar`? or something else?).  Whether the rest of the group shares my view on this was not clear at the meeting.  If others would express any opinions that would be great -- @ajt60gaibb and @gradywright, for example.

Where to go from here: @heatherw3521 has agreed to think about things, listen to input, and come up with a plan.

gradywright:
Getting polar vs. Cartesian policy correct is key.  My preference is to make Cartesian the default for exactly the reasons mentioned by @trefethen.  Besides construction and evaluation, what methods would be nice (or even necessary) for a user to want to also work in polar coordinates?  If these are it, then I think we should just have special ways to construct and evaluate (with a flag for former and a method for latter) and have no `f.coords` parameter or global default.

ajt60gaibb:
Cartesian should be default because most operations do not preserve smoothness in polar coordinates. We probably only have construction and evaluation that are necessary for a user in polar coordinates, I can not think of any more? (I hope `trace` and `transpose` move to Cartesian as they are strange operations in polar.) Therefore, I vote for keeping the same syntax that is currently in Diskfun: 
```
f = diskfun(@(th,r) r.*cos(th),'polar');
```
and 
```
f( pi/3, .1, 'polar')
```
If we do find a bucket load of operations that are useful in polar coordinates, then we can add a global flag later. It's probably too much for two commands.

heatherw3521:
Thanks for all the input both here and at the meeting about how Diskfun should handle the "coords" dilemma.  I am agreeing now that we should anticipate and encourage the use of Cartesian coordinates, but leave the polar option available only in construction and evaluation through a flag. In discussion, @ajt60gaibb suggested we include `g = cart2pol(f)`, where when `f` is a diskfun, `g`  is returned as a chebfun2 on [-pi, pi]x[0, 1] that represents the diskfun in polar coordinates.  I like this idea. A consequence of this format is that you lose the automatic application of the measure associated with the disk.  So `sum2(g)` is not equal to `sum2(f)`.  However, this might be exactly what one expects from such a transformation; it might be off-putting if the measure associated with the disk magically applied itself in the rectangular setting.   

Unless there are objections, I'll eliminate the 'coords' parameter associated with individual diskfuns and add in the `cart2pol` function.

gradywright:
I like the `cart2pol` function.  Someone will definitely then try `pol2cart`.  What should the behavior of this be?

One other place polar coordinates shows up is in the CDR decomposition.  This will not be possible to express in Cartesian coordinates.

gradywright:
@heatherw3521 It looks like `rotate` and `circshift` do the same thing.  If this is true, the help texts should be the same (or really similar) and one of these methods should just call the other.

ajt60gaibb:
I looked through `diskfunv` code and tidied up the help text and inline comments. It looks good to me.  I am happy with it now. 

It is a little weak on extensive testing.  @heatherw3521, can you add tests that execute the commands:   `diskfunv/diffx`,`diskfunv/diffy`, `diskfunv/get`, `diskfunv/divgrad`,`diskfunv/jacobian`, `diskfunv/set`.

ajt60gaibb:
I am happy with the `diskfunv` part of the code now.

ajt60gaibb:
@gradywright, does your recent commit to `diskfun/circshift` to make it faster break encapsulation?

gradywright:
@heatherw3521:  I'm not sure what `isCartesian` is suppose to do.  Can you update the help text, or just remove this method entirely.  It appears to be used in `feval`, but I don't understand how it ever ends up giving a non true result.

bhashemi:
@heatherw3521 thanks for your great job on the diskfun project. Here are a few minor comments I currently have about the diskfun code:

1) In `diskfun/min2`, `diskfun/max2`, and `diskfun/minandmax2`, how about replacing 
>> [C, D, R] = cdr(f); 
>> f = C*D*R.'; 

with
``
f = cart2pol(f, 'cdr');
``

2) The following give unexpected error messages:
>> f = diskfun();
>> a = minandmax2(f)
One or more output arguments not assigned during call to "varargout". 
>> [a,b] = minandmax2(f)
One or more output arguments not assigned during call to "varargout". 

Please compare with the error message we get for
>> f = chebfun2();

3) How about replacing the present error message in `diskfun/mean` with the following:
``
error('CHEBFUN:DISKFUN:mean:dim', 'dim must be 1 or 2.')
``
I suggest the error message in `diskfun/sum` to be also changed accordingly.

4) Please replace `SEPARABLEAPPROX` with `DISKFUN` in a few lines of `diskfun/sum`.

I'll continue reviewing the code.

heatherw3521:
Thanks @bhashemi . I'll make the changes you suggest.

bhashemi:
Here are my latest review comments on `diskfun`. What's left for me to review is just codes in `tests/diskfun` which I'll look at soon. My older comments in the above are numbered from 1 to 4. So, here I start the first label from 5 to avoid further confusion.

5) Please add a test for diskfun/median. It does not currently work as mentioned in #1969.

6) Diskfun currently lacks the following methods which appear in the list when one invokes 
``methods diskfun``:
```
quad2d.m
quiver.m
quiver3.m
restrict.m
tan.m
```
This causes help queries like ``help diskfun/quiver`` to display `separableApprox`.

7) I am a bit confused about how the common roots of TWO `diskfun` objects are to be computed. Is this supposed to work at the moment? Would you please comment on this?

8) Since `diskfun/sampleTest` is just called **internally** by the constructor and won’t be called by _users_, it might be better to make it _private_ so that it does not appear in the list ``methods diskfun``. Same comment applies to `singleSingTest`. This is done e.g. in the `@chebfun3` folder.

9) The help-text of `diskfun/surface` probably needs more details.

10) My guess is that a command similar to 
``f = cart2pol(f, 'cdr');``
is needed in `diskfun/waterfall` before it calls `separableApprox/waterfall` as it needs evaluation at two input arguments.

gradywright:
I have finished my review of the `diskfun` code, which consisted of reviewing the files with names starting from a to l.  I made a few changes here and there, and am now happy.  I also reviewed the tests for these same files and made a few improvements.  My only request @heatherw3521 is to add some tests for `integral` and `integral2`.

Great work!

bhashemi:
Here are my final `diskfun` review comments:

11) We probably need test codes at least for `min`, `max`, `mean` and `rotate`.

12) I have made a few changes I thought are necessary to the help-text of `diskfun/projectOntoBMCII.m`. @heatherw3521 , would you please confirm that these are fine?

13) The second check, `pass(2)`, in `test_svd` passes but is not far from failure. How about removing it or replacing it with a better test?

heatherw3521:
Thanks @ajt60gaibb , @bhashemi and @gradywright for all the work! I have a few more small things to do related to your comments and will let you know when those are all completed. @bhashemi , I think your commentary on `projectontoBMCII` looks good; I left it as is. Also, I am thinking about your question (7) about how to handle finding the common roots of two diskfuns. This is also an issue we should think about in Spherefun.

ajt60gaibb:
I was expect `diskfunfv/roots` to work like this: 

```
% diskfuns:
f = diskfun(@(x,y) cos(20*(y-.1)));
g = diskfun(@(x,y) cos(10*x.*y));

% Convert to chebfun2 objects:
ff = chebfun2(@(th,r) feval(f,th,r,'polar'),[-pi pi 0 1]);
gg = chebfun2(@(th,r) feval(g,th,r,'polar'),[-pi pi 0 1]);

% Find common roots: 
rts = roots(ff,gg);

% Convert back to cartesian coordinates:
[x,y] = pol2cart(rts(:,1),rts(:,2)); 

% Plot: 
contour(g), hold on,
plot(x,y,'.','markersize',30)
```

heatherw3521:
@bhashemi, @gradywright, @ajt60gaibb: I have addressed all the concerns listed and tested the changes made.  What's the next step?

bhashemi:
Thanks @heatherw3521 for this exciting news! I am happy to run all the tests and merge into development if @gradywright and @ajt60gaibb agree, too.

ajt60gaibb:
I agree.

bhashemi:
@heatherw3521, I ran all the test both on my mac and desktop machine. The following tests fail on _both_ machines, none of which fail on the development branch:
```
The following tests failed or crashed:
   diskfun/test_optimization.m
   diskfun/test_sum.m
   misc/test_lagpts.m
   spherefun/test_optimization.m
```
`diskfun/test_optimization.m` fails because the second test gives `-0.999999999997896` for global minimum which is farther than `tol=1e-13` to the exact value `-1`.

`diskfun/test_sum.m` fails because the third test fails. Please take a closer look here as e.g., the function defined in the code `diskfun/sum.m` is called `sum3`, and Line 20 also says `% Default to y direction: `.

`misc/test_lagpts.m` and `spherefun/test_optimization.m` also fail. The first one e.g., has two different variables called `pass(11)`.

heatherw3521:
@bhashemi . Thanks for running the tests. There was an error in `test_sum` which is fixed now. I ran the optimization tests in both spherefun and diskfun and they pass on my machine. Perhaps this is related to the optimization toolbox? I don't have it installed.  @ajt60gaibb , should I adjust the tolerance in the test or is this something to worry about? The code for `misc/test_lagpts` has the `pass(11)` error in this branch and also in the development branch.  For some reason it passes in development but fails in feature-diskfun. I checked through the files; it has not been altered on this branch.

ajt60gaibb:
All the chebtests pass for me now.

PR #1960, opened 2016-08-10
Fix chebfun minandmax

### LINKED ISSUE #1959: Min and max can be fooled by .pointValues
The following code determines that the max of this chebfun `f` is `1` at `x = -1`. However, when we evaluate `f(-1)` we get 0 because of the `.pointValues`.

```
>> f = chebfun(@(x) x.^2);
>> f([-1, 1]) = 0;
>> [y, x] = max(f)
y =
     1
x =
    -1
>> f(x)
ans =
     0
```

nickhale:
Closes #1959.


nickhale:
It turns out this causes a lot of the other tests to fail (mostly it seems for tolerance reasons.) 

I don't have time to investigate further, so I'm closing the pull request.

PR #1961, opened 2016-08-10
Fix lagpts for general alpha

popsomer:
The sum of the quadrature weights is now gamma(alpha+1). Some higher-order terms were corrected for general alpha and two were added to get better accuracy in the range 128<=n<900, as mentioned to @ajt60gaibb . This should not change the barycentric weights, nor the results of chebtest.


ajt60gaibb:
Great. Thanks, @popsomer.

Looking forward to seeing a final draft of the related paper.  I would also be interested to hear what you think about Jim Bremer's new contribution: 
https://www.math.ucdavis.edu/~bremer/papers/zeros.pdf

popsomer:
The final draft might take even more time...

I think Jim's contribution has a very efficient way to compute Gauss quadrature rules. In fact, gausslaguerre.jl, when changed to abs(step) > eps(Float64)*sqrt(n)*40*x[k] to avoid the error, seems to be about 20 times slower than his Table 6. The number of allocations increase with n in Julia however, so to allow a fair comparison, the code would have to be converted to Fortran (which does not seem to have an intrinsic for Bessel or Airy functions). 

It might be expected that only an explicit asymptotic expansion of the nodes and weights can outperform his procedure. That expansion would also avoid evaluating the cosine (or other functions) at large arguments and does not need to compute the zeros consecutively.

Issue #1956, opened 2016-08-03
Evaluation of chebfun2 for single real arguments


guettel:
Evaluation of a chebfun2 does not work as expected when the single argument is real:

f = chebfun2(@(z) sin(z)); 
f(1i) % this works
f(1) % this doesn't

PR #1962, opened 2016-08-15
Compose chebfunN objects

### LINKED ISSUE #1922: Composition of chebfun2 and chebfun3 objects
We can compose chebfuns:

```
f = chebfun(@(t) exp(t)); g = chebfun(@(t) -t);
h = f(g)
```

Composition in Chebfun2 and Chebfun3 does not work, e.g.

```
f = chebfun3(@(x,y,z) x+y+z);
g = chebfun3v(@(x,y,z) x, @(x,y,z) y, @(x,y,z) z)
h = f(g)
Error using chebfun3/subsref (line 26)
Can only evaluate at triples (X,Y,Z) 
```

The following might be useful in Chebfun3:
- composition: chebfun3 or chebfun3v of a chebfun3v with 3 components
- restriction to surfaces: chebfun3 or chebfun3v of a chebfun2v with 3 components
- restriction to curves: chebfun3 or chebfun3v of a chebfun quasimatrix ("chebfun1v")

and in Chebfun2:
- composition: chebfun2 or chebfun2v of a chebfun2v with 2 components
- restriction to curves: chebfun2 or chebfun2v of a chebfun quasimatrix ("chebfun1v")

oliviersete:
This pull request enables compositions `g(f)` between Chebfun, Chebfun2, Chebfun3, Chebfun2v, and Chebfun3v objects, and closes #1922 .

If g is a chebfun3 or chebfun3v, it can be composed with a chebfun3v, chebfun2v, or a chebfun with 3 columns.  Also `g(x,y,z)` is enabled, where x,y,z are either Chebfun3, Chebfun2, or Chebfun objects.
Similar if g is a Chebfun2 or Chebfun2v object.

If g is a Chebfun2 or Chebfun2v and f is a Chebfun, Chebfun2 or Chebfun3, `g(f)` is interpreted as `g(real(f), imag(f))`, regardless whether f is real or not.  (This is the same as `g(z)` is now computed; see #1956 and #1957 .)

If g is a Chebfun (with one, two or three columns), this also adds `g(f)` where f is a Chebfun2 or Chebfun3 object.


bhashemi:
@oliviersete, I've started looking at this branch. Here is a first issue: 
```
>> f = chebfun2(@(x,y) cos(pi*(x+y)));
>> g = chebfun(@(t) t);
>> h = g(f)
Error using chebfun2/compose (line 35)
OP(F) is not defined, since image(F) is not contained in domain(OP).
Error in chebfun/subsref (line 117)
            out = compose(x, f);  
```
The reason we get this error is that floating point comparison is tricky and [this line](https://github.com/chebfun/chebfun/blob/57beda5cdd0f7b96cb84d356f765756dc7a5fe65/isSubset.m#L20) probably needs a little revision. Instead of 
```
if ( ( A(1) < B(1) ) || ( A(2) > B(2) ) )
       out = 0;
       ...
```
I suggest to use something like the following:
```
tol = 10 * chebfuneps;
if ( (B(1) - A(1))/max(abs(A(1)), abs(B(1))) > tol || ...
        (A(2) - B(2))/max(abs(A(2)), abs(B(2))) > tol )
       out = 0;
       ...
```
See e.g., [this page](https://books.google.co.uk/books?id=Zu-HAwAAQBAJ&printsec=frontcover&dq=D._Knuth_The_Art_of_Computer_Programming_%5BV._2_-_Seminum._Algs%5D&hl=en&sa=X&ved=0ahUKEwjm18ersuTOAhUZM8AKHfifAFcQ6AEILDAD#v=onepage&q=strict%20equality%20of%20floating%20point%20values%20is%20of%20these%20definitions%20apply%20to%20unnormalized%20values%20as%20well%20as&f=false) in Section 4.2.2 of Knuth's  The Art of Computer Programming.

bhashemi:
As @oliviersete has mentioned in #1963, if f is a 1D `trigtech`, then currently in Chebfun g(f) is not necessary periodic. At the moment, in the development, a composition like g(f) does NOT work properly if f is a `chebfun2` or a `chebfun3` and it is the aim of this branch to add this capability. I am concerned about the following inconsistency that will then arrise: 

Once this branch is merged, g(f) will be periodic if f is a _periodic_ chebfun2 or chebfun3 but not if f is a _periodic_ 1D chebfun. If I remember correctly there was some reasons for g(f) not to be periodic in the 1D case but I don't remember the details. Any ideas please, specially by @gradywright and @aaustin141.

oliviersete:
@bhashemi Thanks for your comments.  I addressed the first issue by replacing the line in question by
```
if ( ( a(1) < b(1) - tol ) || ( a(2) > b(2) + tol ) )
   out = 0;
   ...
```
where roughly `tol = 100 * eps * max(vscale(f), vscale(g)) * norm(domain(f), inf)`, modeled after the check in `chebfun/compose`.

bhashemi:
@oliviersete , is it possible to update Lines 107-109 of `spherefun/compose` on this branch so that it first creates and stores `real(f)` and `imag(f)` and then use them? It seems that at the moment those lines are calling the `spherefun` constructor a few times more than necessary.

oliviersete:
@bhashemi Thanks for spotting this.  It is fixed now.

bhashemi:
@oliviersete , thanks for the last fix. Here are more comments. In `minandmax3est`, I think we need to use the `sample` method instead of manually sampling the function there. For example, using `vals = sample(f, N, N, N);`.
If the object is of `trig`-type, then it is more sensible to sample at Fourier points and the `sample` method does that automatically. 

The same comment probably applies to `separableApprox/minandmax2est.m`, but it seems that there is no `separableApprox/sample.m` method to be used. Moreover, there is no `chebfun2/minandmax2est`, but there is a `diskfun/minandmax2est`.

oliviersete:
@bhashemi Thank you for your comment.  `chebfun3/minandmax3est()` now uses the `sample()` command, as you suggest.

Actually, there is a `chebfun2/minandmax2est()` (and also for diskfun and spherefun), which all redirect to `separableApprox/minandmax2est()`.

Similarly to the 3d case, `separableApprox/minandmax2est()` now calls `sample()` to get values on a grid.  Depending on what object `f` is, the appropriate `sample(f)` is used (chebfun2, diskfun, or spherefun).

bhashemi:
@oliviersete , thanks for your last changes. In quite a few `subsref` codes modified in this branch, the check for the case that the inputs are of **numeric** type is postponed to the very last step (i.e., the `else` line) of the `if-else` commands. I think the case when inputs are `numeric` is much more important than the case that these are chebfun-type objects and therefore this check should be done right in the beginning of  `if-else` commands. See e.g., 
`chebfun3/subsref.m` and `separableApprox/subsref.m`.

oliviersete:
@bhashemi The input in subsref() is now tested first for numeric or `':'` inputs.

PR #1973, opened 2016-08-28
Fix remez for odd functions

nakatsukasayuji:
This is to fix the nightly failure regarding (C) ATAP/chap24.m. 
The 'fix' on Aug 26 was incorrect for remez(odd functions); it should always be (odd,even) to avoid poles at 0.

### DIFF
--- a/@chebfun/remez.m
+++ b/@chebfun/remez.m
@@ -238,7 +238,7 @@ c = chebcoeffs(f, length(f));
 c(end) = 2*c(end);
 
 % Check for symmetries and reduce degrees accordingly.
-if ( max(abs(c(2:2:end)))/vscale(f) < eps )   % f is even.        
+if ( max(abs(c(2:2:end)))/vscale(f) < eps )   % f is even.
     if ( mod(m, 2) == 1 )
         m = m - 1;
     end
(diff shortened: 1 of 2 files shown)


bhashemi:
@nakatsukasayuji, the case that f is even is looking good and I understand your comment that for remez(odd functions) we should go for (odd, even) degrees to avoid poles at 0. However, I am not completely sure about the suggested implementation for the odd case. Here is an experiment:
```
>> cheb.x; f = x.^3;
>> [p,q] = remez(f,0,2)
Error using chebfun/remez>computeTrialFunctionRational (line 330)
Trial interpolant too far from optimal
Error in chebfun/remez (line 103)
        [p, q, h] = computeTrialFunctionRational(fk, xk, w, m, n, N, dom); 
```
In other words, if f is odd and the user requests degree 0 for the numerator, then we are still in trouble and your older fix of a few days ago was looking more appropriate. Specifically, my guess is that we need something like the following to cover all the four cases possible for the degrees `m` and `n`:
```
elseif ( max(abs(c(1:2:end)))/vscale(f) < eps ) % f is odd. obtain (odd,even) type
     if ( ~mod(m,2) )
         m = max(0, m-1);
     end
     if ( mod(n,2) )
         n = n - 1;
     end
```
This resolves the special experiment `[p,q] = remez(f,0,2)` from above. I haven't looked at the CF code to see if and how this is taken care of there. Am I missing some details here?

nakatsukasayuji:
I agree completely, we don't want to set m=m-1 when m=0! Thanks for spotting this.

PR #1982, opened 2016-09-20
Fix spin multistep

### LINKED ISSUE #2019: bug in `spin` plot titles
If you do a command like `spin('kdv')`, a plot is produced with a title that doesn't go away in response to commands like `clf` or `title(' ')`.  @Hadrien-Montanelli, is this something you can sort out?

Hadrien-Montanelli:
This branch implements a new strategy for starting multistep schemes (suggested by the reviewers of the `spin` paper). To start multistep schemes, one needs `q` values: the initial condition and `q-1` extra values. The new strategy is as follows:
1. Get an approximation of the `q-1` values with a computationally cheap one-step exponential integrator (ETDRK2).
2. Use a fixed point algorithm to refine this approximation.

Another major change: I got rid of the automatic choices of time-steps and grid points.

Minor changes: `spincomp` now measures the computing time of the time-stepping loop only, some polishing in `solvepde`, and closes #2019.


Hadrien-Montanelli:
Any update @nakatsukasayuji?

PR #1991, opened 2016-10-04
Fix bounds checking on third argument of chebfun2/coeffs2

danfortunato:
The third argument of `coeffs2` (i.e. the requested number of columns in the coefficient matrix) is ignored if the second argument (i.e. the requested number of rows) is greater than the number of rows stored in the `chebfun2`. For example,
```
>> f = chebfun2(@(x,y) x+y);
>> coeffs2(f,1,2) % correct
ans =
     0     1
>> coeffs2(f,2,1) % incorrect
ans =
     0     1
     1     0
```
The bug exists on line 44 of coeffs2.m. `if ( mf <= m )` should be `if ( nf <= n )`.

### DIFF
--- a/@chebfun2/coeffs2.m
+++ b/@chebfun2/coeffs2.m
@@ -41,7 +41,7 @@ if ( nargin > 1 )
     else
         cols_coeffs = cols_coeffs(1:m,:);
     end
-    if ( mf <= m ) 
+    if ( nf <= n )
         rows_coeffs = [ rows_coeffs ; zeros(n-nf, rf) ];
     else
         rows_coeffs = rows_coeffs(1:n, :);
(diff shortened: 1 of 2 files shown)


ajt60gaibb:
Good catch! Thanks, @danfortunato.  Would you be willing to add a test for this in `tests/chebfun2/test_chebcoeffs2.m`? (It could be your breaking example code above.)

danfortunato:
Sure. See https://github.com/chebfun/chebfun/pull/1991/commits/f950b7490234c8193b91ed63701304767c5ec451

ajt60gaibb:
Thanks, @danfortunato. Looks good.

PR #1996, opened 2016-11-06
norm(chebfun2) now gives spectral norm, not Frobenius, also added nuc…

nakatsukasayuji:
…lear norm

norm(chebfun2(f)) used to returned the L2 (Frobenius) norm. This 'fix' changes the default to the 2-norm, the largest singular value. Also fixed a cdr bug f = C*D*R' not C*D*R. 
Of course the Frobenius norm can be returned with the flag norm(f,'fro'). 
Also norm(f,'nuc') returns the nuclear norm, the sum of singular values.


nakatsukasayuji:
A correction to my earlier comment, norm(chebfun2) by default should be Frobenius (or L2) norm as before. norm(f,2) was also giving the Frobenius norm, contrary to the documentation (which said it returns the largest singular value). It's corrected to give the largest singular value. 
I've taken this opportunity to also add [U,H] = polardecomp(chebfun2) for the polar decomposition of a chebfun2.

ajt60gaibb:
Yes, one hopes `norm(f)` means L2 norms, i.e., analogue of Frob.  

Nice to have `polardecomp`.

PR #2008, opened 2016-11-23
code for rational trigonometric interpolation.

mhsnjvd:
This branch implements the command ```trigratinterp(f, m, n)``` which allows us to numerically compute a trigonometric rational interpolant of type ```(m, n)``` to a function ```f```.  Least-squares approximations are also possible.


oliviersete:
@mhsnjvd trigratinterp looks pretty neat!  Here are the first comments on the code (they look worse than it is):
- l. 62: the output `NN` seems not needed, maybe remove it?
- l. 105: `s` in `outArgs` should be removed.

In `parseInputs`:
- l. 139-141: definition of `N` can be removed, as it will be overwritten (either in l. 144 or in l. 163)
- l. 175-177: Please add in the help text that you can specify the domain with `'dom', [a,b]`, that is very useful!
- The error identifiers in l. 147 and l. 155 are the same, should they be unique?
- Please improve the error messages in l. 190, 202.
- l. 150-157: The help text does not mention that you can input `Xi` before `N`.  Should that be the case?
- Inputting `Xi` gives an error:
`
trigratinterp([2, 3, 1], 1, 0, 3, [-1, 0.5, 0.8])
`
Error using trigratinterp>parseInputs (line 202)
Argument unknown.

mhsnjvd:
@oliviersete thanks for all the comments, very helpful. I have addressed all of these! Please give it another go!

oliviersete:
@mhsnjvd thanks for the changes!  A few more comments on parseInput():
- l. 140: Maybe add an else statement to directly catch any weird inputs, e.g. f = {@cos, @sin}.
- l. 223-226: I think these lines can be removed because `length(xi) == NN` is always true at this point.  This is ensured when getting xi (l. 182-199).
- l. 229: Should the comparison be relative? (I.e., if you miss by e.g. `1e-16` you are still happy.)
- l. 240-246: Could these lines be moved before scaling xi to `[-1, 1]`? Then `f` can be sampled with `f(xi)`, without another linear transformation.

oliviersete:
@mhsnjvd Thanks for the changes.  Some more comments, it looks longer than it is, and they are mostly harmless:
- l. 74: Should trig_rat_interp be called with the relative tolerance "ts" instead of "tol"?
- l. 371: The inputs U and S of getCoeffs() are not used and could be removed.

`trig_rat_interp()`:
- input `th_type` is not used and can be removed.
- output `s` is not needed and can be removed (should also be removed from `outArgs`).
- l. 412-413: `fOdd` and `fEven` are set to zero, I guess this should be removed.
- l. 414: Currently, we have always `T=2` and this is the only place where `construct_matrices()` is called.  So it would be possible to simlify the `*2/T` there.
- l. 436-438: the scaling could be done without the loop: `diag(1./max(abs(fk), 1)) * sysMat`  (possibly with spdiags if the matrices may become large.)
- l. 475: If `n_new < 0`, could it happen that the loop does not break?

`constructTrigRatApprox()`:
- inputs `xi_type`, `mu`, `nu` are not used and can be removed.  (Or kept if needed for trigbary.)
- l. 486-487: Is the chopping necessary here?  I think this has already been done in `trig_rat_interp()`.  If the chopping is performed, `mu` and `nu` should be updated.

oliviersete:
@mhsnjvd Thanks for the changes!

PR #2016, opened 2016-12-05
AAA rational interpolation

oliviersete:
This pull requests adds the command `aaa(F, Z)` for AAA rational approximation.


trefethen:
@oliviersete, please change "Lloyd Nick" to "Lloyd N." (I never go by Lloyd Nick), and I suggest adding the arXiv reference.

"subsequent" should be "successive" 

In describing cleanup, I suggest the text 
``` 
 'cleanup', 'off': turns off automatic removal of numerical Froissart doublets
```

Thanks.

oliviersete:
@trefethen , thanks for the comments.  I made the suggested changes.

PR #2018, opened 2016-12-05
Fix colormap issue in separableApprox/plot

### LINKED ISSUE #2017: Colormap changed by phase portraits
Currently, `cheb.gallery2('waffle')` returns a surface plot of the waffle function.  Plotting
```
cheb.gallery2('airycomplex')
cheb.gallery2('waffle')
```
produces a surface plot with "wrong" colormap:  It's the one used for phase portraits, not the default one used for real-valued functions.

oliviersete:
closes #2017 : When plotting a complex-valued function, the
colormap is set to hsv (phase portraits).  In a subsequent plot of a
real-valued function, the colormap is not changed back to default.

### DIFF
--- a/@separableApprox/plot.m
+++ b/@separableApprox/plot.m
@@ -118,6 +118,7 @@ if ( ~isempty(varargin) )
 else
     if ( isreal( f ) )
         h = surf( f );
+        colormap default
         
     else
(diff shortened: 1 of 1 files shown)


trefethen:
It was a one-liner and it seems to work.

PR #2028, opened 2016-12-12
Suppress warning message in aaa()

### LINKED ISSUE #2027: small bug in `aaa`
There's a bug in `aaa` that I presume is very small.  This works fine:
```
>> aaa(@exp);
```
but this does not:
```
>> aaa(@exp,'mmax',3);
Warning: Function not resolved using 16385 pts. 
> In aaa>aaa_autoZ (line 387)
  In aaa (line 46) 
```

oliviersete:
when 'mmax' is provided and the sample set is not.
Fix #2027 .

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -38,12 +38,12 @@ function [r, pol, res, zer, zj, fj, wj, errvec] = aaa(F, varargin)
 
 
 % Parse inputs:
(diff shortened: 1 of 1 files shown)


trefethen:
Beautiful.  Thanks.

PR #2029, opened 2016-12-12
A new gallerysphere example

gradywright:
Includes a function representing the vorticity of an idealized zonal jet stream in the northern hemisphere that is used in numerical weather prediction test problems.


trefethen:
Looks good.  The new example "jet" wasn't in the random switch, so I added it.  Now I will merge.

gradywright:
Thanks @trefethen for doing this so quickly!

PR #2032, opened 2016-12-13
Remove terms with zero weights from aaa-approximant.

oliviersete:


bhashemi:
@oliviersete, (just to be sure) are you trying to find points with the "exact" zero weigths (and not "small" weights)?

oliviersete:
@bhashemi thanks for looking at this! Yes, this is indeed intended to remove weights that are exactly zero, which sometimes appear.
The corresponding term in the barycentric formula, w_j / (z-z_j), does not contribute to the value of the rational function, but I think it should be removed for aesthetic and practical reasons.
The practical reason is that a zero weight results in a (numerical) Froissart doublet at the point z_j, due to the way we currently compute the poles and zeros of the rational function.

PR #2036, opened 2016-12-15
Fixing a bug with InputParser where an error should be thrown when th…

### LINKED ISSUE #2035: Chebfun not constructed with Chebyshev coefficients of the second kind
I tried to create a chebfun using Chebyshev coefficients of the second kind. The 'chebkind' argument seems to be ignored when 'coeffs' is given as an option. I think this would be a sensible and convenient extension to chebfun's functionality. 

```
f = chebfun('exp(x)');

c1 = chebcoeffs(f, 'chebkind', 1);
c2 = chebcoeffs(f, 'chebkind', 2);

f1 = chebfun( c1, 'coeffs', 'chebkind', 1);
f2 = chebfun( c2, 'coeffs', 'chebkind', 2);

x = linspace(-1,1);
plot(x, f1(x), x, f2(x))
```

kuanxu:
…e properties coeffs and chebkind are both specified.

This is to fix #2035.


oliviersete:
@kuanxu , the new error message is triggered even if `'coeff'` and `'chebkind'` are not specified, e.g. when running `chebfun('x')`.

kuanxu:
@oliviersete Thanks! I was just too hasty and forgot to run all tests. All tests pass now. Could you please run all tests on your machine once again to double-check?

Merry X'mas (belated) and happy new year, everyone!

bhashemi:
@kuanxu , the code in the issue #2035 still does not give an error message in this branch. Did you want to say something like
```
    if ( iscell(op) || isnumeric(op) && ...
```
instead of 
```
    if ( iscell(op) && isnumeric(op) && ...
```
in Line 689 of `@chebfun/chebfun`?

kuanxu:
@bhashemi Thanks, Behnam. This is a bit embarrassing. Now it's fixed and two tests have been added. See if all tests pass on your machine.

PR #2039, opened 2016-12-16
add diskfun helmholtz solver and tests 

heatherw3521:
This PR adds a Helmholtz solver on the disk; this code is based on and very similar to  the `diskfun.poisson` code.


ajt60gaibb:
There was good reason to have two codes for `spherefun.poisson` and `spherefun.helmholtz` because of the integral constraint.  For diskfun, there is no distinct and so I have changed `diskfun.poisson` to call `diskfun.helmholtz` with `K=0`.

PR #2044, opened 2017-01-02
Feature NUFFT

ajt60gaibb:
This introduces quasi-optimal complexity algorithms (`O(Nlog N log(1/eps)/loglog(1/eps))` operations) for: 

1) The nonuniform DFT of type-I (NUDFT-I), which is equivalent to evaluating a Fourier series with noninteger frequencies at equispaced points, 

2) The nonuniform DFT of type-II (NUDFT-II), which is equivalent to evaluating a Fourier series at nonequispaced points, (This is the default transform.) 

3) The nonuniform DFT of type-III, which is equivalent to evaluating a Fourier series with noninteger frequencies at nonequispaced points, 

4) Inverse transform for NUDFT-I, which is equivalent to computing coefficients of Fourier interpolants with noninteger frequencies from equispaced function samples, and

5) Inverse transform for NUDFT-II, which is equivalent to computing coefficients of Fourier interpolants from nonequispaced function samples.

This MATLAB code may nicely accompany the recent preprint "Trigonometric Interpolation and Quadrature in Perturbed Points" by Austin and Trefethen [https://arxiv.org/abs/1612.04018](url). This provides functionality in Chebfun for experimenting with trigonometric interpolants of periodic functions on `[0,1]` at perturbed points. 

For example, to evaluate a Fourier series at random samples we can do:
```
N = 1e5; 
cfs = 1./(1:N)' + 1i./(1:N)';            % Fourier coefficients
x = rand(N,1);                            % Random evaluation points on [0,1]. 
tic, vals = chebfun.nufft(cfs, x); toc  % NUDFT-II, including computing a "plan"
Elapsed time is 0.133295 seconds. 
```

Just like the FFTW software system, one can make the transform faster by "planning it". See help text of `chebfun.nufft` for more details. In above, planning reduces the timing to `0.080807` seconds, which is a similar cost to 20 FFTs for me. (Planning is more significant for smaller values of `N`.) 

The underlying algorithm is described in a manuscript that is in preparation.  This is joint work with Diego Ruiz-Antoln from the Universidad de Cantabria.


kuanxu:
@ajt60gaibb This is nice! I have been longing this for quite a while. How is the new method compared with Dutt's? Do you have a breakeven number in Matlab or Julia? Why did you put it in `chebfun` class? Intuitively, these useful routines could stay standalone and be called by `chebfun`. Of course, I don't exactly the context.

ajt60gaibb:
The underlying algorithm is based on low-rank approximation and linear algebra.  It is quite different from Dutt&Rokhlin, Potts' NFFT, etc.  In particular, no oversampling is required. 

In many circumstances, we just observe that the NUDFT matrix `tilde{F}` can be expressed as 
```
tilde{F} = A \circ F,
```
where `A` can be well-approximated by a low rank matrix, `F` is the DFT matrix, and `\circ` is the Hadamard product.  Once you calculate a low rank approximation for `A` in a numerically stable way (which is hardest part), then you can express `tilde{F}` as a sum of diagonally-scaled DFT matrices... and hence, there's a fast algorithm based on FFTs. 

Give me till the end of the week and I'll get the finished manuscript to you.

ajt60gaibb:
I put it in the `chebfun` class because that's where `ndct`, `dlt`, etc. are.  I will happily move it or add extra command that are available to the user. 

If I am allowed to plan, then it's always faster:
```
N = 10; 
x = rand(N,1);
c = rand(N,1) + 1i*rand(N,1);

tic 
f = zeros(size(x,1),1);
omega = 0:size(x,1)-1;
for j = 1:numel(f)
    f(j) = exp(-2*pi*1i*x(j)*omega)*c;
end
toc 
Elapsed time is 0.001639 seconds.

tic
[fast, p] = chebfun.nufft( c, x );
toc
Elapsed time is 0.026604 seconds.

% Planned version: 
tic
p(c);
toc
Elapsed time is 0.000286 seconds.
```
but maybe you consider that cheating??    (This is why I did not add a "break-even if" statement in the MATLAB code. The FFTW software deals with this for me.) 

Without planning, the break-even point is about `N = 1024` in MATLAB. In Julia, it's very similar for me (with `FFTW.set_num_threads = 4`, note that MATLAB has multithreaded FFTs by default while Julia does not yet).   The breakeven point is about `N = 512` if I turn off multithreaded FFTs in Julia because of the overhead of threading, I guess.

kuanxu:
@ajt60gaibb Thanks for explanation. The idea is neat. I guess this is a clearer way to understand the second last theorem (or corollary, can't remember exactly) in Dutt's first NUFFT paper, where they show a single Fourier mode with non-integer frequency can be well approximated by a very small number of Fourier modes of nearby integer frequencies. But they didn't explore much of the sparsity/skeletonity(sorry for this made-up word) of the matrix `A`.

> If I am allowed to plan, then it's always faster:

You meant faster than both the naive implementation that you listed and Dutt's or Pott's algorithms, right?
```
tic
[fast, p] = chebfun.nufft( c, x );
toc
Elapsed time is 0.026604 seconds.
```
So this is the total time of planning and actual computing (and the plan `p` is returned as part of the result)?
 
My impression is that there are a few NUFFT implementation on github (can't remember the languages). If any of them is in Matlab or Julia, you might consider to compare with them (w & w/o planning). I think at the time of Dutt's writing their papers, FFT planning had not been invented. Look forward to seeing the paper anyway.

ajt60gaibb:
We have a MATLAB, Julia, and FORTRAN implementation of the algorithm. So, yes, we are hoping to compare against the NUFFT software and NFFT software (Julia implementation of it). We are perhaps a little faster, but I don't see that as particularly important. I am arguing that this is just a more convenient algorithm or way of thinking about it that can be understood in a minute. 

Hmmm... I never thought of Dutt's corollary as another way of saying low rank approximation. That's a ***very*** good point.  Thanks, I will think about that!  Nice.

ajt60gaibb:
Yes, I think of "planning transforms" as from the 1998 FFTW paper.

ajt60gaibb:
Related manuscript: 
https://arxiv.org/abs/1701.04492

bhashemi:
@ajt60gaibb , here are 7 minor comments. 5 of them are about the help-text. I might have overlooked the code, but the idea is just enhancing readability. I'd be happy to merge once the comments are addressed.

1) I guess there is a typo in the help text of the subroutine `constructAK` in `chebfun.nufft`. Shouldn't `x-s_j/N` be `x_j-j/N` like Formula (2.3) in your paper?

2) I found it a bit confusing that the variable `c` is used in the function-handle which calls MATLAB's `fft` while `c` is an input to the main code also. I suggest the variable `c` in the function-handles to be renamed e.g., to `coeffs` or something else. In the code `chebfun.nufft`, this comment applies to Lines 44, 87 and 103. This applies to a few lines of  `chebfun.inufft` as well.

3) In `chebfun.inufft`, it would be more consistent to rename the subroutines `nuifft1` and `nuifft2` to `inufft1` and `inufft2`, respectively. Error messages in this code also need a corresponding update. Another issue is that some of the error messages in `chebfun/inufft` mention `nufft` instead of `inufft`.

4) May I suggest an interchange of the names of the variables `c` and `f` in `chebfun.inufft`? Here, `f` is given and `c` is the output which conceptually agrees with the help text. But the actual code is using reverse names!

5) Isn't there a redundant `N` in Line 10 of the help text of `chebfun/nufft` and `chebfun/inufft` where `A_{jk}` is defined for the default type-II transforms?

6) Aren't there 2 unnecessary `N`'s in Line 11 of  `chebfun/nufft2` where `F_{st}` is described?

7) The code `test_nufft2` currently gives an error message w.r.t sizes.

Beautiful work with beautiful mathematics!

ajt60gaibb:
@bhashemi, I've fixed your things above.

PR #2056, opened 2017-02-09
improve accuracy in Chebfun2/roots

trefethen:
I've just changed half a dozen lines in separableApprox/roots to compute an empirical arc length before interpolating to the Chebyshev grid, to handle non-smoothly spaced data.  This seems to improve accuracy of arc length and area coefficients by several digits.  For example (see `tests/chebfun2/test_roots`), the sequence
```
f = chebfun2( @(x,y) x.^2 + y.^2 - 1/4 );
c = roots(f);
arclength = sum(abs(diff(c)));
area = abs(sum(real(c).*diff(imag(c))));
```
used to give 2 digits of accuracy and now it is 5 digits.  A discussion of this mathematics can be found in the Chebfun team dropbox folder as Chebfun2Roots.pdf.

It would be nice if somebody could review this quickly as it's short and kathryn.gillow@maths.ox.ac.uk would like to take advantage of this feature.

### DIFF
--- a/@separableApprox/roots.m
+++ b/@separableApprox/roots.m
@@ -3,11 +3,11 @@ function r = roots( f, g, varargin )
 %   R = ROOTS(F), returns the zero contours of F as a quasimatrix of chebfuns.
 %   Each column of R is one zero contour. This command only finds contours when
 %   there is a change of sign and it can also group intersecting contours in a
-%   non-optimal way. Contours are computed to, roughly, four digits of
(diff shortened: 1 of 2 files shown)


nakatsukasayuji:
@trefethen @nakatsukasayuji and weideman@sun.ac.za played around with chebfun2/roots, and now we get 8 digits instead of 2 for simple problems! We have added a Newton refinement step in addition to interp1 curve fitting.

PR #2057, opened 2017-02-10
Fix the way BC norms are computed for chebmatirx inputs to chebop bcs.

nickhale:


asgeirbirkis:
@nickhale I'm happy to merge this, but curious where this was causing errors?

PR #2059, opened 2017-02-13
Feature randnfun

trefethen:
With discussion with abdullateef.hajiali@maths.ox.ac.uk, I've added new codes 'randn' and 'randfun' in our root directory.  This are continuous analogues of 'randn', producing smooth random functions (periodic or not) with a specified spatial scale DT and domain.  The underlying mathematical ideas may evolve but I am eager to get some of this material into Chebfun to help our explorations and also to help in researching associated questions of stochastic ODEs driven by smooth data and in writing Chapter 12 of _Exploring ODEs_.

These codes are short and should be easy to review.

Questions:

(1) Is this "rectangular window" of Fourier coefficients right, or would Gaussian be better?  (I am speaking here of the covariance function when 'randnfun' is regarded as generating a sample from a Gaussian process.)

(2) Is our method of getting nonperiodic functions by restricting periodic ones good?  Can the unwanted correlations across boundaries that it introduces be quantified?

(3) What are the right rectangle, box, disk, and sphere analogues?  (belyaev@maths.ox.ac.uk) is a local expert at Oxford.)

(4) What literature is there on related methods?

### DIFF
--- a/randnfun.m
+++ b/randnfun.m
@@ -1,26 +1,33 @@
-function f = randnfun(dt, dom)
+function f = randnfun(dt, n, dom)
 %RANDNFUN   Random smooth function
-%   F = RANDNFUN(DT) returns a smooth chebfun on [-1,1] with space
-%   scale DT and standard normal distribution N(0,1) at each point.
-%   It can be regarded as a sample path of a Gaussian process.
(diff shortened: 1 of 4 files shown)


sfilip:
This looks good and clean. I'm happy to merge. Thanks @trefethen!

PR #2062, opened 2017-02-15
Use NaN data for the dummy plot in chebfun.plot().

### LINKED ISSUE #2061: chebfuns are grayed out in plot legends in R2016b.
On development in MATLAB R2016b, if I do
```
>> f = chebfun(@exp);
>> g = chebfun(@sin);
>> plot(f, 'b', g, 'r');
>> legend('f', 'g');
```
I get the following plot:

p1

See how the items in the legend are grayed out?  Compare with the plot
generated by plotting the functions "manually" as in the following:
```
>> x = linspace(-1, 1, 256).';
>> plot(x, f(x), 'b', x, g(x), 'r');
>> legend('f', 'g');
```

p2

Here, everything looks OK.

What happens if we plot `f` manually and `g` using the `chebfun` plot method?

```
>> plot(x, f(x), 'b');
>> hold on;
>> plot(g, 'r');
>> hold off;
>> legend('f', 'g')
```

p3

The item for `f` looks OK, but that for `g` has been grayed out.

The reason this happens has to do with [these lines](https://github.com/chebfun/chebfun/blob/15179d2751ffb7441b1e23dc9d885f6e477f0498/%40chebfun/plot.m#L414-L420) in `@chebfun/plot.m`.  Specifically, the issue is that the legend is built from the graphics objects on the "dummy" plot, but those objects have their `Visible` property set to `'off'`.  Apparently, as of R2016b MATLAB now grays out items in the legend that are invisible on the plot.  Remarkably, this fact actually appears in the R2016b documentation for the `legend()` function:
```
The legend reflects the visibility of graphics objects in the axes. Graphics objects that
have a `Visible` property set to `'off'` appear as grayed out items in the legend.
```
I find no mention of this in the R2016b release notes (though they do mention some other changes to `legend()` that I don't think affect us).  Nevertheless, I haven't been able to find similar text in the documentation for R2015a or R2015b, so I presume that R2016b is the release that introduced this behavior.

Unfortunately, I don't know what to do about this.  The games we play with the `Visible` and `HandleVisibility` properties in `plot()` are all there for good reasons.  We can't set the dummy plot's `Visible` property to `'on'` or it will draw on top of the plots that are already there.  Neither can we set `'HandleVisibility'` to `'on'` for any of the other plots, for then while the legend will not be grayed out, the line styles in the legend entries be wrong.  I also looked to see if the graying-out behavior could be adjustable via some legend graphics property, but it does not look that way.

aaustin141:
Previously, we just replotted the existing line data and set the plot's
"Visible" property to 'off'.  This doesn't work properly anymore in
R2016b, since MATLAB grays out the legend entries for graphics items
marked invisible.

The fix is easy enough:  we plot NaN data for the dummy plot (but set
the line and point styles appropriately) and leave "Visible" set to
'on'.  MATLAB doesn't draw anything on the plot for NaN data, so we're
good.  A little ugly, but we already use this trick elsewhere for other
things.

Closes #2061.

### DIFF
--- a/@chebfun/plot.m
+++ b/@chebfun/plot.m
@@ -370,8 +370,18 @@ if ( ~verLessThan('matlab', '8.4') )
     set(gca, 'ColorOrderIndex', originalColorOrder);
 end
 
+% Generate data for the dummy plot for the legend.  dummyData is the same as
+% lineData but with the numeric entries replaced by NaNs of suitable dimension
+% so that they don't show up on the plot but are still "active" for the legend.
+dummyData = lineData;
+for i = 1:1:length(dummyData)
+    if ( isnumeric(lineData{i}) )
+        dummyData{i} = NaN(1, size(dummyData{i}, 2));
+    end
+end
+
 % Plot the dummy data, which includes both line and point style:
-hDummy = plot(lineData{:});
+hDummy = plot(dummyData{:});
 if ( ~isempty(lineStyle) || ~isempty(pointStyle) )
     set(hDummy, lineStyle{:}, pointStyle{:});
 end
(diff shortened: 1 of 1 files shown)


oliviersete:
Thanks @aaustin141 , the fix seems to work fine.

PR #2063, opened 2017-02-19
Chebfun2 roots again

trefethen:
I've made various improvements to Chebfun2 roots so that in simple cases at least, it gets very good accuracy.  Since @sfilip reviewed the last round of improvements to roots, perhaps he might take a look at this one too?

Here's an example of the arc length of a parabola.  All digits displayed are correct.
```
>> f = chebfun2(@(x,y) y-x.^2);
>> sum(abs(diff(roots(f))))
ans = 2.957885715089195
```

Roots is still not perfect, especially when there are intersecting or near-intersecting curves.


sfilip:
I'm happy to merge this. Thanks @trefethen!

PR #2077, opened 2017-03-16
fix aaa cleanup

nakatsukasayuji:
This fixes a bug in the cleanup phase of aaa (invoked to remove spurious pole-root pairs when present).


oliviersete:
@nakatsukasayuji Thaks for the fix.  I think swapping the two lines where the suport points are removed from Z and F already does the trick, though.

PR #2083, opened 2017-03-28
speedup in rational remez

sfilip:
Considerable speedup in the search for the next reference set. Thanks @nakatsukasayuji for the hard work!


trefethen:
This is very exciting!

Below is a case that seems to give trouble.

I also have a general thought.  This code produces text as it runs, which I think is fine given the advanced nature of the topic and the complexity of what is being done.  However, when a string of text is being printed, the user would like to know at the end whether we apparently have success or failure.  This is obvious no doubt to @sfilip and @nakatsukasayuji, but it won't be obvious to users.  Could suitable final text lines be added?

```
Trial interpolant too far from optimal.
Trying AAA-Lawson-based initialization...
Trial interpolant too far from optimal.
Trying CDF-based initialization with step size 1...
(0,1)

Trial interpolant too far from optimal.
Trying CDF-based initialization with step size 2...
(0,1)

Trial interpolant too far from optimal.
```

PR #2084, opened 2017-03-29
feature-transpose-is-adjoint

nickhale:
Add `ctranspose` method for CHEBOP class so that `A'` is a wrapper for `adjoint(A)`.


trefethen:
@sfilip and I have looked at this very nice idea; thanks @nickhale!  We tweaked the help text a bit and will now merge.

Issue #2094, opened 2017-04-13
f(x) should give an error message if f is a chebfun2


trefethen:
If you try to evaluate a chebfun2 at a single argument, a number is produced:
```
>> cheb.xy; x(0)
ans = 0
```
It seems to me this should give an error message.  I note that this is what happens in Chebfun3:
```
>> cheb.xyz; x(0)
Error using chebfun3/subsref (line 58)
Can only evaluate at triples (X,Y,Z), a CHEBFUN with 3 columns, a CHEBFUN2V or a CHEBFUN3V. 
```
@ajt60gaibb, might you be able to fix this?

oliviersete:
@trefethen Currently f(z) is interpreted as f(real(z), imag(z)), when f is a chebfun2.  This was introduced to fix #1956 .

trefethen:
Ach, that's fascinating!  Of course I had forgotten.

If f is a complex chebfun2, we'd like to be able to evaluate e.g. f(.3).
If f is a real chebfun2, perhaps f(.3) should give an error message?
What do you think @oliviersete?  Do others have a view?

oliviersete:
For a real chebfun2, `f(x)` could indeed give an error message (or a warning that `f(x)` is interpreted as `f(x,0)`).

ajt60gaibb:
I can easily make a change, if it is decided that `f(z)` should be an error when `isreal(f) = 1`.

trefethen:
Yes please!

PR #2099, opened 2017-04-14
Give error for f(x) if isreal(f) = 1 and isreal(x) = 1.

ajt60gaibb:
```
f = chebfun2(@(x,y) x);
>> f(0)
Error using separableApprox/subsref (line 77)
Object is a function of two real variables
```


ajt60gaibb:
Closes #2094.

PR #2104, opened 2017-04-16
Arrowplot improvements

### LINKED ISSUE #2103: handle to `arrowplot` doesn't include the arrowhead
On a plot with a bunch of curves I like to be able to delete some and add others, e.g. like this:
```
h1= arrowplot(...);
delete(h1)
h2 = arrowplot(...);
...
```
However I notice that when I execute `delete(h1)`, it deletes just the curve, not the arrowhead.  Is it possible for both curve and arrowhead to be attached to a single handle?

asgeirbirkis:
* Introduce 'markerSize' option for arrowplot (case insensitive)

* Return both arrow and line graphic objects.

* Cleanup of logic for parsing inputs.

Closes #2103.

### DIFF
--- a/@chebfun/arrowplot.m
+++ b/@chebfun/arrowplot.m
@@ -9,11 +9,12 @@ function varargout = arrowplot(f,varargin)
 %   Arguments can also be quasimatrices, in which case several curves are
 %   plotted.
 %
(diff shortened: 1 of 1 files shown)


asgeirbirkis:
Closes #2103 .

PR #2111, opened 2017-04-21
feature-quasimatrix-conv

nickhale:
Add quasimatrix support for CONV().

### DIFF
--- a/@chebfun/conv.m
+++ b/@chebfun/conv.m
@@ -48,7 +48,7 @@ transState = f(1).isTransposed;
 % Support for quasimatrices:
 nf = numColumns(f);
 ng = numColumns(g);
-if ( nf > 1 && ng > 1 )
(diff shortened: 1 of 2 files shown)


sfilip:
Hi Nick, @trefethen and I were looking at your changes. The following example doesn't work
```
>> cheb.x;
>> f = abs(x);
>> g = exp(-100*x^2);
>> plot(conv(f,g))
>> plot(conv([f 2*f],g))
```
We think you intended for this to work. Are we correct in our assumption? Is there an error in the if and else if statements?

Thanks!

nickhale:
Yes, there was a silly mistake. Thanks for picking it up.

On a separate note, I wonder if it's worth having a static method that deals with this `if-else` statements? Presumably there are several places where we must loop over entries to support array-valued chebfuns/quasimatrices in this way? @aaustin141 may have thoughts?

Issue #2114, opened 2017-04-24
Bug in / vs ./


sfilip:
Playing around with the following example caused Chebfun to issue a lot of warnings for the second call:

```
>> f = chebfun('1./(2-x)');
```

vs

```
>> f = chebfun('1/(2-x)');
```

@nickhale , can you take a look to see what's going on?
Thank you!

nickhale:
Note there is also a problem with `chebfun(@(x) 1/(2-x))` so it is not an issue of strings inputs.

The issue is that 
```
>> 1/(2-chebpts(9))
ans =
    0.3333         0         0         0         0         0         0         0         0
```
returns a row vector (since we are calling `/` on a double and not a chebfun, we have no control over this). Vectorize then thinks that you've made a mistake and so it transposes the result `op -> (1/(2-x))'`. However this is not the same as `1./(2-x)`.

Possible solutions are:
 * Throw an error instead of a warning, which will cause the op to be vectorized rather than transposed.
 * Transpose the input in this situation, rather than the output (this may or may not work in general).

Unfortunately I don't have time to investigate this right now.

sfilip:
Thank you! I'll try to play around and investigate this further.

PR #2117, opened 2017-04-28
Vectorizing eg. 1/(2-x)

sfilip:
I've made a very small change trying to address issue #2114 . It consists of vectorizing an input, instead of transposing it. It's very minor, but in a crucial region of logic. @trefethen and I won't be comfortable unless @nickhale looks at this.


nickhale:
This looks reasonable.

sfilip:
Thanks @nickhale! I'm closing and reopening this since I had a small hiccup in pulling the latest changes from development.

PR #2138, opened 2017-05-11
Feature spinsphere imexbdf4

Hadrien-Montanelli:
Implement multistep scheme `imexbdf4` as an alternative of `lirk4`.


nakatsukasayuji:
Looking awesome!

PR #2147, opened 2017-05-18
Feature dotty plotcoeffs

trefethen:
I've made fundamental changes to `plotcoeffs`: it now plots dots instead of lines, and values are never set to an artificial epsilon level.  To illustrate, here are the results of
```
plotcoeffs(chebfun('cos(5*x)'))
```
with traditional Chebfun and the new revision.  You can always override linetypes in the usual fashion.
fooold
foonew


sfilip:
Thanks @trefethen!

Issue #2157, opened 2017-05-23
Spherefun roots is broken


trefethen:
For some time now, Spherefun roots has been broken; unfortunately the associated test file is so minimal that it doesn't detect this.  Below is an example.

It is likely that the problem arose when @trefethen improved roots for Chebfun2 and SeparableApprox.  Nevertheless perhaps one of @ajt60gaibb, @gradywright, @heatherw3521 might be able to take a look and see what they think?

```
>> cheb.xyzsphere
>> roots(x-.2)
Error using spherefun/grad
Too many output arguments.

Error in separableApprox/roots (line 80)
        [fx, fy] = grad(f);

Error in spherefun/roots (line 25)
rts = roots@separableApprox(varargin{:});
```

gradywright:
This is the same bug as #2118.

ajt60gaibb:
This looks like a relatively serious bug because of the way vectors in `spherefunv` are represented in the cartesian coordinate system.  This could take a while...

ajt60gaibb:
Partial fix merged in #2161.

ajt60gaibb:
Duplicate of #2118 so closing.

Issue #2118, opened 2017-04-28
Roots is broken in spherefun 


Hadrien-Montanelli:
Try the following:

`f = spherefun(@(x,y,z) 2*sinh(5*x.*y.*z));
r = roots(f);`

gradywright:
This may take sometime to sort out.  If all you are interested in is making a plot of the zero contours of `f` then try `contour(f,[0 0])`.

PR #2161, opened 2017-05-23
Fix spherefun roots

gradywright:
Partial fix for #2157 and #2118.  I say partial fix because the accuracy of the computed level curves is not great.  Also, the curves may turn back slightly on themselves because periodicity in lambda has not been properly enforced.


ajt60gaibb:
Kinda cheating, but I'm OK with this for now.

@spherefun/roots.m
```diff
@@ -1,28 +1,33 @@
...
+if ( isempty( f ) )
     r = []; 
     return
 end
 
-rts = roots@separableApprox(varargin{:});
+% Convert f to a chebfun2 on a larger domain so that we can properly
+% pick up roots that go accross -pi to pi in lambda.  This will allow us 
+% to account for the inherent periodicity of the sphere.
+rts = roots@separableApprox( chebfun2(f,[-pi-2*pi/100 pi 0 pi]) );
```
ajt60gaibb: Why is the domain only increased in the `lambda` variable and along only one edge?

gradywright: Because it's possible to have a contour run along `lambda=-pi` (or `pi`), `0 <= theta <= pi`, but it's not possible for there to be a contour along `theta = 0` (or `pi`) and `-pi <= lambda <= pi`.

ajt60gaibb: I see. Thanks.

PR #2167, opened 2017-05-25
Feature improve rotate and fix bug in Spherefun constructor

gradywright:
Improves the speed of the `fastSphereEval` code by removing the giant loop over all the evaluation points.  (Perhaps @nickhale can even offer further improvements with cell arrays).

Fixes a bug in the Spherefun `constructor` where the maximum allowable rank calculation in Phase I was not quite right. 

With the construct bug fixed, rotate now does not need to use twice the number of sampling points. As a result rotate is now even faster.


ajt60gaibb:
The `2*max(m,n)` factor in `spherefun.rotate()` was a constructor bug! That solves that mystery. This means we have another `x2` speed up. Sweet:
```
f = randnfunsphere(0.03);
tic, f = rotate(f, .1,.1,.1); toc 
Elapsed time is 0.755444 seconds
```

PR #2172, opened 2017-05-31
Even faster spherefun rotate

ajt60gaibb:
Slightly modify the low-rank representation of the internal matrices `Ay` and `Ax` so that the most expensive line can be done in real arithmetic.  This gives us another x2 speed up in `spherefun.rotate`! 

Now, we have 
```
f = randnfunsphere(0.03); 
tic
g = rotate(f, 0.1, 0.1, 0.1); 
toc
Elapsed time is 0.388473 seconds.
```

We never use the `feval` method now, the `nufft` is almost always faster, and usually by orders of magnitude. 

...We need to start finding out what the world-record is... 👍

### DIFF
--- a/@spherefun/fastSphereEval.m
+++ b/@spherefun/fastSphereEval.m
@@ -133,12 +133,13 @@ X = X(c(:,2));
 breaks = find(diff(srt_ic)); 
 breaks(end+1) = numel(idic); % Include the endpoint
 cnt = 1;
+ov = ones(size(YY,2),1);
 vals = zeros(M, N); % Allocate storage for output.
 for idx = 1:size(c,1)
     % Get the (ii,jj) values that use the same X and Y.
     kk = idic(cnt:breaks(idx));
     % Do the inner product:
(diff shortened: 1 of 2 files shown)


ajt60gaibb:
I would like to just make sure that the structure that I exploited *always* holds before this PR is merged.

nickhale:
Nice observation! I think it would always be true, since it's not depending on the data. 

This line could be improved though:
```
V1 = chebT(K1-1, 2*mm'/m) / Dy;
```
I think you want
```
DyInv = diag(1./(1i).^(0:K1-1));
V1 = chebT(K1-1, 2*mm'/m) * DyInv;
```
or something similar.

ajt60gaibb:
OK. Yes, the structure to get real-arithmetic in the inner-loop always holds.   This is a property of the even-periodic, odd-antiperiodic decomposition of the rank-1 terms from the spherefun constructor and hence, also in the CDR factorization of a Spherefun's coefficient matrix.

PR #2173, opened 2017-06-05
Faster chebtech restrict

ajt60gaibb:
I have been wanting to do this for a while because it completes the list of immediate applications of NUFFT in Chebfun.  Here, we do a number of things: 

1) Add `chebtech.fastChebyshevEval`, which is a nonuniform discrete Chebyshev transform. This also could be adapted to be a general `chebfun.nudct` code too (at the moment `chebfun.nudct` only allows for Legendre nodes).  @nickhale: only 16 FFTs are required in the code, as desired

2) An optimal complexity `chebtech.restrict`, via an optimal complexity `chebtech.feval`. That is, given a polynomial of degree `n` on [-1,1], one can now restrict it on `[a,b]` in O(n log n) operations. Previously, the complexity was `O(n^2)`. The breakeven point inside `chebtech.feval` was obtained from the following code:
```
nn = round(logspace(1,5,50));
j = 1; 
for n = nn 
c = rand(n,1); 
x = linspace(-1,1,n)'; 
s = tic;
vals = chebtech.clenshaw( x, c ); 
t_clen(j) = toc(s); 
s = tic;
vals = chebtech.fastChebyshevEval( x, c ); 
t_fast(j) = toc(s); 
j = j + 1; 
n
end

%%
loglog(nn, t_clen, 'linewidth',2), hold on, 
loglog(nn, t_fast, 'linewidth',2)
xlabel('n','fontsize',16)
legend('Clenshaw','NUDCT', 'Location','SouthEast')
set(gca,'fontsize',16)
```
timings

3) An optimal complexity subdivision algorithm in the 1D `roots` command. This makes chebfun.roots slightly faster for (super!) high degree polynomials.  **Warning:** I needed to change the definition of `tailMmax = eps*norm(c, 1);` to `tailMmax = 5*eps*norm(c, 1);` to get this to work. 

```
f = chebfun(@(x) sin(15000*pi*x)); 
% Only clenshaw for subdivision:
tic, 
r = roots(f);
toc
Elapsed time is 9.127030 seconds.
norm( r - linspace(-1,1,30001)' )
ans =
     1.820224107610997e-14
% Fast transform when poly. degree is large: 
tic 
f = roots(f); 
toc
Elapsed time is 4.840047 seconds.
norm( r - linspace(-1,1,30001)' )
ans =
     2.643939665708378e-14
```

I am not expecting this to be merged without further discussion (because it slightly changes the roots command) but would like to hear the opinions of @nickhale and @trefethen on this PR.


nickhale:
This needs some work.
 * There are two `fastChebyshevEval` codes; one inside `feval` and one in its own .m file.
 * Currently there is no improvement in `restrict` or `feval` as the code is branching on `m`, the number of columns.
 * `fastChebyshevEval` eval does not currently support array-valued chebtechs.
 * I don't like the name. I think it would make more sense to replace the existing `nudct` code. 
 * If the new `nudct` code is slower than the previous special case, we can keep that one inside the `dlt` code.

nickhale:
With regards to replacing the existing specialised `ndct`, the test below gave the following results:
fig1
fig2

The timings of the two fast codes seem comparable and unfortunately both have a high break-even with the direct method (clenshaw). However, since both can work directly in theta space, they are more accurate, even for small n. In a head-to-head I find `ndct` about 2 times faster than `fastChebyshevEval` (not sure why this isn't the same as in the loop and the figure about..).

```
nn = ceil(logspace(2,4,50));
seedRNG(0)
for n = nn
    pause(eps)
    disp(n);
    
    c = rand(n, 1);
    c = c./(1:n).^1.5'; % A small amount of decay in coefficients
    
    tic
    v1 = chebfun.ndct(c); 
    t1(n) = toc;
    
    tic
    [x, ~, ~, t] = legpts(n);
    v2 = chebtech.fastChebyshevEval(x, c, t); 
%     v2 = chebtech.fastChebyshevEval(x, c); 
    t2(n) = toc;
    
    tic
    x = legpts(n);
    v3 = chebtech.clenshaw(x, c); 
    t3(n) = toc;
    
    err(n) = norm(v1 - v2, inf);
    err2(n) = norm(v1 - v3, inf);
    err3(n) = norm(v2 - v3, inf);
    
end

figure(1)
loglog(nn, t1(nn), nn, t2(nn), nn, t3(nn), 'LineWidth', 3)
legend('ndct', 'fast', 'clenshaw')
title('time'), xlabel('n')
set(gca, 'fontsize', 20)
figure(2)
loglog(nn, err(nn), nn, err2(nn), nn, err3(nn), 'LineWidth', 3)
legend('ndct vs fast', 'ndct vs clenshaw', 'fast vs clenshaw')
title('comparative error'), xlabel('n')
set(gca, 'fontsize', 20)
```

ajt60gaibb:
I did the following: 
- Replaced the `ndct` code with the `fastChebyshevEval` code,
- `dlt` now calls the function `ndct_legpts`, 
- Removed `fastChebyshevEval` as it is now in `ndct` (Is that OK, because now chebtech knows that chebfun exists?)

nickhale:
This is better, but more thought needs to be given to the branching on clenshaw vs fast. In particular, evaluating long chebfun at a single point should be done with clenshaw, not the ndct.

The naming still doesn't seem right (I know @ajt60gaibb is also not happy). Perhaps one option is to have nudct be the fast algorithm for the related discrete cosine transform, and have the fast Chebyshev evaluation call this?

ajt60gaibb:
ping

nickhale:
All this needs is for some experimentation to decide the switching strategy between the direct approach (Clenshaw) and the asymptotically fast approach.

ajt60gaibb:
OK. Here is a comparison of the two approaches. On my computer, I observe that `clenshaw` is faster in the green region and `ndct` is faster in the blue region. The black lines indicate where I put the switch between the two algorithms in the `chebtech/feval`: 

breakeven

trefethen:
Well that's pretty fascinating!  In effect you've plotted a single contour of speed(clenshaw)/speed(ndct).  Can you give us a plot with a range of contours so we know how big the advantages are deep in the green and deep in the blue?

ajt60gaibb:
Here it is: (If you see the surface at `4` above `(m,n)`, then `ndct` is observed to be x4 faster than `clenshaw` for `m` evaluation points and a degree `n` chebtech.) 

speedup

trefethen:
Beautiful!  Clearly each code is much faster than the other in its preferred territory.

ajt60gaibb:
Note that this PR will reduce the observed computational complexity of `chebfun/roots`.  For a degree `n` chebfun, this PR takes our current `O(n^2)` rootfinder to an `O(n log n)` rootfinder (when seeking only real roots).

trefethen:
Please be very careful, @ajt60gaibb and @nickhale.  `roots` is as close to the heart of Chebfun as anything.

Changing from n^2 to n log(n) sounds very nice, but is there an advantage in practice, and how big, and with what crossover value?

nickhale:
> OK. Here is a comparison of the two approaches. On my computer, I observe that `clenshaw` is faster in the green region and `ndct` is faster in the blue region. The black lines indicate where I put the switch between the two algorithms in `chebtech/feval`:
> 
> breakeven
> 
> (I want to be a little conservative about switching to `ndct`.)

Can you please share the code you used to generate this? (It would be good to compare on a few different machines.)

ajt60gaibb:
```
mm = round(linspace(1, 1e4, 100)); 
nn = round(linspace(1, 1e4, 100));
iter = 5; 
j = 1; k = 1; 
for n = nn 
    j = 1; 
    for m = mm 
        x = 2*rand(m,1)-1;
        c = randn(n,1)./(1:n)'.^2; 
        s1 = tic; 
        for ss = 1:iter
            y1 = chebtech.clenshaw(x, c ); 
        end
        t1(j,k) = toc(s1)/iter;
        s2 = tic; 
        for ss = 1:iter 
            y2 = chebfun.ndct( x, c ); 
        end
        t2(j,k) = toc(s2)/iter;
        j = j + 1; 
    end
    k = k + 1;
    n
end

%%
[mmm, nnn] = meshgrid(mm, nn);
surf(mmm, nnn, t1, 1+0*t1, 'edgealpha', 0, 'facecolor', 'interp'), hold on
surf(mmm, nnn, t2, .5+0*t2, 'edgealpha', 0, 'facecolor', 'interp')
xlabel('Number of evaluation points','fontsize',16)
ylabel('Degree of chebfun','fontsize',16)
view(0, 90)
set(gca,'fontsize',16)
```

nickhale:
Thanks. Here are my results
res1

And here's where nudct is twice as fast
res2

ajt60gaibb:
I think we can go with the switching that we have right now:  

If evaluation points < 4000 and degree of chebfun < 4000, then use Clenshaw; otherwise, use ndct.

ajt60gaibb:
@nickhale, I am happy for this to be merged.

ajt60gaibb:
All tests pass on my machine.

ajt60gaibb:
🍾

Issue #2122, opened 2017-05-01
Overly aggressive vectorization of * in chebop


tobydriscoll:
Sometimes a * really should be a * and not a .*. For example:
```
N = chebop(0,1);
V = operatorBlock.volt(@(x,t) exp(t.*(x-t)),[0,1]);
N.op = @(x,y) diff(y) + y - x.*(1+2*x).*(V*y)-1-2*x;
```
The result is stored incorrectly as
```
>> N.op

ans =

  function_handle with value:

    @(x,y)diff(y)+y-x.*(1+2.*x).*(V.*y)-1-2.*x
```
which throws an error for the Volterra block when using the chebop. 

It's a bit of an edge case, but I don't know another syntax for setting up problems with Volterra or Fredholm operators. Perhaps the vectorizer should try to evaluate the expression before keeping it?

tobydriscoll:
My workaround for this issue led to another issue. 
```
N = chebop(0,1);
N.op = @(x,y) diff(y) + y - x.*(1+2*x).*mtimes(V,y)-1-2*x;
N.lbc = 1;
y = solvebvp(N,0);
```
gave
```
No appropriate method, property, or field 'iszero' for class 'adchebfun'.

Error in adchebfun/subsref (line 1474)
                    out = f.(index(1).subs);

Error in  *  (line 175)
                C.iszero = A.iszero || B.iszero;

Error in chebop.vectorizeOp>@(x,y)diff(y)+y-x.*(1+2.*x).*mtimes(V,y)-1-2.*x

Error in chebop/feval (line 190)
    out = N.op(x, u);

Error in chebop/linearize (line 150)
        Nu = feval(N, x, u{:});

Error in chebop/solvebvp (line 109)
[L, residual, isLinear] = linearize(N, u0, x);
```
Perhaps this is a separate bug. In any case, I don't know how to solve integro-differential problems any longer.

nickhale:
The following works on development:
```
N = chebop(0,1);
K = @(x,t) exp(t.*(x-t));
N.op = @(x,y) diff(y) + y - x.*(1+2*x).*volt(K,y) - 1 - 2*x;
% N.lbc = 1;
N.bc = @(x,y) y(0) - 1;
y = N\0;
plot(y), shg
norm(N.op(y)) + abs(y(0)-1)
```
```
ans =
   7.4962e-14
```

(The boundary condition is imposed in this way because of #2179.)

PR #2180, opened 2017-06-19
Fix-fred-ivp

### LINKED ISSUE #2179: Fredholm and Volterra operators not working for integro-DIFFERENTIAL equations
Note that pure integral operators seem to work, but mixed operators fail with a `treeVar`-related message.

The tests for these operators are very basic. As far as I can tell they only check linear integral operators in `tests/chebop/test_intops.m`. This is probably why this has gone unnoticed.

Sample:
```
K = @(x,y) exp(-(x-y));
A = chebop(@(x,u) diff(u) + fred(K,u));
A.lbc = 1;
A\1
```
Output:
```
Undefined function 'fred' for input arguments of type 'treeVar'.
Error in @(x,u)diff(u)+fred(K,u)

Error in treeVar.toFirstOrder (line 58)
    fevalResult = funIn(t, args{:});
Error in chebop/solveivp (line 130)
        treeVar.toFirstOrder(N.op, rhs, N.domain);
Error in  \  (line 49)
    [varargout{1:nargout}] = solveivp(N, rhs, pref, varargin{:}); 
```

nickhale:
Fix 'IVPs' with Fredholm or Volterra terms.

Closes #2179 and #2122.

### DIFF
--- a/@treeVar/treeVar.m
+++ b/@treeVar/treeVar.m
@@ -290,8 +290,15 @@ classdef  (InferiorClasses = {?chebfun}) treeVar
         end
         
         function f = fred(varargin)
+            %FRED   Not supported.
+            % We don't support integral equations with our first order
+            % reformulation. However, we could accidentally end up here in case
+            % of first order integral equation, where the conditions are
+            % specified via N.LBC/RBC. Throw a meaningful error message in this
+            % case.
             error('CHEBFUN:TREEVAR:FRED:notSupported', ...
-                'Fred is not supported in treeVar.');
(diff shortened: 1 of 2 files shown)


asgeirbirkis:
Thanks @nickhale for the bug fix; can confirm that the approach taken is the ideal one (throw an error with the "notSupported" keyword, which makes the solver silently retry to solve the problem globally rather than via first order reformulation). I added a bit to the error message and documentation to make it match the error thrown by `treeVar/cumsum`, sorry it took me so long to review and merge!

PR #2181, opened 2017-06-23
Feature Chebyshev Convolution

kuanxu:
This introduces a new algorithm for convolution of `chebfuns`, which is fully based on Chebyshev polynomials. It's equally fast as the current algorithm which relies on conversion between Chebyshev and Legendre coefficients. 

The new method also offers a unified framework for approximating convolution operator in coefficient space for all classic orthogonal polynomials, including Chebyshev, Gegenbauer (Legendre as a special case), Jacobi, and Laguerre. It generalizes the work by Hale and Townsend (2014).

Manuscripts describing the method will be furnished soon.

All tests pass.

### DIFF
--- a/@bndfun/conv.m
+++ b/@bndfun/conv.m
@@ -1,4 +1,4 @@
-function h = conv(f, g)
+function h = conv(f, g, pt)
 %CONV   Convolution of BNDFUN objects.
 %   H = CONV(F, G) produces the convolution of BNDFUN objects F and G:
 %                     - 
@@ -17,6 +17,7 @@ function h = conv(f, g)
 % See http://www.chebfun.org/ for Chebfun information.
 %
 % Nick Hale and Alex Townsend, 2014
+% Kuan Xu, 2017
 
 %% %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
 % DEVELOPER NOTE:
@@ -24,6 +25,13 @@ function h = conv(f, g)
 % For further details, see Hale and Townsend, "An algorithm for the convolution
 % of Legendre series", SIAM Journal on Scientific Computing, Vol. 36, No. 3,
 % pages A1207-A1220, 2014.
+% 
+% For convolution based on all classic orthogonal polynomials, including 
+% Chebyshev polynomials, Gegenbauer polynomials, Jacobi polynomials, and 
+% Laguerre polynomials, see
+%
+% [1]. Xu and Loureiro, "Spectral approximation of convolution operator" (2017).
+% [2]. Loureiro and Xu, "convolution of Chebyshev polynomials" (2017).
 % 
 % In the following, it is assumed that the length of the domain of g is greater
 % than the length of the domain of f. If this is not the case, then simply
@@ -43,7 +51,7 @@ function h = conv(f, g)
 % The triangular pieces at end are dealt with using a convolution theorem for
 % Legendre polynomials, which leads to a convenient recurrence relation. The
 % cost of this is O(m*n) and results in a polynomial of degree m+n. See
-% EASYCONV() for details.
+% legConv() for details.
 %
 % Similarly one can show the interior rectangle results in a polynomial of
 % degree n. This can be computed by patching with R = floor[(d-c) / (b-a)]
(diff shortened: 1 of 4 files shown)


kuanxu:
The current code provides options for different methods by an extra flag - `cheb` for the new algorithm and `leg` for the old Legendre-based algorithm. Here is an example.

```
>> f = chebfun(@(x) cos(2*x), [-1 1])
f =
[      -1,       1]       19     -0.42    -0.42 
vertical scale =   1 
>> g = chebfun(@(x) sin(5*x), [2 4])
g =
[       2,       4]       26     -0.54     0.91 
vertical scale =   1 
>> hc = conv(f, g, 'cheb')
hc =
[       1,       3]       45  -2.6e-17    0.092 
[       3,       5]       45     0.092  3.3e-17 
vertical scale = 0.33    Total length = 90
>> hl = conv(f, g, 'leg')
hl =
[       1,       3]       25  -1.6e-17    0.092 
[       3,       5]       26     0.092  3.5e-17 
vertical scale = 0.33    Total length = 51
>> norm(hc-hl, inf)
ans =
     3.184773059124818e-16
```

ajt60gaibb:
Sounds good, @kuanxu!  I would really like to know how it works and would like to read the manuscript when you are ready?  Shall we leave the PR until there's a manuscript?

MikaelSlevinsky:
I'd like to know too!

kuanxu:
@ajt60gaibb Thanks. Hopefully, it won't take long before the first draft is finished. 

@MikaelSlevinsky I guess I'll see you in Canterbury next week?

ajt60gaibb:
@kuanxu, I'm in Glasgow.  I assume your talk is on this stuff. I will attend.

kuanxu:
@ajt60gaibb Yes, just arrived. See you folks tomorrow!

ajt60gaibb:
👍

MikaelSlevinsky:
Yes @kuanxu, I present after you next Monday!

kuanxu:
@MikaelSlevinsky Yes, I was told that our session has been moved to Monday. Not bad to have the talk done on the first day.

ajt60gaibb:
This is nice work that realizes that the derivation of the 4-term recurrence for Bleft in Hale and Townsend (2014) does not *need* to work in Fourier-space with spherical Bessel functions.  While Chebyshev-based convolution avoids the `cheb2leg` transforms, the resulting Bleft matrix has a more complicated structure. One expects that they have a comparable computational cost. However, the new Chebyshev-based code seems to be a slower than the old Legendre-based code when the inputs are of high polynomial degree. For example, 

```
f = chebfun(@(x) cos(6000*x), [-1 1]);
g = chebfun(@(x) sin(8000*x), [2 4]);
tic, 
hc = conv(f, g, 'cheb');
toc
Elapsed time is 4.448127 seconds.
tic
hc = conv(f, g, 'leg');
toc
Elapsed time is 1.404993 seconds.
```
Any possible speed-ups in the code, Kuan?

kuanxu:
@ajt60gaibb Thanks for pointing this out. Your experiment result is as expected. When the lengths of `f` and `g` are close enough like in this example, the Chebyshev convolution matrix doesn't have the 'banded' part, but only the sub-diagonal columns and the top rows. Therefore all entries are computed via recurrence, which is the most expensive part of the entire algorithm. But Legendre convolution matrix is always symmetric, which means that only half of the entries are calculated via recurrence. So half of the cost is saved. Your experiment above didn't rule out the overhead. Here is a slightly more careful test:
```
f = chebfun(@(x) cos(6000*x), [-1 1]);
g = chebfun(@(x) sin(8000*x), [2 4]);

conv(f, g, 'cheb');
conv(f, g, 'leg');
tc = 0;
tl = 0;
for i = 1:10
    tic
    conv(f, g, 'cheb');
    tc = tc + toc;
    
    tic
    conv(f, g, 'leg');
    tl = tl + toc;
end
[tc/10 tl/10]
```
The second number is almost the half of the first, as expected.
```
ans =
   4.053269800000000   1.834876800000000
```
However, when you do convolution of two `chebfuns` with the length of one much larger than that of the other, `chebconv` takes pretty much the same time as `legconv`, or even becomes slightly quicker.
```
f = chebfun(@(x) cos(x), [-1 1]);
g = chebfun(@(x) sin(9000*x), [2 4]);

conv(f, g, 'cheb');
conv(f, g, 'leg');
tc = 0;
tl = 0;
for i = 1:10
    tic
    conv(f, g, 'cheb');
    tc = tc + toc;
    
    tic
    conv(f, g, 'leg');
    tl = tl + toc;
end
[tc/10 tl/10]
```
gives
```
ans =
   0.133913400000000   0.156065500000000
```
What we learned, I think, is `leg2cheb` and `cheb2leg` are **really** fast and its cost is totally negligible in the current context and `fft` routines are unbeatable, especially in Matlab. 

I've tried my best when coding `chebconv`. I think the room for speed-up in terms of coding is quite little. But I trust that it won't take another 5 years for some of us to improve it to something like O(nlogn) or O(n(logn)^2).

So, I suggest that we set `legconv` as default for the time being. How would you say?

ajt60gaibb:
I think we should close this (but keep the branch). At the moment, the elegant chebyshev approach is more computationally expensive than the Legendre approach.

PR #2185, opened 2017-06-30
Fix-trigcolloc-fredvolt

nickhale:
Support for Fredholm operators in `trigcolloc` discretisation.

Volterra operators are complicated for the same reason that the `cumsum` (indefinite integral operator) is complicated for trig-discretisations, i.e., that unless you enforce something like a zero-mean condition, then the indefinite integral of a `trigtech` is not necessarily a `trigtrech`.


Hadrien-Montanelli:
@nickhale - could you please add a test?

Hadrien-Montanelli:
(Otherwise happy to merge it.)

Hadrien-Montanelli:
Any update @nickhale?

nickhale:
Done.

Hadrien-Montanelli:
Thanks @nickhale.

PR #2189, opened 2017-07-13
feature-ultra-to-ultra

nickhale:
Ultraspherical to ultraspherical conversion.

This is essentially just a wrapper for `jac2jac`, but the scalings can be tricky to compute (particularly for large n) so I think it's worthwhile having such a method. (I had my own version and I've been using it a lot.)

I also fixed two small bugs in `jac2jac` that occur when `n = 1` or `n = 2`.


ajt60gaibb:
Add a test?

PR #2190, opened 2017-07-14
Feature-vals2coeffs-1stkind

nickhale:
Modified vals2coeffs code to work with either first- or second-kind Chebyshev grids.
Added a test for all of these codes (which was totally lacking?).
Add chebvals2chebvals for converting between Chebyshev grids.


ajt60gaibb:
Very useful pull request.

PR #2194, opened 2017-07-21
Feature chebop system vector bc

asgeirbirkis:
Following a suggestion by @trefethen , this branch enables specifying boundary conditions for first order coupled systems as a vector. As an example, the SIR model (one of the chebgui demos) can now be specified and solved as follows:
```
>> N = chebop(@(t,u,v,w) [diff(u)-(-0.003.*v).*u; 
        diff(v)-0.003.*v.*u+0.3.*v; 
        diff(w)-0.3.*v], [0 30]);
>> N.lbc = [500; 1; 0]; % Old syntax @(u, v, w) [u-500; v-1; w];
>> N
N =
   Nonlinear operator:
      u,v,w |--> diff(u)-(-0.003.*v).*u
                 diff(v)-0.003.*v.*u+0.3.*v
                 diff(w)-0.3.*v
      [0,30]
   with
    left boundary condition(s):
      u = 500
      v = 1
      w = 0
>> [u, v, w] = N\0;
>> plot([u,v,w])
```
sirplot


nickhale:
This code looks good to me (and a very tidy syntax). My only concern is that no error is thrown in the not-allowed case of higher-order systems. Is there anything that can be done about this?

I also think the test could be expanded a little. In particular, testing the case when the conditions are given in `.rbc` rather than `.lbc`.

On another note, I see that calling `N\0` with no outputs throws an error. I propose that this should plot the result in the same way as MATLAB's ODE solvers, e.g., `ode45(@(t,y) vdp1(t,y), [0 10], [0 1])`. In our case, we could even automatically add axis labels and a legend.

asgeirbirkis:
Thanks @nickhale for looking at the code and your comments. Regarding your questions:

* The non-allowed case requires `diffOrder` information in order for us to be able to catch it. Doing that via linearisation is rather expensive, but I've realised that `treeVar` should be able to provide us with that information cheaply, but it requires a bit of rewriting for that class. However, since a meaningful error does get thrown (134-139 in `parseBC`) when an attempt is made to solve the system, may I suggest we defer implementing further checks to a second PR which I should be able to issue rather soon?

* I've added a new test 096d8a0, assigning `.rbc` rather than `.lbc`. I agree that it would be desirable having even more tests (although @trefethen has tried this for around 20 examples I believe and found it always to work, which is a good sign at least).

* Weird, I've never noticed this error of calling backslash without an output! I'm happy to go with the Matlab like behaviour, should we open up a separate issue for this and if others agree, I could issue a PR for it?

asgeirbirkis:
Thanks @nickhale for the fix in 7163750 and for merging this!

Issue #2199, opened 2017-08-09
The Newton method in Lagpts needs improving


ajt60gaibb:
@popsomer, any idea? 

```
[x,w,v] = lagpts(5001);
Error using lagpts>alg_rh (line 223)
Newton method may not have converged.
Error in lagpts (line 118)
    [x, w] = alg_rh(n, strcmpi(method,
    'RHW'), alpha);  % Nodes and weights
```

sslchi:
I found that for `alpha = 3, n = 400`, Newton method may not have converged.
```
[x, w, v] = lagpts(400,3);
Error using lagpts>alg_rh (line 223)
Newton method may not have converged.

Error in lagpts (line 118)
    [x, w] = alg_rh(n, strcmpi(method, 'RHW'), alpha);  % Nodes and weights
 ```

popsomer:
The error can be avoided by increasing the tolerances in lagpts, but there will always be an n above which the Newton step remains above the tolerance for the given maximal number of iterations. A better solution would be to use the explicit expansions in my pull request #2200 which avoid Newton iterations. Do you agree, @ajt60gaibb @sslchi ?

PR #2200, opened 2017-08-11
Explicit expansions in lagpts, used in hermpts

popsomer:
The  default algorithm for low n in lagpts is now forward recurrence because the weights in Golub-Welsh become exactly zero long before they should underflow, as in issue 52 on FastGaussQuadrature.jl.

In response to #2199 here, the tolerances in the Newton method were increased, but even for higher tolerances, there will always be an n above which the Newton step remains above the tolerance for the given maximal number of iterations.

The included explicit expansions avoid such problems with Newton iterations on asymptotic expansions and they are extremely fast, such that it is now the default algorithm for high n. Higher-order asymptotics were not implemented in the Airy region (currently only  O(n^{-4}) for the nodes and O(n^{-2/3}) for the weights) as the latter easily underflow, like the subsampled quadrature rule in reference [3] of hermpts.

METHOD = 'LAG' was added to hermpts. It uses a relation with Gauss-Laguerre for alpha^2 = 1/4 which can easily be proven. The user can further exploit this by using the corresponding code in hermpts, letting lagpts only compute nonunderflowing weights with associated nodes, and only storing those.

### DIFF
--- a/hermpts.m
+++ b/hermpts.m
@@ -11,12 +11,14 @@ function [x, w, v] = hermpts(n, varargin)
 %   quadrature. [X,W,V] = HERMPTS(N) returns in addition a column vector V of
 %   the barycentric weights corresponding to X.
 %
-%   [X, W] = HERMPTS(N, METHOD) where METHOD is one of 'GW', 'REC', 'GLR', or 
+%   [X, W] = HERMPTS(N, METHOD) where METHOD is one of 'GW', 'REC', 'GLR', 'LAG' or 
 %   'ASY' allows the user to select which method is used. 'GW' will use the 
 %   traditional Golub-Welsch eigenvalue method [1], best when n<=20. 'REC' 
 %   uses Newton's method with polynomial evaluation via the 3-term 
 %   recurrence for Hermite polynomials. 'GLR' uses Glaser-Liu-Rokhlin 
-%   fast algorithm which is much faster for large N [2]. 'ASY' uses Newton's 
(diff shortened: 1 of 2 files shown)


ajt60gaibb:
I am getting tonnes of failures from the `lagpts` command now.  What went wrong?  Do you see that too?      Also, can you add things in the tests the examples in #2199. 

```
The following tests failed or crashed:
   misc/test_lagpts.m
>> test_lagpts
ans =
  1×14 logical array
  Columns 1 through 9
   1   1   0   1   1   1   1   1   0
  Columns 10 through 14
   0   1   0   0   0
```

popsomer:
The default algorithm is now Golub-Welsh again for low n, but where the small and inaccurate weights are replaced by the ones from the recurrence relation.

The explicit expansions of the nodes were fine for low alpha, but the weights seem to need more terms. So, they are used only for n >= 1024 and between the limits, the RH algorithm is applied. When that convergence error occurs, we switch to the Glaser-Liu-Rokhlin fast algorithm for alpha=0 and to the explicit expansions for other alpha, although they may be inaccurate. 

The situation for n>=128 is a temporary fix: I will try to derive more terms for the weights such that they become accurate enough for 128 <= n < 1024 as well.

The requested tests are added and pass, as well as tests for alpha=0.7 for the three regimes for n with reasonable tolerances.

popsomer:
By increasing the number of terms in the explicit expansions, the tests now also pass when using them from n=128 (default algoritm).

PR #2203, opened 2017-08-25
A comment for baryWeights

sslchi:
A comment for baryWeights.

### DIFF
--- a/baryWeights.m
+++ b/baryWeights.m
@@ -29,7 +29,7 @@ end
 
 % Compute the weights:
 if ( (n < 2001) )              % For small n using matrices is faster.
-   V = C*bsxfun(@minus, x, x.');
+   V = C*bsxfun(@minus, x, x.'); % this is not consistent with chebpts and n >= 2001 (for even number)
    V(1:n+1:end) = 1;
    VV = exp(sum(log(abs(V))));
    w = 1./(prod(sign(V)).*VV).';


ajt60gaibb:
Can you explain your comment a little more?  What is inconsistent with `n>=2001`, `n = even` and why does it matter?

aaustin141:
I'm also confused:
```
>> [~, ~, vk] = chebpts(2048);
>> vk(end-9:end)
ans =
  -1.000000000000000
   1.000000000000000
  -1.000000000000000
   1.000000000000000
  -1.000000000000000
   1.000000000000000
  -1.000000000000000
   1.000000000000000
  -1.000000000000000
   0.500000000000000
>> vk = baryWeights(chebpts(2048));
>> vk(end-9:end)
ans =
  -0.999999999981396
   0.999999999985290
  -0.999999999988070
   0.999999999988457
  -0.999999999993303
   0.999999999992671
  -0.999999999999554
   0.999999999989409
  -0.999999999968535
   0.499999999977817
```
I don't see any inconsistency here, only differences due to rounding error...

ajt60gaibb:
@sslchi, we do expect these rounding errors in `baryWeights`.  I wondered if it is the sign of the first barycentric weight that is inconsistent somewhere, but like @aaustin141 I can not find it.

sslchi:
@ajt60gaibb , I am sorry, It should be `n < 2001, n = even`. In `baryWeights`, when `n > 2001, n = even`, the  first barycentric weight is negative,  when `n < 2001, n = even`, the  first barycentric weight is positive. For `chebpts`, the last barycentric weight is positive.
```
>> [~, ~, vk] = chebpts(4);
>> vk

vk =

   -0.5000
    1.0000
   -1.0000
    0.5000

>> vk = baryWeights(chebpts(4));
>> vk

vk =

    0.5000
   -1.0000
    1.0000
   -0.5000

```
Though the sign of barycentric weights make no difference for interpolation, I think it may be better to make them have the same sign.

ajt60gaibb:
Good spot! They should be the same sign.  Do you want to make them the signs consistent on this pull request, @sslchi?

sslchi:
Yes, we can discuss this problem at #2205 .

PR #2204, opened 2017-08-28
Support for CHEBMATRIX syntax for IVPs.

asgeirbirkis:
Syntax for indexed variables, such as
    N.op = @(x,u) [diff(u{1},2) + u{2}; sin(u{1}) + diff(u{2})]
has long been supported for BVPs. This was missing for IVPs, which
this PR aims to rectify.

### DIFF
--- a/@chebop/solveivp.m
+++ b/@chebop/solveivp.m
@@ -119,6 +119,18 @@ if ( strcmp(solver, 'values') || strcmp(solver, 'coeffs') || ...
     return
 end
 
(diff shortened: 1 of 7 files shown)


trefethen:
This runs successfully with 99 examples in Appendix B of _Exploring ODEs_.  Since I'm finishing the book today and I need this in development for one of the examples, I'm going to merge this without looking at the code.  Thanks for the instant fix @asgeirbirkis.

PR #2207, opened 2017-08-30
add test_baryWeights.m

sslchi:
`test_baryWeights.m` only check the consistency of  signs of barycentric weights.
Now, for all `*pts`, the first barycentric weight is positive.

There is a big error between `lagpts` and `baryWeights`, the following is the test codes:

For `n = 4`,
```
>> [x, ~, v] = lagpts(4);
>> norm(baryWeights(x) - v)
ans =
   8.6988e-16
```
but for  `n = 5`
```
>> [x, ~, v] = lagpts(5);
>> x
x =
    0.2636
    1.4134
    3.5964
    7.0858
   11.6751
>> v
v =
    0.4940
   -1.0000
    0.6962
   -0.2131
    0.0207
>> baryWeights(x)
ans =
    0.4897
   -1.0000
    0.7124
   -0.2358
    0.0337
>> norm(baryWeights(x) - v)
ans =
    0.0310
```

### DIFF
--- a/@chebfun/chebfun.m
+++ b/@chebfun/chebfun.m
@@ -385,6 +385,12 @@ classdef chebfun
         
         % ODE45 with CHEBFUN output.
         [t, y] = ode45(varargin);
+
+        % ODE78 with CHEBFUN output.
+        [t, y] = ode78(varargin);
+
+        % ODE89 with CHEBFUN output.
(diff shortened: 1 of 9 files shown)


ajt60gaibb:
The bug in the comment is also here: #2208.

PR #2210, opened 2017-09-10
Fix tree var sum

asgeirbirkis:
This is to fix an issue originally raised on the Google group, https://groups.google.com/forum/#!topic/chebfun-users/0dpsggp1RyA .

`treeVar` was missing a `sum` method, which has now been added to work analogous to the `cumsum` method of the `treeVar` class (throw a meaningful error, and try switching automatically to global methods, rather than time-stepping).


asgeirbirkis:
Thanks @nickhale for the merge!

PR #2213, opened 2017-09-11
Enforce that the constructor return a chebfun column when the input is a chebfun row

### LINKED ISSUE #2209: Mixed-tech row-times-column product runs away when the row is based on trigtech.
The following computations all work just fine:
```
>> f = chebfun(@(x) sin(pi*x), 'tech', @trigtech);
>> g = chebfun(@(x) cos(pi*x));
>> f'*f
ans =
     1
>> g'*g
ans =
   1.000000000000000
>> g'*f
ans =
     7.409082265771725e-18
```
But if I do `f'*g`, the computation seems to hang.

This is with current development (7dec37a3) on MATLAB R2016b.

gradywright:
This was causing the constructor to hang since evaluation with a column of values would return a single row of output values.  Fixes #2209.

### DIFF
--- a/@chebfun/chebfun.m
+++ b/@chebfun/chebfun.m
@@ -779,6 +779,9 @@ function [op, dom, data, pref, flags] = parseInputs(op, varargin)
             op = vectorCheck(op, dom, vectorize);
         end
         if ( isa(op, 'chebfun') )
+            if ( op.isTransposed )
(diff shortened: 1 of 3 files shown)


sfilip:
Thanks @gradywright ! I've taken a look at this and am happy to merge.

PR #2214, opened 2017-09-19
Feature fast poisson

ajt60gaibb:
An optimal complexity Poisson solver on a rectangular domain based on ultraspherical polynomials and the alternating direction implicit (ADI) method. I have wanted an optimal Poisson solver for a ***long*** time. 

The code computes the `mxn` bivariate Chebyshev expansion coefficients of the solution to 
```
lap(u) = f,  in [a,b],         u = g on boundary
```
in a total of `O( mn*(log(n) + log(m))*log(1/eps) )` operations. Here, `eps` is the tolerance of the final solution and is set to `chebfun2eps` by default. 

To see the computational benefit over chebop2 (which is running an `O(m^3+n^3)` algorithm), try: 

```
n = 2049; 
N = chebop2( @(u) lap(u) );
N.bc = 0;
f = chebfun2(@(x,y) 1+0*x);

tic, u = chebfun2.poisson(f, 0, n, n); toc
Elapsed time is 21.609899 seconds.

tic, v = solvepde(N, f, n, n); toc
Elapsed time is 126.888612 seconds.

norm(u-v)     % Chebfun2 constructor rounds this to zero. 
ans =
     0
```

This is joint work with @danfortunato. A publication that contains fast Poisson solvers for rectangles, cylinders, solid sphere, and (perhaps) cubes is in preparation. The publication will contain C code for the inner ADI iteration, making `chebfun2.poisson` *much* faster. The publication is nearly finished.

~A subsequent pull request will use the automatic differentiation in `chebop2` to automatically execute `chebfun2.poisson` when Poisson's equation is detected.~  We may also write code for low-frequency Helmholtz.


danfortunato:
The fast Poisson solver is now integrated into `chebop2`. In particular, `chebop2` now detects Poisson's equation and calls `chebfun2.poisson` automatically. Additionally, `chebfun2.poisson` uses the adaptive discretization capability of `chebop2` to automatically solve Poisson's equation when the user does not supply a discretization size:
```
v = chebfun2( @(x,y) (1-x.^2).*(1-y.^2).*cos(2*pi*x.*y) );
f = lap(v);
u = chebfun2.poisson(f); % defaults to zero Dirichlet data
norm(u-v)
ans =
     0

v = chebfun2( @(x,y) (1-x.^2).*(1-y.^2).*cos(2*pi*x.*y)+1 );
f = lap(v);
u = chebfun2.poisson(f, 1);
norm(u-v)
ans =
     0
```

ajt60gaibb:
A draft of the manuscript accompanying this PR is here: 
[optimalPoisson.pdf](https://github.com/chebfun/chebfun/files/1410217/optimalPoisson.pdf)

gradywright:
@ajt60gaibb and @danfortunato The new code looks good.  However, some of the changes you made now cause nearly half the tests in `chebop2` fail.  The problems seem to stem from a bug in`chebop2/checkBC`.

gradywright:
The `test_domain` test is the only one that fails for me now.  After some investigation, it appears that the automatic discretization method does not determine a high enough grid resolution to compute a solution of some Poisson equations to within the tolerances specified.  Here is an example (parsed out from `test_domain`):
```
d = [-2 2 -2 2];
g = chebfun2(@(x,y) (2-x).*(2+x).*((y+2)/4),d);
f = chebfun2(0,d);
u = chebfun2.poisson(f,g);
norm( u(:,d(3)) - g(:,d(3)) )
norm( u(:,d(4)) - g(:,d(4)) )
norm( u(d(1),:) - g(d(1),:) )
norm( u(d(2),:) - g(d(2),:) )
ans =
     6.882748627692141e-12
ans =
     2.827460542331166e-10
ans =
     1.952311194297190e-10
ans =
     1.952311194296997e-10
```
Here the grid size is determined to be 129-by-129.  If instead you supply `poisson` with a larger grid size, then the errors are reduced to within the original tolerances of `test_domain`.
```
u = chebfun2.poisson(f,g,513);
norm( u(:,d(3)) - g(:,d(3)) )
norm( u(:,d(4)) - g(:,d(4)) )
norm( u(d(1),:) - g(d(1),:) )
norm( u(d(2),:) - g(d(2),:) )
ans =
     8.067473530467818e-16
ans =
     6.043399886646788e-14
ans =
     5.457729975502695e-14
ans =
     5.457729975502695e-14
```
An easy solution is just to lower the tolerances in `test_domain` so that the test passes, but you might want to revisit the automatic discretization code in `chebop2` to make sure it's working correctly.  Errors on the order of 1e-10 just seem a bit too large for this problem.

danfortunato:
Thanks for finding this, @gradywright! I have added a check for the resolution of the boundary conditions in the adaptive discretization code. After slightly modifying a couple tolerances, all tests now pass for me.

Issue #1936, opened 2016-07-01
besseli calculation has a missing minus sign


rajb245:
The chebfun function besseli uses an incorrect formula. There is a minus sign missing from the relationship. Test code comparing Chebfun against matlab's built-in:

```
x = chebfun('x');
y = besseli(1,x);
figure, plot(x,real(y),x,imag(y));
%that besseli(1,x) returns small imaginary parts is
%another issue, but it's plotted for completeness
xx = linspace(-1,1,1000);
hold on;
plot(xx,besseli(1,xx),'r');
hold off;
legend('Chebfun Re', 'Chebfun Im', 'Matlab Builtin');
```

The correct formula uses a scale factor of 1i^-nu, not 1i^nu. [I trust Wikipedia](https://en.wikipedia.org/wiki/Bessel_function#Modified_Bessel_functions:_I.CE.B1_.2C_K.CE.B1) but there's this [NIST formula](http://dlmf.nist.gov/10.27#E6) that simplifies to the same thing. Patch for the code and documentation is below.

```
--- a/@chebfun/besseli.m
+++ b/@chebfun/besseli.m
@@ -11,14 +11,14 @@ function g = besseli(nu, f, varargin)
 %
 %   The relationship between I_NU(Z) and J_NU(Z) = BESSELJ(NU, Z) is
 %
-%        I_NU(Z) = 1i^NU * J_NU(1i*Z).
+%        I_NU(Z) = 1i^-NU * J_NU(1i*Z).
 %
 % See also AIRY, BESSELH, BESSLJ, BESSELK, BESSELY.

 % Copyright 2016 by The University of Oxford and The Chebfun Developers.
 % See http://www.chebfun.org/ for Chebfun information.

-scl = 1i.^nu;
+scl = 1i.^-nu;
 g = scl*besselj(nu, 1i*f, varargin{:});

 end

```

rajb245:
Closed by https://github.com/chebfun/chebfun/pull/2215 (with no attribution for those two minus signs!)

PR #2215, opened 2017-09-22
small bug in besseli computation

sfilip:
This pull request solves #1936 by correcting a small bug in the computation of besseli functions.


trefethen:
Just two minus signs.

PR #2216, opened 2017-09-25
code for outputting the poles of the best approximation

sfilip:
This pull request adds the following feature: when calling `minimax`, the poles of the best approximation are now computed and stored in the output argument `status.pol`.


trefethen:
Thanks @sfilip.  I fixed a couple of typos and now I will be happy to merge.

Issue #2217, opened 2017-09-28
Stop AAA failing when a data value is infinite


trefethen:
It's awkward that `aaa` fails if any data values are infinity (it tries to take the SVD of a matrix with an infinite entry).  For example:
```
>> X = linspace(-1,1); aaa(gamma(X),X)
Error using svd
Input to SVD must not contain NaN or Inf.

Error in aaa (line 84)
    [~, ~, V] = svd(A(J,:), 0);         % Reduced SVD.
 ```
Could we put in the easy minimal fix -- delete any entries from X and F corresponding to infinite values of F?  @sfilip or @nakatsukasayuji, perhaps you could do this?

PR #2218, opened 2017-09-29
Fix for aaa failing with infinite data values

sfilip:
This pull request offers a quick fix to `aaa` failing when infinite data values are supplied by the user (see issue #2217). Following the suggestion of @trefethen, it simply removes infinite values from F(Z) and the corresponding entries from Z.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -55,6 +55,12 @@ if ( needZ )
     return
 end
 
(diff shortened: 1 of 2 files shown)


trefethen:
We added NaN as well as Inf, to take care of things like sin(x)/x.  Looks good.  Will merge.

PR #2219, opened 2017-10-12
Fix randnfun normalization

trefethen:
Aurya Javeed of Cornell's Center for Applied Math found that `randnfun` had the wrong normalization by a factor of sqrt(2) in `'norm'` mode.  I think I have now fixed this.

### DIFF
--- a/randnfun.m
+++ b/randnfun.m
@@ -11,7 +11,7 @@ function f = randnfun(varargin)
 %   RANDNFUN(LAMBDA, N) returns a quasimatrix with N independent columns.
 %
 %   RANDNFUN(LAMBDA, 'norm') normalizes the output by dividing it by about
-%   2*SQRT(LAMBDA), so white noise is approached in the limit LAMBDA -> 0,
+%   SQRT(LAMBDA/2), so white noise is approached in the limit LAMBDA -> 0,
 %   with an indefinite integral corresponding to standard Brownian motion.
 %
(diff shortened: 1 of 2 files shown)


aj463:
Pretty insignificant, but I think the comment on Line 14 means to say SQRT(LAMBDA/2). (See lines 68 and 70.) 

More verbosely, without the 'norm' flag, the output of randnfun in 1D is

F(t) = A_0 + SQRT(2) SUM_{k = 1}^m A_k COS(2*PI k t/L) + B_k SIN(2*PI k t/L),   (*)

where A_k and B_k ~ N(0, 1/(2m + 1)) (Equation (12.9) in "Exploring ODEs").

On the other hand, Kahane says F(tL) approximates the derivative of Brownian motion on [0, 1] when A_k and B_k ~ N(0, 1).

Therefore 'norm' divides (*) by SQRT(L/(2m + 1)) to scale A_k and B_k to N(0, 1) (and also to scale the integral of (*) into Brownian motion over [0, L]). Since LAMBDA is approximately L/m, we have that the divisor SQRT(L/(2m + 1)) is about SQRT(LAMBDA/2).

trefethen:
Thanks Aurya.  I agree and I have corrected that line of the help text.

sfilip:
I'm happy to merge this. Thanks @aj463 and @trefethen!

PR #2222, opened 2017-10-17
fix-fracdiff-quasi

nickhale:
Add quasimatrix support for fractional derivatives.


Hadrien-Montanelli:
@nickhale can you please add a test? Happy to merge it otw

PR #2225, opened 2017-10-23
norm -> big in randnfun

trefethen:
I've changed the flag `'norm'` (which is confusing -- is it the big version of a random function that we call normalized, or the small one?) -- to `'big'`.  Thus now:

randnfun(lambda):  a function of size O(1)

randnfun(lambda,'big'): a function of size O(1/sqrt(lambda)), suitable for integration for Brownian motion and other stochastic effects


sfilip:
Thanks @trefethen! Happy to merge this.

PR #2227, opened 2017-11-01
Change 'norm'  to 'big' in randnfun2

sfilip:
This change makes the `randnfun2` code consistent with `randnfun` (see pull request #2225).


trefethen:
Thanks @sfilip.  This looks fine and I will merge it.  It's likely that our normalization of 'big' in 2D isn't quite right, actually, and further thought about 2D stochastic effects is needed.  But the purpose of this branch is merely to change the word 'norm' to the word 'big'.

Issue #2230, opened 2017-11-22
minimax when input is 'easy to approximate' complains pmin undefined


NicolasBoumal:
Hello,

Running the following code
`n = 11;

f = @(t) cos(2*pi*t);

a = -1;
b = 1;

fc = chebfun(f, [a, b]);

pn = minimax(fc, n);`

Gets me the following error:
`Undefined function or variable 'pmin'.

Error in minimax>minimaxKernel (line 601)
    p = pmin;

Error in minimax (line 128)
        [p,err,status] = minimaxKernel(f, fHandle,m, n, N, rationalMode,...

Error in bugreport (line 10)
pn = minimax(fc, n);`

I suspect it is because the while-loop starting at line 518 in minimax.m doesn't execute any iteration (because the error criterion is immediately satisfied), so that it gets to line 601 (p = pmin;) without having ever defined pmin.

(I'm using chebfun 5.7.0 and Matlab 2016b.)

Thanks,
Nicolas

sfilip:
Hi Nicolas,

Thanks for catching this. You are right. The issue stems from a hard coded value of the leveled error variable `h`, which prevented any iteration of the algorithm to be executed in this case. I made a pull request #2231 that fixes it. 

Silviu

NicolasBoumal:
Thanks a lot for your quick fix @sfilip, it works for me.
Best wishes,
Nicolas

PR #2231, opened 2017-11-22
fix buggy initialization of the levelled error

sfilip:
This pull request fixes issue #2230 . It modifies a hard coded value of the leveled error variable `h` that prevented the code from executing correctly in certain cases.


nakatsukasayuji:
Thank you Nicolas for the report and @sfilip for the fix! Looks great.

NicolasBoumal:
Thank you both for your (very!) prompt response, and for the toolbox all around. The fix works for me.
Best wishes,
Nicolas

PR #2232, opened 2017-11-25
Fix truncation level in randnspherefun

gradywright:
There was a mistake in how we were relating the truncation level of a random spherical harmonic series to the wave-length of the maximum resolved wave.  The primary purpose of this PR is to fix this mistake.  There was an additional bug where the sampling of the random spherical harmonic series at grid of values was not quite large enough to guarantee that we recovered the highest degree sine mode.  This also fixed with this PR.  Finally, the help text was updated.

### DIFF
--- a/randnfun.m
+++ b/randnfun.m
@@ -26,7 +26,7 @@ function f = randnfun(varargin)
 %   N, if present, is preceded by an explicit specification of LAMBDA.
 %
 %   Reference: S. Filip, A. Javeed, and L. N. Trefethen, "Smooth random
-%   functoins, random ODEs, and Gaussian processes," manuscript, Nov. 2017.
+%   functions, random ODEs, and Gaussian processes," manuscript, Nov. 2017.
 %
 % Examples:
 %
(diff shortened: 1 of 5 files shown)


trefethen:
This is great; thanks @gradywright.  I will merge this.  Along the way I've fixed test_randnfunsphere and made three adjustments in the comment lines:

(1) I've removed mention of the equator.  A random function on the sphere is isotropic (at least I hope so!); the equator has no special status.  Incidentally, an easy argument to see that m ~ 2pi/lambda is to note that the circumference of the sphere is L=2pi and we take m ~ L/lambda in the case of randnfun.  And just as sin(mx) has Laplace eigenvalue -m^2, the corresponding spherical harmonic has Laplace eigenvalue -m(m+1).

(2) I've changed "random smooth function" to "smooth random function" in randnfun, randnfun2, randnfunsphere and randnfundisk.  This matches the terminology in _Exploring ODEs_ and in our new paper by Filip, Javeed, and Trefethen.  

(3) I've included a reference to this paper in the randnfun help text.

PR #2233, opened 2017-11-27
Gaussian process regression command for 1D data fitting

sfilip:
This pull request introduces a basic command ```gpr``` for Gaussian process regression in 1D using a Gaussian (square exponential) covariance function (kernel) and its periodic extension. 

For instance,
```
        rng(0),n = 10; x = -2 + 4*rand(n,1);
        y = sin(exp(x));
        MS = 'markersize'; LW = 'linewidth';
        [f,fvar,smpl] = gpr(x,y,'domain',[-2,2],'samples',3);
        plot(f,LW,3), hold on
        plot(smpl,'color',[.8 .8 .8],LW,2), plot(x,y,'.k',MS,14), hold off
```
gpr

computes a Gaussian process fit to the data pairs ```(x,y)```. The blue curve ```f``` is a ```chebfun``` estimate of the posterior mean, essentially the linear interpolant for ```(x,y)``` with basis functions consisting of the Gaussian kernel centered at each element of ```x```. The grey lines correspond to ```chebfun``` estimates of samples from the Gaussian process posterior; they also pass through ```(x,y)```.

Further comments and references are provided in the help text.


trefethen:
I've spent a bit of time with this -- thanks @sfilip! -- and here are some comments.    I've changed the comments but not the code, and I've added some tests to test_gpr.

(1) SIGMAF seems a complicated name.  Could it be SIGMA?

(2) The comments say that the default variance is sigmaf = 1.  This doesn't sound scale-invariant; also a test I added for scale-invariance fails.  Could everything be made scale-invariant?

(3) Rather than ` 'hyperparams', [sigma,L] ` could we just have ` 'sigma', sigma ` and ` 'L', L ` ?

(4) The comments say that one can have noisy data, but they don't explain (or point to an explanation of) what fit is then returned.

(5) In the help text, there should also be an example involving noisy data.

(6) The test file needs to test the `'trig'` option.

sfilip:
Thanks @trefethen! I've tried to address all your feedback.

(1) SIGMAF is now replaced with SIGMA.

(2) changed the default variance to something scale-invariant: SIGMA= max(abs(Y)).

(3) the hyperparameters can be given separately now, as per your suggestion; also added tests to check this.

(4) added short explanation in the help text of what changes when noisy data is used.

(5) augmented the existing example with a case where the data is perturbed.

(6) added tests for the ```trig``` options and tests for its scale-invariance.

trefethen:
@sfilip, thanks.  I've tweaked a couple of lines in the help text but otherwise left it alone.  However, tests 15 and 16 fail on my machine.  I didn't investigate -- would you see how they do on your machine?  If they pass for you, let me know and I will investigate on mine.

sfilip:
Thanks @trefethen. I ran the tests on two machines with different MATLAB versions. Tests pass on both.

trefethen:
@sfilip, I've taken a look at `test_gpr` and I don't think it's reliable.  It does random things with random numbers.  To put it under better control I've put `rng(2)` before each block that calls `randn`.  I expect you too will now find that some tests fail.  On the other hand if you change `rng(2)` to `rng(1)` they pass.

What's going on, it seems to me, is that the tests are not well designed.  They are "too random".  If you do data fitting with abscissas as random as these, sometimes you get fine results and other times you get awful results.  That's not an appropriate regime for testing.

So could you please try to revise the tests to make them more appropriate and really test the behavior that you think `gpr` is supposed to reliably exhibit?  I didn't look carefully, but it seemed to me that some odd things are going on in the tests as currently written, like data fitting in `'trig'` mode for a function that is not smoothly periodic.

sfilip:
Thanks @trefethen for noticing this. You are right, some of the tests were "too random". Since we are looking for smooths fits, I've limited how close the abscissas can get to one another and this improves the reliability of ```test_gpr```.

I've also updated the first example using ```'trig'``` mode to sample from a smoothly periodic function and added some extra comments when calling ```gpr``` in ```trig``` mode and a domain is not specified.

trefethen:
Thanks @sfilip!

PR #2250, opened 2018-03-04
Operator Krylov subspace methods for solving second-order ODEs

ajt60gaibb:
At the Chebfun and Beyond conference in 2012, there was a discussion topic on operator Krylov methods for solving ODEs/PDEs via operator-function products.  In this pull request, Marc Aurele Gilles (a graduate student at the Center for Applied Math at Cornell University) and I implement operator CG (`pcg(L,f)` command), MINRES (`minres(L,f)` command), and GMRES (`gmres(L,f)` command) methods for solving second-order ODEs with Dirichlet conditions. Marc derived these analogues himself, though we greatly benefited from the 2012 discussion, Malek and Strakos' 2014 SIAM book on the subject [here](http://epubs.siam.org/doi/book/10.1137/1.9781611973846), and @dlfivefifty's GMRES paper.  I emphasize that our operator Krylov methods work at the "continuous" level with operators and functions. There is no discretization required. In practice, we just allow Chebfun to adaptively resolve the functions that appear in the algorithm to machine precision as the software desires. 

The commands are `pcg(L,f)`, `minres(L,f)`, and `gmres(L,f)` for solving ODEs of the form `-(a(x)u')' + b(x)u' + c(x)u = f` with Dirichlet conditions. The syntax follows MATLAB's `pcg(A,b)`, `minres(A,b)`, and `gmres(A,b)` as closely as possible.  Analogous to the matrix case, we recommend CG, MINRES, and GMRES in different situations:

- If `a(x)>0`, `b(x)=0`, and `c(x)>0`, then the ODE operator is uniformly elliptic, self-adjoint and has non-negative eigenvalues. We recommend CG. 

- If `b(x)=0`, then the ODE operator is self-adjoint and we recommend MINRES. 

- If the above two cases are not met, then we recommend GMRES. 

For example, on an ODE with piecewise functions: 
```
a = @(x) 2+cos(pi*abs((x-0.5)));
c = @(x) 1+10*x.^2;
L = @(x,u) -diff(a(x).*diff(u)) + c(x).*u;
f = chebfun(@(x) abs(x) - 0.5 ,'splitting','on');

N = chebop( L );
N.bc = 0; 
tic
u = N \ f         % Rectangular spectral collocation
toc
Elapsed time is 0.154866 seconds.
u =
   chebfun column (3 smooth pieces)
[      -1,-1.1e-16]       40  -9.7e-17   -0.033 
[-1.1e-16,     0.5]       17    -0.033   -0.015 
[     0.5,       1]       16    -0.015  8.6e-18 
vertical scale = 0.034    Total length = 73

% chebbop/pcg:
tic
v = pcg(N, f)     % Operator CG method 
toc 
Elapsed time is 0.761808 seconds.
v =
   chebfun column (3 smooth pieces)
[      -1,-1.1e-16]       44   2.5e-20   -0.033 
[-1.1e-16,     0.5]       22    -0.033   -0.015 
[     0.5,       1]       18    -0.015   -1e-16 
vertical scale = 0.034    Total length = 84

norm(u - v)
ans =
   3.4261e-15
```
Just like the situation for matrices, it is only computationally beneficial to use iterative methods when `chebop/solvebvp` requires a large discretization size (`>2000`, say). Note that all differential operators in Chebfun have fast operator-function products.

We use projection operators to ensure that the operator Krylov methods respect the Dirichlet boundary conditions. We use operator preconditioners to ensure that the CG error (in the operator analogue of the energy norm) converges geometrically to zero. 
Marc and I are a few days away from finishing a manuscript on the subject of this PR. 

*Warning* 
The implementation is not fully “general”: (a) It is restricted to second-order differential operators and (b) It is restricted to Dirichlet boundary conditions. We shouldn’t run before we are confident in walking (my 14-month old son mostly runs places to his own detriment). Marc and I are hoping that this PR can lead to a discussion on how to generalize our initial implementation. 

There was a `chebop/gmres` command that only worked for integral operators and just immediately called `linop/gmres`. How does one check if a `chebop` object is an integral operator?  If I knew that, then I could ensure backward capability.


Hadrien-Montanelli:
Great job. Have you guys been in touch with @jaurentz? I remember he implemented a PCG for ODEs when he was at Oxford.

ajt60gaibb:
Didn't know that! Emailing @jaurentz now. Thank you @Hadrien-Montanelli.

asgeirbirkis:
@ajt60gaibb sorry for being late to the party.

Regarding your question "How does one check if a chebop object is an integral operator? ",  I don't think there is a simple way. One can check the `diffOrder` of the linop after linearizing:

```
>> L1 = linop(chebop(@(u) diff(u,2))); L1.diffOrder
ans =
     2
>> L2 = linop(chebop(@(u) cumsum(u))); L2.diffOrder
ans =
    -1
```

However, that doesn't work too well when you mix differential and integral operators:

```
>> L3 = linop(chebop(@(u) diff(u) + cumsum(u))); L3.diffOrder
ans =
     1
```

This is caused by the code taking the max of the `diffOrder`s of the input operators for the operator arithmetic, which means that the information about the integral order gets lost. If you want to ensure backwards compatibility, probably the best bet would be to introduce a new `intOrder` property, and keep the information about differential and integral orders separate.

I think that idea has actually popped up couple of times before, so there might be further applications for being able to access this info separately. The logic for implementing this should be pretty straightforward, and would probably just require adding a line to the code everywhere that a `diffOrder` gets updated.

dlfivefifty:
> This is caused by the code taking the max of the diffOrder

I think the current behaviour is “correct”: the behaviour of an integro-differential operator is dictated by the largest derivative, e.g. your example would require 1 boundary condition.

asgeirbirkis:
Yes, that's a good point. However, for my example, I'd expect that for `L3`, one would arrive at `L3.diffOrder = 1`, `L3.intOrder = 1` (or -1, depending on the convention that'd be implemented), so that when it comes to solving a problem, the `diffOrder` could be compared against the number of boundary conditions supplied.

Another application where it might be useful to be able to distinguish pure integral or differential operators from mixed ones is in the adjoint work of @jaurentz . As far as I understand, it does not support integral operators:
```
>> L = chebop(@(u) cumsum(u)); Ls = adjoint(L)
Error using linop/linopAdjoint>parseInputs (line 56)
LINOPADJOINT doesn't support integral operators for the moment.
Error in linop/linopAdjoint (line 26)
parseInputs(L, bcType);
Error in chebop/adjoint (line 41)
[Lstar, op, bcOpL, bcOpR, bcOpM] = linopAdjoint(L, getBCType(N)); 
```

However, the check for integral operator can be tricked by adding an identity operator:
```
>> L = chebop(@(u) u+cumsum(u)); Ls = adjoint(L)
Ls =
   Linear operator:
      u |--> u+cumsum(u)
      [-1,1]
```
while if the `intOrder` could be checked separately, that shouldn't be an issue.

dlfivefifty:
I didn’t know about ` adjoint `, cool! Is it written up somewhere?

nickhale:
Comments:
1.) Currently these methods are quite limited in scope, but there are extensions in the pipeline.
2.) It's not clear that '@chebop/' is the right place for these methods, but '@linop' and '@chebfun' also aren't right.
3.) There is potential confusion with existing '@chebfun/gmres', but this exists already.

Some of these points need further consideration, but since these methods have limited interaction with the existing code base I'd be happy to merge and refine things later. 

I'll wait 24 hours to allow any objections to be raised before merging.

ajt60gaibb:
I already discussed with @nickhale about several things and it is probably worth briefly replying to the above comments:
1) I agree. In particular, we need second-order differential operators with Dirichlet conditions. But, I believe Chebfun should have operator Krylov methods and I hope that the functionality of `pcg`, `minres`, and `gmres` can be expanded over time. (We have to start somewhere and I think this is a good start to a very awkward problem.)
2) Since we will only ever do linear differential operators, `chebop` may not feel correct. However, `chebop` is one of the only data structures in chebfun that gives us the differential operator as a function handle (which is what we really need) as well as allows us to check `diffOrder` and boundary types. 
3) The confusion is already there and we haven't changed it.  We could fix the confusion, but I think it should be a different PR. 

I believe that this PR does not affect the paper of @tobydriscoll, here: 
http://www.math.udel.edu/~driscoll/papers/2010-Driscoll-5980.pdf
(As all the examples in that paper seem to call `chebfun/gmres` not `chebop/gmres`.) 
If we could answer the question: "How does one check if a chebop object is an integral operator?", then `chebop/gmres` should call `chebfun/gmres` if the incoming `chebop` object represents an integral operator.  

The world of iterative solvers in Chebfun clearly needs more consideration to get right.

PR #2254, opened 2018-04-25
Deal with repeated values in Z in AAA

abigopal:
When an element in Z is repeated and one instance of it is chosen as a support point, AAA throws an error because it doesn't drop the appropriate row of the Loewner matrix. This pull request deals with this issue by removing duplicate elements of Z at the start, altogether. I think this is a better strategy than just removing the appropriate indices in the J vector, because it reduces the size of the Loewner matrix that needs to be formed and is therefore more efficient, while remaining mathematically equivalent.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -65,6 +65,7 @@ F = F(toKeep); Z = Z(toKeep);
 toKeep = ~isnan(F);
 F = F(toKeep); Z = Z(toKeep);
 
+% Remove repeated elements of Z and corresponding elements of F:
 [Z, uni] = unique(Z); F = F(uni);
 
 M = length(Z);
--- a/aaa.m
+++ b/aaa.m
@@ -350,9 +350,9 @@ if ( ni == 0 )
     % Nothing to do.
     return
 elseif ( ni == 1 )
-    fprintf('1 Froissart doublet.\n')
+    warning('CHEBFUN:aaa:Froissart','1 Froissart doublet');
 else
-    fprintf('%d Froissart doublets.\n', ni)
+    warning('CHEBFUN:aaa:Froissart',[int2str(ni) ' Froissart doublets']);
 end
 
 % For each spurious pole find and remove closest support point:
--- a/aaa.m
+++ b/aaa.m
@@ -64,6 +64,9 @@ toKeep = ~isinf(F);
 F = F(toKeep); Z = Z(toKeep);
 toKeep = ~isnan(F);
 F = F(toKeep); Z = Z(toKeep);
+
+[Z, uni] = unique(Z); F = F(uni);
+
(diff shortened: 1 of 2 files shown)


trefethen:
@abigopal, could you explain why this is an issue?  Normally svd(A) works when A has repeated rows, doesn't it?  Is there something special about the AAA matrix?  I'd like to know the answer, and indeed, if there's a very short explanation maybe a line or two of comment could be added in the code.

abigopal:
@trefethen, svd(A) works when it has repeated rows, but NOT when there are Inf's in the matrix! When forming A, we use the Cauchy matrix, which we call C, whose entries are of the form C(i,j) = 1/(Z(i) - z(j)) where Z is the collection of sample points and z is the collection of support points. This matrix will clearly have an infinite entry in the rows where Z(i) = z(j). To deal with this, AAA has a vector called J, which initially is set to the numbers 1,2,...,M that correspond to indices of the sample points. When a sample point is chosen as a support point, the corresponding index is dropped from J e.g. if on the kth iteration Z(i) is chosen as a support point, we do J(i) = []. We then only do the SVD with A(J,:) to avoid the rows with infinite values. This works fine unless a sample point, which is chosen as a support point, is repeated in Z. In that case, as it currently stands, the repeated indices after the first aren't removed from J and therefore A(J,:) still has rows with Inf's.

The following code won't work in AAA as currently implemented:
`Z = linspace(-1,1,10)'; Z = [Z; Z]; F = sin(Z); r = aaa(F,Z)`

As a result, we have in the past had to perturb the data when this was a risk and have had to do always drop the first point when discretizing the unit circle.

Also, I think changing this was actually your idea! You sent me an email on 27 March 2018 with this in it:

> Also, it seems embarrassing that the user currently has to 
perturb the data to avoid
error messages in the SVD if two columns are identical.  Could some 
appropriate
fix for this problem be incorporated into AAA?

trefethen:
Thanks @abigopal.  I'm pleased with your change and I've added another: now the Froissart doublets message is a warning, so one can turn it off (this has been annoying me).  Please take a look and if you're happy with that, merge this pull request.

abigopal:
@trefethen, this looks good, but I have a quick question. With this change, if we run the AAA tests we get a warning, because one of the tests (test 15) ends up having to do a cleanup. Should I go ahead and turn off the warning for the tests?

trefethen:
Good point, thanks.  Yes indeed, please do that.  There are a bunch of our existing tests that turn off warnings.  (I fear they weren't all coded with perfect consistency.)

abigopal:
@trefethen I made the change. Could you review?

trefethen:
Looks Good.  Thanks.

PR #2257, opened 2018-05-28
Improved residue computation

abigopal:
As we have seen in numerical experiments, the current contour integral based residue computation can yield low accuracy results when the contour discretization values aren't represented faithfully in floating point arithmetic. This commit instead changes the residue computation to use the identity Res(f/g, z0) = f(z0)/g'(z0), where f and g are analytic at z0. This also obviates difficulties encountered when poles are located close to each other.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -280,64 +280,6 @@ end
 end % End of PARSEINPUT().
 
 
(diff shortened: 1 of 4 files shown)


trefethen:
Looks good.  Thanks @abigopal!

Issue #2270, opened 2018-08-09
Bug in lagpts for small n


nickhale:
I just discovered the following bug
```
>> lagpts(5) - lagpts(5, 'gw')
ans =
  -0.000000000000001
  -0.000000000000001
                   0
   0.000000000000004
  -0.965661467809641
```
Since this (or something similar) occurs for each of the `lagpts` `'method'` options other than `'gw'`, I suspect that the intial guess for this small value is not sufficient for the Newton-like approaches to converge.

Running the following, I find it's a problem for `n = 5,6,7,8`. I therefore suggest defaulting to the Golub-Welsch approach when `n <= 10`.
```
>> for k = 1:20, try, err(k) = norm(lagpts(k) - lagpts(k, 'gw'), inf) > 1e-4;, catch err(k) = 1; end, end, err
err =
     0     0     0     0     1     1     1     1     0     0     0     0     0     0     0     0     0     0     0     0
```

LourencoPeixoto:
Hi

I've worked with REC (Newton iterations on the recurrence relation) for the zeros of Laguerre polynomials in the past. In fact this method is very efficient for such zeros when `n<365` with approximations in Stroud and Secrest. Less than 10 iterations are required. The same holds to Laguerre generalized polynomials. Furthermore REC is more faster than GW.

LourencoPeixoto:
Hi @nickhale . 

I believe that there's the same problem for `n=190, 195, 200, 201, 210` too.

nickhale:
On further investigation, it seems that there are some serious issues with the `lagpts` code in the master branch.

Here are my results (you can find the code below - where oldlagpts is an old version from before the GLR algorithm was removed):

![err](https://user-images.githubusercontent.com/2035010/44529725-1388ea00-a6ed-11e8-807e-c6b1f99441c1.png)
![time](https://user-images.githubusercontent.com/2035010/44529726-1388ea00-a6ed-11e8-8b28-07b0f028a506.png)

My conclusions are: 
- 'rec' seems to break after around n = 370 (overflow?)
- 'exp' is completely broken (the problem seems to be with a few of the larger roots)
- 'RH' gives reasonable accuracy, but only when n is sufficiently large.

My suggestions are (unless there are quick fixes):
- Reinstate 'glr'
- Disable 'exp'
- Use 'gw' for n < 128
          'glr' for 128 <= n < 512
          'RH' for 512 <= n.

```
nn = round(logspace(2, 3, 100));
[...]

PR #2275, opened 2018-09-04
fix-lagpts

nickhale:
* Reinstate GLR
 * Add new REC
 * Adjust default methods. 

See #2270 for discussion.

% Determine default method (heuristic):
n < 300 = rec
n < 1000 = gw
n < 3000 & alpha ~= 0 = gw
n < 3000 & alpha = 0 = glr
n > 3000 = RH


Hadrien-Montanelli:
Thanks @nickhale, I'll take a look this week.

PR #2276, opened 2018-09-10
Improve Spherical Harmonics algorithm

NBoulle:
This improves the `sphharm` code by implementing a fast and stable algorithm to compute the values of the associated Legendre polynomials.

The algorithm is no longer based on matlab's `legendre` function and uses the Modified Forward Column method described in the [Holmes and Featherstone paper (2002)](https://link.springer.com/article/10.1007/s00190-002-0216-2).


gradywright:
Good idea @NBoulle and @ajt60gaibb!  Please include this in the PR and I'll review all the code after that.

NBoulle:
I didn't find a way to speed up this code further using low rank form so you can merge this if you want @ajt60gaibb.

Issue #2278, opened 2018-09-29
bug in `chebop/matrix` when domain has internal breakpoints


trefethen:
The chebop `matrix` command allows you to instantiate a spectral discretization of a specified size.  For example,
```
L = chebop(@(u) diff(u,2), [-1 0 1]);
L.lbc = 0; L.rbc = 0;
matrix(L,[10,10]))
```
should produce a 24x24 matrix: 20x24 for the differential operators on the subintervals [-1,0] and [0,1], plus an addditional 4x24 matrix whose rows correspond to the left BC, the right BC, and the coupling conditions of order 0 and 1 at x=0.  The bug, however, is that the coupling conditions are missing, so you get a matrix of order 22x24 instead.

@tobydriscoll suggests (private email communication) that a fix may be to insert the line

```L = deriveContinuity(L);```

after line 40 of `chebop/matrix.m`.  Hopefully one of us can find the time to check this a bit, e.g. on the examples in my _SIAM _Review paper "Block operators and spectral discretizations" coauthored with @jaurentz, and issue a pull request.  Probably very little depends on `matrix`, so it's a low-risk fix.  

I regard this fix as unimportant for Chebfun's functionality but quite important for its educational role, i.e., in teaching Chebyshev spectral methods.

nickhale:
Closed by #2279.

PR #2279, opened 2018-10-02
fix-chebop-matrix

nickhale:
Enforce continuity constraints in `chebop/matrix` as suggested in #2278.

As far as I can tell, there are no dependencies on `chebop/matrix' so I'm happy to make this change.

I also adjusted the help documentation a little (but it's still not perfect) and added some tests.


trefethen:
Thanks @nickhale!  I am happy to merge this.  I changed the help text a bit further, and some associated variable names, so that now it speaks of `matrix(L, DIM)` rather than `matrix(N, DIM)`.  The latter was confusing (I know from experience) since `N` looked like an integer; and it's supposed to be a linear operator anyway.

PR #2283, opened 2018-11-16
feature-chebfun2-end

### LINKED ISSUE #2282: Chebfun2 does not have an 'end' function
Given a chebfun2 `u` with a domain [0 2 0 2], one might expect `u(end,end)` to return `u(2,2)`. However, it returns `u(1,1)`, presumably because there is no `chebfun2/end()` method.

Example:
```
>> u = chebfun2(@(x, y) x+y, [0 2 0 2])
>> u(end,end)
ans =
     2
>> u(2,2)
ans =
     4
>> u(1,1)
ans =
     2
```

nickhale:
Add chebfun2/end method and test.

Closes #2282.

### DIFF
--- a/@chebfun2/end.m
+++ b/@chebfun2/end.m
@@ -1,0 +1,31 @@
+function e = end(f, k, n)
+%END   Rightmost point of a CHEBFUN2 domain.
+%   END(F, K, N) This function is called for indexing expressions involving a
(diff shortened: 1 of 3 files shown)


gradywright:
Should `end` be implemented at the `separableApprox` level instead?

PR #2285, opened 2018-11-30
Feature ballfun

NBoulle:
Add ballfun for computing with functions and vectors on the unit ball.


trefethen:
I have been playing with versions of Ballfun in the past month and suggesting various changes that have now been incorporated.  It seems very promising and I hope some others will now take a look.

A small correction to the `help ballfun` text please: currently it is stated that one can use either (x,y,z) or (r.lambda,theta), but the user is not told how to inform the constructor of which of these choices is being used.

Might the help text also have a couple of examples?  For example:

```
Examples:

   f = ballfun(@(x,y,z) 1); sum3(f)/(4*pi/3)

   plot(ballfun(@(x,y,z) sin(40*x))), colorbar
```

gradywright:
I've started my initial code review.  I will post comments as I come across them, but try to batch them up according to chunks of time I have to looking at the code.  Below is my first batch.

- [ ] Add option for constructing a ballfun of a given length (similar to chebfun2, chebfun3, spherefun, and diskfun) - this could be moved to a feature request in a later release.
- [ ] Add option for constructing a ballfun of a given tolerance (similar to chebfun3) - this could be moved to a feature request in a later release.
- [ ] "vertical scale" is not included in `display()` output.  It is amongst all other chebfuns (chebfun, spherefun, diskfun, etc.)
- [ ] `fevalm` states that -1<=r<1, but should this be 0<=r<=1?
- [ ] `subsref' says f(r, :, :, 'spherical') returns a spherefun, which is really cool.  However, this may be confusing since spherefuns live on the unit sphere.  So, we should be clear that the domain has been scaled to the unit sphere.  Since we do the scaling for f(r,:,:,'spherical') to the unit sphere, we should also allow f(:,:,theta,'spherical'), which would be a slice along the plane at height z=cos(theta), and scale this to the unit disk so a diskfun can be returned.
- [ ] The error at line 80 of `ballfun/subsref` is not the most informative since evaluations other than triples of values are allowed (e.g. `f(1,:,:,'spherical')` is allowed but `f(:,pi/4,:,'spherical')` is not)

gradywright:
- [ ] Comment about rank does on line 55 of `constructor' not make sense since the construction does not use a low rank representation.
- [ ] Is the third dimension called "tubes" or "fibers", it seems both terminologies can be found in the code and one should be consistent.
- [ ] `plot(f,'slices')` makes more sense to me than `plot(f,'slice')`.  I suggest adding both options to `plot`.
- [ ] A handle to the plot should be returned in `plot`.
- [ ] Why is the paper position being set on the plot at end of the sub-function `plot2` in `plot`?
- [ ] It would be nice `plotcoeffs` included what the rows, columns, and tubes correspond to in terms of the spherical coordinates.
- [ ] Add a `surf` command that just calls `plot`; right now `surf(f)` causes an error
- [ ] The use of `mean` is not consistent with other chebfuns, this should be `mean3`.  I suggest adding `mean(f)` (or `mean(f,1)`) to give the mean in the radial direction and return a spherefun.  Similarly, 'mean(f,3)' should be a diskfun.  I don't know what `mean(f,2)` is in terms of chebfuns.  I guess it should throw an error saying this is not supported.  `mean3` should be added and do what `mean` currently does.  `mean2` should also be included, but you don't have to implement it yet, just say it's not supported.

NBoulle:
Thanks for the comments! I'm starting to work on them.
Concerning the first two comments, I'll create an issue to move them to a feature request as it affects the constructor.

gradywright:
- [ ] Feature: add option for real-valued solid harmonics
- [ ] `f = ballfun([])` does not return an empty `ballfun`.  This is inconsistent with the other chebfuns (e.g. chebfun2, chebfun3, diskfun, etc.). Adding this capability will require adding an `isempty` method and then checking for empty ballfuns in several of the other methods.
- [ ] Add example in help text for `helmholtz`
- [ ] Add examples in help text for `plot` for the different options.
- [ ] It is confusing to call `plot` on a complex-valued ballfun and get a plot of the real part.  I think an error should be thrown.
- [ ] Remove axis labels in the default plot option; it is not standard to include these, but to let the user add them.
- [ ] Any call to `gallery.ballfun` should produce a plot of the function if the user does not specify an output.
- [ ] The options `deathstar` and `stripes` in `cheb.galleryball` result in the warning 'Not well-approximated by a Chebyshev-Fourier-Fourier expansion.'
- [ ] The help text for `abs` says that the function will not work if the function has any negative values.  However, there is no check to see if the function is negative, instead it just tries to construct the absolute value of f, which can take a long time to fail if f does have negative values.  You could do a quick check as in `separableApprox/singleSignTest` and throw an error if negative values are detected.

gradywright:
- [ ] Add example in help text for `poisson`
- [ ] The definition in `iszero` for whether a `ballfun` is zero does not match the other chebfuns.  The coefficients (or values) have to be identically zero for `iszero` to return true, not small up to some tolerance.  For example, `iszero(ballfun(@(x,y,z) 1e-32))` returns 1, but
`iszero(spherefun(@(x,y,z) 1e-32))` and `iszero(chebfun3(@(x,y,z) 1e-32))` return 0.  This change will impact how `isequal` is currenty implemented.

gradywright:
I'm a bit confused with the new `subsref` capabilities `f(:,:,c)`.  For example, for `f = ballfun(@(x,y,z) cos(z + pi*sin(pi*x)))`, what should `g = f(:,:,0.9)` produce?  Right now it's a `diskfun` with non-sensical astronomical values.

trefethen:
My feeling (not based on deep thought) is that you might have Ballfun produce diskfuns in response to `f(:,:,s)` or `f(:,s,:)` or `f(s,:,:)` if s=0, but otherwise throw an error.

gradywright:
That's a reasonable approach @trefethen.  Alternatively, `f(:,:,s)` or `f(:,s,:)` or `f(s,:,:)` could return a `diskfun` representing the values of `f` on the corresponding plane intersecting the sphere.  Of course, this would not be technically correct since diskfuns live on the unit disk, so it would be a scaled version of the function.  This feature would be consistent the currently allowed notation `f(c,:,:,`spherical`)` , for `0<=c<=1`, which returns a `spherefun`.

NBoulle:
Yes this is what I tried to do. I need to do more tests as there might still be some bugs there.

NBoulle:
> It is confusing to call plot on a complex-valued ballfun and get a plot of the real part. I think an error should be thrown.

I added a warning if the function isn't real saying that the real part is plotted

gradywright:
- [ ] Include a `sample` function that evaluates the ballfun on a non-doubled-up tensor product grid (similar to the other chebfuns).  This would be different than `coeffs2vals`, which gives the values on the doubled-up grid.

gradywright:
- [ ] It appears as though `diff` has the option of doing derivatives with respect to r, theta, and lambda, but this is not advertised in the help text.  Is this meant to be a hidden option?  If not, then some text should be added describing this option.

NBoulle:
Yes, differentiation in spherical coordinates doesn't automatically give smooth functions so we don't advertise it.

heatherw3521:
I have MATLAB R2018a and when I try to construct a ballfun from values, I get an error related to use of `mean2(g(1,:,:))`, where `g` is an array. 

Error is: `To use 'mean2', the following product
must be licensed, installed, and
enabled:
  Image Processing Toolbox`

ajt60gaibb:
Odd. So that call to `mean2(x)` should be replaced by `sum(x(:), 'default') / numel(x)`.

heatherw3521:
In the `diskfun` command, I can ask for a slice at `z = 20` and it returns a diskfun. Should it throw an error instead? The same issue for the `spherefun` command: I can ask for a spherefun where `r > 1` (also when `r=0`). For `diskfun`, this  is an issue in `feval`: points far outside the domain can be evaluated without an error or warning.

trefethen:
The Chebfun tradition is that you can evaluate outside the domain of definition without an error; I wouldn't object to a warning, however.  Computing a whole slice outside the domain, on the other hand, is taking such behavior to a new level! -- maybe that should indeed be illegal.

NBoulle:
I agree with @trefethen : we can still evaluate a ballfun at r>1 to get a spherefun. However, I now throw an error if we try to compute a slice outside the ball.

heatherw3521:
just a few things to do: 

- can you add some tests for `mean` and `mean2`?
- the help text in `helmholtz` for Neumann BC needs to be fixed. 
- should we add an `integral` command? (see, for example, chebfun3)

heatherw3521:
- should `integral` in ballfun match `integral` in chebfun3? at least in the sense that `integral(f) = sum3(f)`? 

- For ballfunv,  I think the help text in `HelmholtzDecomposition`, `PT2ballfunv`, and `PTdecomposition` needs a bit more explanation. For example, you need to tell me what `rT` refers to. 

- you might want to use a higher tolerance for  throwing your warning about a divergence-free field in `PTdecomposition`. Also, the warning should read `the vector field is not divergence free. `

- help text in `ballfunv/times` is confusing.

NBoulle:
Ok, I'll change `integral` and remove `integral2` and `integral3`.

heatherw3521:
Once the comments above are addressed, I'm happy to merge and close this PR. @gradywright are you happy/good with that?

heatherw3521:
I think it would be really nice if some of the gallery functions are set to automatically plot in different and interesting ways. For example, I really like the wedge plots and it would be cool if you showed them off in the gallery. We did this kind of thing, for example, in  `cheb.gallerydisk`.

PR #2288, opened 2019-01-29
Fix-chebfun-conv

nickhale:
Resort to quadrature approach when there are many patches required in convolution due to vastly different domain sizes. (Previously we just threw an error.)

Also tidied the code and added some tests.


ajt60gaibb:
This is very nice, @nickhale.

PR #2290, opened 2019-02-04
Fix chebfun2 constructor bug

### LINKED ISSUE #2289: Chebfun2 constructor bug
If one tries to execute the following harmless looking code, then the Chebfun2 never resolves the function: 
```
r = chebfun2(@(x,y) sin(2*pi*x));
f = chebfun2(@(x,y) r(x,y).*y);
```
It is linked to the following unintentional behavior: 
```
x = linspace(-1,1,10); 
size( r(x,x) ) 
ans = 
   10   1  
```
The result of `r(x,x)` should be a row vector, not a column vector. 

(This bug was identified by Prof. Donna Calhoun.)

ajt60gaibb:
Closes #2289.

### DIFF
--- a/@chebfun2/constructor.m
+++ b/@chebfun2/constructor.m
@@ -177,7 +177,7 @@ while ( ~isHappy && ~failure )
                 colVals(PP(kk+1:nn, 1), kk)*(rowVals(kk, :)./pivotVal(kk));
         end
         
-        % If function is on rank-1 then make rowValues a row vector:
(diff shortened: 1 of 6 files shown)


gradywright:
I agree @ajt60gaibb.  I reviewed the code and am happy to merge it, but I'll wait to see if any one else wants to comment.  

Are there some examples that now won't run because of this change?

ajt60gaibb:
I haven’t found any examples that break.

PR #2293, opened 2019-02-24
fix-simplify-conv

nickhale:
Simplify the output of `chebfun/conv()`. 

Reasons for doing so: 
 * Consistency with old conv method (pre-Legendre series)
 * Convolution of two smooth functions typically has a much shorter length than input.


trefethen:
Thanks @nickhale from five of us at Oxford.  Merging.,

PR #2296, opened 2019-03-07
fix-chebfun-chebpts

nickhale:
Remove some unnecessary  overheads in `chebfun.chebpts'.

Before:
```
>> tic, for k = 1:1000, chebpts(10); end, toc
Elapsed time is 0.240198 seconds.
```
After:
```
>> tic, for k = 1:1000, chebpts(10); end, toc
Elapsed time is 0.156915 seconds.
```

It's not a huge difference, but it's a trivial change.


trefethen:
Thanks @nickhale.  The Oxford gang have looked at this and will merge.

PR #2307, opened 2019-06-05
Improve chebop2 help text, fix subassign, fix set

ajt60gaibb:


trefethen:
@danfortunato, here at Oxford we've looked superficially at this and are happy.  If you're happy, would you go ahead and merge?

PR #2308, opened 2019-06-05
Faster poisson solver

ajt60gaibb:
Old code, which called chebop2's automatic differentiation code on Laplace operator:  
```
tic
u = chebfun2.poisson(chebfun2(1),0);
toc
Elapsed time is 0.727262 seconds.
```
Updated code, which just fills in the Laplace operator (as we know it is Laplace):   
```
tic
u = chebfun2.poisson(chebfun2(1),0);
toc
Elapsed time is 0.592926 seconds.
```

### DIFF
--- a/@chebfun2/poisson.m
+++ b/@chebfun2/poisson.m
@@ -58,7 +58,7 @@ if ( nargin == 1 )
     % Call is POISSON(F) so set G = 0 and use chebop2 to determine the
     % discretization size:
     g = 0;
-    N = chebop2(@(u) diff(u,2,1) + diff(u,2,2), f.domain);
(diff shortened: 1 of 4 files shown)


danfortunato:
@ajt60gaibb, did you want to reopen this? I have merged #2307, but the faster Poisson solver (dc9e9ca) was not part of that.

ajt60gaibb:
I realized that this broke encapsulation. (Currently, `chebfun2/poisson` needs to know the properties of the `chebop2` object.) The right strategy is to write a command in `chebop2` called `SetupPoisson` and then get `chebfun2/poisson` to call that.

PR #2314, opened 2019-06-28
Add poisson operator

ajt60gaibb:


nickhale:
Can I propose (a) naming this method `laplace` rather than `setuplaplace` and (b) taking the default domain from `cheboppref` if none is passed? 

Might it be useful to add other operators (e.g., Helmholtz)?

ajt60gaibb:
I did `setuplaplace->laplace`.  Did you really mean `cheboppref` (the preferences for the 1D code)? We don't have a `chebop2pref` but we could pull from that preference structure if we make one...

nickhale:
`chebop2` gets a default domain from somewhere:
```
>> chebop2(@(u) lap(u))
chebop2 of rank 2: (linear bivariate operator)
         domain               operator  
[ -1,  1] x [ -1,  1]  @(u)lap(u)
```

ajt60gaibb:
hardcoded. gulp.

PR #2321, opened 2019-07-02
Add best l1

ajt60gaibb:
This PR gives Chebfun the functionality to compute best L1 polynomial approximants to continuous functions.


trefethen:
Fantastic to have this!

I am merging the PR and then I will install the associated example (renamed to `BestL1`).  However, we need some tests!

PR #2324, opened 2019-07-08
feature-dont-use-deal-in-clenshaw

nickhale:
Don't use deal in Clenshaw. Too much overhead.


ajt60gaibb:
Interesting. What's the lesson: "Only use `deal` to avoid copying large matrices."

nickhale:
I'm not even sure that's the case. I think `deal` just provides convenient syntax for some situations, rather than any performance benefits. 

In this case the overheads involved are high compared to the rest of the Clenshaw algorithm, and the syntax gain is minimal. (This is often the case with some other functions, such as `flipud` and `cell2mat`.)

PR #2335, opened 2019-08-15
add degree option to aaa, make sure no clash with mmax

nakatsukasayuji:
This allows aaa to be called with the degree specified to n, i.e., the output rational function will be of type (n,n). This is done by 
r = aaa(F,Z,'degree',n);
which is equivalent to the previously available 
r = aaa(F,Z,'mmax',n+1);
but to many users degree is more relevant/intuitive, especially with AAA-Lawson refinement (which is now called by default).

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -1,5 +1,5 @@
 function [r, pol, res, zer, zj, fj, wj, errvec, wt] = aaa(F, varargin)
-%AAA   AAA and AAA-Lawson (near-minimax) rational approximation.
+%AAA   AAA and AAA-Lawson (near-minimax) real or complex rational approximation.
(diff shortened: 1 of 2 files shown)


trefethen:
This seems incomplete -- the help text didn't mention 'degree', although 'degree' should probably be more publicized than 'mmax' now.  I have made some modifications which perhaps will help.

Also I notice test #24 doesn't pass.  I haven't investigated -- is this the result of some change in the code, or has it been failing all along?

After hearing comments from @nakatsukasayuji I will continue reviewing.

trefethen:
OK, looks good now.  Thanks @nakatsukasayuji.  I will merge now.

PR #2336, opened 2019-08-21
feature-padeapprox-radius

nickhale:
Add option to pass radius when computing coeffs in padeapprox. (Implemented at request of JAC Weideman.)

Allows the easy construction of Pade' approximants of, for example, f = x/(1-x). This was not previously possible as one could not change the radius of the complex valued sampled used to compute the Taylor coefficients of f. (Of course, one could get around this by constructing the coefficients manually.) Now it works directly:

```
[r, a, b, mu, nu, poles] = padeapprox(@(x) x./(1-x), 5, 5, struct('r', 0.5))
>> [r, a, b, mu, nu, poles] = padeapprox(@(x) x./(1-x), 5, 5, struct('r', 0.5))
r =
  function_handle with value:
    @(z)polyval(a(end:-1:1),z)./polyval(b(end:-1:1),z)
a =
     0
     1
b =
   1.000000000000000
  -1.000000000000000
mu =
     1
nu =
     1
poles =
   1.000000000000000
```

Note, I decided to go with passing the structure rather than additional parameters as (a) this seems to be the direction MATLAB is moving (see most of the optimisation routines and ODE solvers) and (b) I think it's tidier this way as it avoids including empty input values.

### DIFF
--- a/padeapprox.m
+++ b/padeapprox.m
@@ -1,4 +1,4 @@
-function [r, a, b, mu, nu, poles, residues] = padeapprox(f, m, n, tol)
+function [r, a, b, mu, nu, poles, residues] = padeapprox(f, m, n, opts)
 %PADEAPPROX   Pade approximation to a function or Taylor series.
 %   [R, A, B, MU, NU, POLES, RESIDUES] = PADEAPPROX(F, M, N, TOL) constructs a
 %   Pade approximant to F using the robust algorithm from [1] based on the SVD.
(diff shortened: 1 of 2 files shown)


trefethen:
It will be great to have this functionality -- good idea!  However, I'm afraid I am not happy with the `'struct'` syntax.  The Chebfun project has not made the decision to go down that route, which is relatively hard on users.  Instead, I believe this should follow our standard parseinputs type approach, allowing e.g.
```
padeapprox(..., 'tol', tol, 'radius', radius )
```
Since the current syntax assumes there's a `tol` but not a `'tol'`, this will entail introducing an error message if the users calls `padeapprox` in the old way specifying `tol` but not `'tol'`.

nickhale:
My aim was to not do anything which would break the existing syntax, but I don't feel strongly either way. Does ATAP make use of passing a tol?

trefethen:
I just checked, and ATAP uses `tol` twice (to set it to zero): on pp. 243 and 259, the latter being important for effective eta extrapolation.  I'd prefer that such users hit a warning message than that we introduce a `'struct'`.  Alternatively with a little more programming we could make `parseInputs` work with the current syntax _and_ accept a `'tol'` and/or a `'radius'`.

nickhale:
Would you be happy with 
```
function [...] = padeapprox(f, m, n, tol, r, N)
```
where placing an empty value, e.g., 'padeapprox(@(x)1./(x-1) 2, 3, [], 0.5)' will default the missing value to some set default? This is kind of syntax `pcg` and `gmres` use.

This is then consistent with the current syntax and we wouldn't need a warning.

trefethen:
Yes, that's ok with me.  BTW I'm signing off now... back online this evening.

trefethen:
Thanks @nickhale!

PR #2339, opened 2019-09-19
fix-chebtech-abstract-compose

nickhale:
When running Chebfun in MATLAB 2019b I get the following warning:
```Warning: In the 'chebtech' class, the 'compose' method has both the Abstract attribute and a function body. ```

I think the solution is simply to remove the declaration of `compose` as an abstract method in `@chebtech`, as implemented in this pull request. Although `compose` is implemented in `@chebtech1` and `@chebtech2` these are not *required* as the `@chebtech` method provides the necessary functionality. Therefore `compose` should not be abstract.


trefethen:
The lack of this fix is causing trouble now when I publish examples, so I'm going to merge this PR.

PR #2340, opened 2019-09-30
Conformal

trefethen:
This PR introduces a new code `conformal` to compute the conformal map of a smooth simply-connected region onto the unit disk.  The region is specified by its boundary given as a periodic chebfun.  The default algorithm is a discretization of the Kerzman-Stein integral equation (with thanks to Anne Greenbaum and Trevor Caldwell), and as an alternative (much less robust) one can specify `'poly'` to invoke a simple algorithm based on solving a Dirichlet problem with a polynomial.

Conformal maps of polygons via the lightning Laplace solver are under investigation, but not covered in this PR, which works with smooth domains only.

@abigopal has agreed to do the review, and I hope @tobydriscoll and @nakatsukasayuji will also take a look.  It's really fun to play with this via, e.g.,
```
C = chebfun('exp(pi*1i*t)','trig')*(1+.15*randnfun(.2,'trig')); conformal(C, 'plots');
```

### DIFF
--- a/conformal.m
+++ b/conformal.m
@@ -1,0 +1,266 @@
+function [f, finv, pol, polinv] = conformal(C, varargin)
+%% CONFORMAL  Conformal map to unit disk
+%   CONFORMAL(C, ctr) computes a conformal map F of the region bounded by the
+%   complex periodic chebfun C to the unit disk and its inverse FINV, with the
+%   interior point ctr mapping to 0.  Both maps are represented by function
+%   handles evaluating rational functions, whose poles are optionally returned
+%   in the vectors POL and POLINV.  If ctr is omitted it is set to 0.
+%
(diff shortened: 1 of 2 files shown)


abigopal:
Hi,

This looks very good. One quick question. Line 102 in conformal.m currently reads:

``warning(w2.state, 'MATLAB:rankDeficientMatrix')``

My knowledge of MATLAB's warning system is admittedly rusty, but should this perhaps be

``warning(w2.state, 'CHEBFUN:aaa:Froissart')``?

Also, you might want to consider a diagonal scaling for the least-squares system in line 85 in the long-term, but I'm fine to leave this as is for this pull request.

trefethen:
Thanks @abigopal.  I've fixed the error you spotted and also rewritten a few things to make the code more scale-invariant (it's not perfectly so, since `aaa` is not perfectly so).  If you're happy, please merge.

abigopal:
Looks good!

PR #2341, opened 2019-10-21
Fix the bug of a small gap when plotting a periodic complex chebfun C

NBoulle:
This should fix the plotting issue raised by @trefethen when plotting the following complex curve:
```
C = chebfun(@(x)exp(1i*pi*x),'trig');
plot(C), axis equal
```

### DIFF
--- a/@trigtech/plotData.m
+++ b/@trigtech/plotData.m
@@ -38,7 +38,6 @@ end
 len = max([length(f), length(g), length(h)]);
 npts = min(max(501, round(4*pi*len)), trigtech.techPref().maxLength);
 
-
 % Initialise the output structure:
 data = struct('xLine', [], 'yLine', [], 'xPoints', [], 'yPoints', [], ...
     'yLim', [], 'defaultXLim', 1, 'defaultYLim', 1);
@@ -104,6 +103,10 @@ elseif ( isa(g, 'trigtech') )
     % yLim:
     ydata = [get(g, 'lval'); data.yLine; get(g, 'rval')];
     data.yLim = [min(ydata(:)) max(ydata(:))];
+    
+    % Include end points to the curve
+    data.xLine = xdata;
+    data.yLine = ydata;
     
     if ( isa(h, 'trigtech') )
         % PLOT3(F, G, H)


trefethen:
Thanks @NBoulle, but isn't there a more fundamental question?  At present, it seems that `@trigtech/plot` is never called when a trigfun is plotted.  Is this as it should be?  I am concerned that the fix you have coded doesn't get at the root of the problem.  If I'm mistaken about `@trigtech/plot`, please explain.

NBoulle:
I think it's the same as `@chebtech/plot`, which is not called as well. The chebfun class has its own plot function and `@chebtech/plot` and `@trigtech/plot` exist just to visualize objects created by these subclasses. In theory one could unify these commands I guess.

trefethen:
@nickhale and @gradywright, would you give us your views?  Here's the issue: at present, it seems Chebfun never calls `@chebtech/plot` or `@trigtech/plot`.  Is this a mistake, or intentional, or what?  It seems to me that if these codes are not part of our actual design, they should be deleted.  Please advise.

gradywright:
My understanding is that this was an intentional design.  The plot functions in `trigtech` and `chebtech` only exist to plot those instantiations of those classes.  For me, this was useful for debugging purposes when writing `trigtech`, but I don't know what value the codes have at this point.

NBoulle:
To me the current way of doing `plot` is fine as we only have to maintain `chebfun/plot`, which calls `chebtech/plotData` or `trigtech/plotData` depending on the function. `trigtech/plot` and `chebtech/plot` are still useful for debugging so I think they should be kept.

aaustin141:
@gradywright is correct.  `@chebtech/plot` and `@trigtech/plot` are minimal plotting functions to aid with debugging `chebtech` and `trigtech`; they are not intended to be called by `@chebfun/plot`.

Plotting in Chebfun is pretty complicated.  The model we ended up going with was to have `chebfun` request the data for the plot from the techs using the `plotData` functions and then use what they return to assemble the plot at the top level.  This works, but it breaks encapsulation:  `chebfun` must be aware of everything a lower-level class could need/want in terms of plotting.  (NB:  `display` works similarly.)

It would be much nicer if each lower-level class were responsible for plotting itself directly, but I'm not sure this is possible, and even if it is, it might be too ugly to be desirable.  Consider even the basic problem of plotting a no-frills, one-fun function on an interval other than [-1, 1].  `chebtech` doesn't know about non-unit intervals, so if it plots itself, an upper layer (`bndfun`) is going to have to grab the already-plotted data, rescale the x-coordinate, and replot it.

We could do this, but it's pretty ugly.  And plotting (which is already one of the slowest things we do) might take a performance hit from the replotting that takes place.  That might be avoidable by playing games with delayed drawing, plot visibility, etc., but now things are getting complicated again.  And this is only for the most basic sort of plot.  Throw delta functions and singularities into the mix, along with the nightmare that is determining axis limits, and I'm not sure the result will be any nicer than what we have now.

If we could redesign this such that the upper layers only needed to add _new_ plotting data instead of modifying _existing_ data as it bubbles up, we'd be in business, but I don't think that's possible.

trefethen:
Thanks @NBoulle.

PR #2342, opened 2019-11-10
Fix diskfun roots

heatherw3521:
fixes a few small inconsistencies with `diskfun/roots` and `chebfun2/roots`: 

1. no longer searches for roots across doubled up domain (this was resulting in contours found for negative radius values)
2. returns contours as complex-valued chebfuns, rather than real valued parametrizations.


trefethen:
Thanks @heatherw3521!  I will now merge, and also adjust the Chebfun Guide.

PR #2343, opened 2019-11-18
fix-rdivide-spurious-inf-and-nan

nickhale:
Avoid spurious NaNs and Infs from removable singularities in `@chebfun/rdivide`.


trefethen:
Thanks @nickhale.

PR #2344, opened 2019-11-29
fix-jacpts

nickhale:
Fix some problems in `jacpts` pointed out by Javier Segura.


ajt60gaibb:
Thanks, @nickhale. 

We currently have one test that fails: 
```
>> test_conformal
ans =
  1×4 logical array
   0   1   1   1
```
But, since `test_conformal` doesn't call `jacpts` or `ultrapts`, I don't think it is this PR at fault.

PR #2347, opened 2020-02-20
fix-tests

nickhale:
Fix
* `chebop/test_quiver.m`
* `diskfun/test_optimization.m`
* `trigtech/test_compose.m`
by adjusting some tolerances.


abigopal:
Oh, I guess the issue with trigtech/test_compose was also a tolerance issue. I ran the tests and they look good. Thanks, Nick!

PR #2348, opened 2020-02-21
Feature smoothie

gradywright:
This pull request is created so we have official comments and tracking of the new `smoothie` function introduced by @trefethen.  Here are my comments and questions after reviewing the code:

- Where does the name "smoothie" come from?   My first thought when hearing this name was about a blended frozen fruit beverage.

- There is not a test for smoothie.  It looks like most functions in the chebfun directory have tests.

- The code is quite slow when not using 'trig' mode when the domain is increased.  Here's an example:
```
tic, f = smoothie([-10 10]); toc
Elapsed time is 5.797423 seconds.
tic, f = smoothie([-10 10],'trig'); toc
Elapsed time is 0.057221 seconds.
```
The issue is with line 47, which involves evaluating a trigfun at a very large number of chebyshev points.  This evaluation is done via Horner's method. So according to the definitions in the code, the cost scales like O(diff(domain)^2), with a very large constant here (>1e7).   It seems like the NUFFT could be used to speed things up, or perhaps there is a more direct approach to obtaining a smoothie using Chebyshev coefficients directly.

### DIFF
--- a/smoothie.m
+++ b/smoothie.m
@@ -54,7 +54,7 @@ c(1,:) = sqrt(2)*real(c(1,:));
 c = (exp(-sqrt((1:m)'/L)).*c)/sqrt(L);  % root-expoential decay
 c = [conj(c(end:-1:2,:)); c];           % symmetrize for real result
 if cmplx
-disp('not implemented');                 
+    disp('not implemented');                 
 end
 f = chebfun(c, dom, 'trig', 'coeffs');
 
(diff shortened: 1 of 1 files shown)


trefethen:
Thank you @gradywright !  Good idea to issue a pull request, much appreciated.

About tests, there is one -- see `test_smoothie.m` in `tests/misc`.  (It's just a sanity check.}

About the name "smoothie", I chose that to be memorable, having found in talking with half a dozen mathematicians around here that nobody had seen an example of a C-infty but nowhere-analytic function.  So I like the idea of the Chebfun team putting a memorable name on something of some mathematical interest.  If the name really bothers you, however, please get back to me.

About timing, you are entirely right, and I am aware of this.  Smoothies, being non-analytic, have long Fourier and Chebshev series -- on an interval of length 20 as in your example, the Cheb series has length around 72000.  (The trig series are equally long, but nothing of complexity O(n^2) happens when you construct one.  Compute its roots, though, and you'll see the slowdown.)  Yes, it would be interesting to develop algorithms to improve this, and I agree the NUFFT might be a part of this, on which the Chebfun world has some experts.

Meanwhile, with gratitude, I will now merge the PR.

PR #2354, opened 2020-04-16
Implement bumpy spherefun plotting

danfortunato:
This PR implements "bumpy" surface plots for `spherefun`, in which the radial component is bumped out according to the function values.
```
f = cheb.gallerysphere('moire');
subplot(121), surf(f)
subplot(122), surf(f, 'projection', 'bumpy')
```
Screenshot 2020-04-16 16 54 43

### DIFF
--- a/@spherefun/surf.m
+++ b/@spherefun/surf.m
@@ -8,6 +8,7 @@ function varargout = surf(f, varargin)
 %   SURF(..., 'PROJECTION', type) plots F using one of the following map
 %   projections specified by the string type:
 %     'sphere'          : 3D plot on the sphere (default)
+%     'bumpy'           : 3D plot on the sphere with bumps
 %     'equirectangular' : Standard latitude-longitude cylindrical projection
 %     'hammer'          : Hammer (or Hammer-Aitoff) pseudoazimuthal projection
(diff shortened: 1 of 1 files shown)


gradywright:
This looks great @danfortunato!  Thanks for review it @heatherw3521.

trefethen:
Thanks @danfortunato!  I've tried this with pleasure.  Incidentally, in my experience most surface plots are better if you turn on the camlight, as in the demo below.  If this appeals to you too, feel free to add another example in the `spherefun/surf` help text if you wish (which wouldn't need a pull request).  I suppose it's conceivable you might event want camlight to be on by default when bumpy is invoked, but I don't have enough perspective to have an opinion on that.

```
f = spherefun.sphharm(10,4);
>> subplot(1,2,1), surf(f,'projection','bumpy'), axis off
>> subplot(1,2,2), surf(f,'projection','bumpy'), camlight, axis off
```

camlightfig

danfortunato:
Thanks @trefethen! Good idea. I've added an example to the help text in 67abeed111eca639b44a83df34e560945f5c60be that shows bumpy + camlight. I'm similarly not sure if camlight should be on by default.

PR #2365, opened 2020-07-10
writing more helpful error message for chebfun2 * adchebfun2

ilic-stefan:
I've never done this before, so let me know if I messed up/need to change something.

### DIFF
--- a/@adchebfun2/adchebfun2.m
+++ b/@adchebfun2/adchebfun2.m
@@ -212,6 +212,15 @@ classdef (InferiorClasses = {?chebfun2}) adchebfun2
                 return
             end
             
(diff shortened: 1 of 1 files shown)


ajt60gaibb:
Ah, I just found this @ilic-stefan. This PR ensures that an error is thrown when someone tries to do:
```
N = chebop2(@(x,u) x*diff(u,2)); 

Cannot use * between chebfun2 and adchebfun2.
Did you mean .* ?
```
Recall that in Chebfun2, `f*g` computes the integral of `F(s,y)G(x,s)` over `s` (the continuous analogue of matrix-matrix multiplication). The user almost surely intended to use `.*`.

PR #2366, opened 2020-07-12
Vectorize ballfun/fevalm to get faster evaluations

NBoulle:
This speeds up the evaluation of a ballfun by a factor of 1.5-2. It should also speed up the code in many places such as quiver for ballfunv.
All the ballfun tests pass.


ajt60gaibb:
Thanks, @NBoulle. Taking a look now.

ajt60gaibb:
Nice little speedup here.

PR #2369, opened 2020-08-15
Corrects a small error in aaa_autoZ

baddoo:
This corrects an issue in aaa.m where an automatic selection of sample points (aaa_autoZ) would set the number of Lawson iterations to infinity and thus disable cleanup.


nakatsukasayuji:
Thanks @baddoo ! Just note this bug was found while aaa-trig was developed/implemented by @baddoo .

PR #2370, opened 2020-09-16
slightly clarify AAA Loewner matrix, simplify residual

jrus:
The Loewner matrix can be computed in the same manner as the Cauchy matrix, one column at a time. Doing this in parallel style seems less confusing. It also saves some marginal computation, though the bottleneck is the SVD so it doesn’t make much practical difference.

If we are already computing the Loewner matrix and the Cauchy matrix, we can directly use those for computing the residual for a rational approximant, by multiplying the columns of each by the newly computed weights, taking a sum across columns, and then dividing. This seems simpler and more legible to me, but YMMV.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -115,36 +115,35 @@ reltol = tol * norm(F, inf);
 SF = spdiags(F, 0, M, M);
 
 % Initialization for AAA iteration:
-J = 1:M;
+J = logical(ones(M, 1));
 zj = [];
 fj = [];
+A = [];
 C = [];
 errvec = [];
-R = mean(F);
+R = F - mean(F); % residual
 
 % AAA iteration:
 for m = 1:mmax
     % Select next support point where error is largest:
-    [~, jj] = max(abs(F - R));          % Select next support point.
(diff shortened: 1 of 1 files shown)


jrus:
(The context for me making an essentially cosmetic change here is that I separately made an aaa_relative function which takes a pair of complex values for each value of Z and F, representing some anchor point + a displacement, because I am trying to see if I can get a Gopal & Trefethen ([2019](https://arxiv.org/abs/1804.08127)) style global AAA-approximant to a Schwarz-Christoffel map (e.g. from the disk to the isospectral drum shape) to converge all the way down to 11 or 13 digits of accuracy or however many I can  [under the assumption that the prevertices calculated by SC Toolbox are accurate]. That alternative version of aaa is clearer to code up when the Loewner matrix is calculated this way. But I also thought the style change clarified the original code a bit, hence the pull request here. Cf. [my notebook](https://observablehq.com/@jrus/scpie) about pie-slice approximations to SC maps, for which I want to have global approximation converging as far as I can for comparison, so I can make some plots demonstrating the relative convergence between the two approaches.)

PR #2372, opened 2020-10-09
Feature phaseplot colors

trefethen:
I've changed the colors in phaseplot a little to make them smoother, to avoid those awkward yellow/cyan/magenta stripes.
Here is the difference now between `phaseplot(@(z) z)` and `phaseplot(@(z) z,'classic')`:
colorfig

@nakatsukasayuji, I'd be grateful if you could check this out.  I had a bit of Git trouble along the way... hopefully everything is in shape.


jrus:
Let me highly highly recommend using different colors which do not create visual artifacts at arbitrary angles.

I made up what I think to be a much better color scheme a few years ago, and would be happy to help you integrate it.

You can see a variant in practice at Ricky Reusser’s notebook here, https://observablehq.com/@rreusser/complex-function-plotter

I’ll try to properly write up the details at some point, but here are a couple examples:

v f *z
bessel

trefethen:
@nakatsukasayuji, please hold off a little merging this -- I want to make a couple more changes.  Among other things it's driving me crazy to always type `axis equal` or `axis square`; I think `phaseplot` should do this.

@jrus, your color scheme is very attractive.  At the same time, the link with the books and calendars by Wegert is important to us, since these images are so well known, so for the moment, at least, I want the default color scheme to look more or less like his (although smoothed).  I suggest that after this code is merged, if you are interested, you might code a new branch 'rus' option along with the 'classic' option.  Please only do this if it will fit in the `phaseplot` code; we do not want to introduce additional files in this area which is not close to the core of Chebfun.

jrus:
@trefethen Fair enough. I communicated a bit with Wegert a few years ago, but the email chain dropped and I didn’t keep following up. I think a similar color scheme to this could be helpful for his work too; in particular, a comparable color scheme could be made to match the available color gamut of CMYK printers (the RGB rainbow version doesn’t translate very well).

I also would like to see if I could get the [DLMF](https://dlmf.nist.gov) people to adopt it, as I think it does a better job at meeting their “significant phases at recognizable hues” criterion than [their current color scheme](https://dlmf.nist.gov/help/vrml/aboutcolor).

I guess I should start by making a proper write-up, before doing more marketing push. :-)

trefethen:
Closing this now.  See `feature-phaseplot-colors2` instead.

PR #2373, opened 2020-10-15
Feature phaseplot colors2

trefethen:
@nakatsukasayuji, I would be grateful if you would review this improvement version of `phaseplot` (with smoothed colors as compared with Wegert's original).  You'll note that this branch is now called `feature-phaseplot-colors2` rather than `feature-phaseplot-colors`.  That's because I made a mistake I was unable to fix with my version of `feature-phaseplot-colors` -- which I will now accordingly delete.


nakatsukasayuji:
Looks very good to me.

PR #2374, opened 2020-11-25
feature-pswf

nickhale:
Some basic implementations allowing us to work with prolate spheroidal wave functions (PSWFs) in Chebfun.

In particular `P = pswf(N, c)` for constructing a degree N PSWF with bandwidth c and `[x,w] = pswfpts(N, c)` for computing associated quadrature formulas.

I'm not an expert on these methods, and the algorithms are almost certainly not the most state of the art, however they should be reasonably reliable for moderate values of N and c.


trefethen:
Wonderful stuff!  This will be an exciting addition to Chebfun.  Here are a few comments about `pswf`.  I haven't looked at `pswfpts` yet.

(1) Cosmetics.  I changed "NUMEL(N)" to "LENGTH(N)", "two-vector" to "2-vector", and "wavefunction" to "wave function" in the help text.

(2) Notation.  The help text needs to make it clear exactly what pswf(n,c) corresponds to in some standard notation e.g. from the Digital Library of Mathematical Functions.

(3) Reference.  Relatedly, I think it is important for the help text to state that we are following the development of [ref]. Should [ref] be DLMF, or Xiao-Rokhlin-Yarvin, or what?

(4) Example.  I've added a trivial example to the help text, just plot(pswf([1 3 5],100)).  Maybe another example or two of a more seductive nature should be added.

(5) Tests.  I don't see any test files.  Among the tests should be at least a couple of numerical ones comparing computed values to known ones.

(6) Numbering.  Isn't it standard to number n = 0,1,2,... rather than n = 1,2,3,...?

(7) Domain scaling.  I think there is a problem here.  For example pswf(10,1) looks very different from pswf(10,[-2 2]).

(8) 'coeffs' parsing.  It seems that to use the 'coeffs' mode, one has to give all three arguments: pswf(n,c,dom,'coeffs'). This seems unfortunate as specifying dom in particular should not be necessary.

(9) Legendre coefficients.  I note that the coeffs delivered often go far below machine precision, often with oddities in their shape below 1e-20, and sometimes with a tail of rounding errors at an amplitude like 1e-30, e.g. with psfw(1,1000,[-1,1],'coeffs'. This seems puzzling and I presume these numbers are not correct 
in any case.  Should the Legendre coefficients be trimmed at some point below machine precision?

(10) Eigenvalues.  The help text says this is a "vector of length N in ascending order" but I think in fact it's a vector of eigenvalues coresponding to whatever is in the scalar or vector N.  Thus if N = 5 we get one number, not 5, and if N = [3 1] we get two numbers in descending order. I'm guessing the functionality is fine and it's the help text that needs adjustment.

trefethen:
Now here are some comments on pswfpts, which also looks exciting, especially GGQ (which I have by no means looked into thoroughly).

(1) Cosmetics.  I made a few changes in the help text.

(2) Numbering.  Again we'll need to sort out n = 0,1,2,....

(3) Reference.  The user needs information about where these quadrature formulas come from.

(4) Example.  I've added an example showing off GGQ.

(5) Tests.  A test file is needed.

(6) 'GGQ' parsing.  For 'GGQ' mode, one has to give all three arguments: pswfpts(n,c,dom,'GGQ').  This should not be necessary.

(7) Symmetry.  The comments in the code mention that symmetry is not enforced in nodes or weights.  But isn't it trivial to do this, at least in cases where the domain is of the form [-L,L], to avoid user annoyance such as  middle node that is not quite 0?  It seems to me all that's needed is x = mean([x -flipud(x)],2); and w = mean([w; fliplr(w)]); .

nickhale:
> (2) Notation. The help text needs to make it clear exactly what pswf(n,c) corresponds to in some standard notation e.g. from the Digital Library of Mathematical Functions.

to do

> (3) Reference. Relatedly, I think it is important for the help text to state that we are following the development of [ref]. Should [ref] be DLMF, or Xiao-Rokhlin-Yarvin, or what?

to do

> (5) Tests. I don't see any test files. Among the tests should be at least a couple of numerical ones comparing computed values to known ones.

There is a test in `misc/test_pswf`. It tests both `pswf` and `pwsfpts` (but neither very rigorously).
 
> (6) Numbering. Isn't it standard to number n = 0,1,2,... rather than n = 1,2,3,...?

You're right. Done.
 
> (7) Domain scaling. I think there is a problem here. For example pswf(10,1) looks very different from pswf(10,[-2 2]).

Fixed, I think. Although I need to think more carefully about what the correct scaling should be.
(I also removed the option to pass a single input, since this complicates things and I'm not sure is useful.)

> (8) 'coeffs' parsing. It seems that to use the 'coeffs' mode, one has to give all three arguments: pswf(n,c,dom,'coeffs'). This seems unfortunate as specifying dom in particular should not be necessary.

Done.
 
> (9) Legendre coefficients. I note that the coeffs delivered often go far below machine precision, often with oddities in their shape below 1e-20, and sometimes with a tail of rounding errors at an amplitude like 1e-30, e.g. with psfw(1,1000,[-1,1],'coeffs'. This seems puzzling and I presume these numbers are not correct
> in any case. Should the Legendre coefficients be trimmed at some point below machine precision?

I've done this, but need to think about whether there's any reason that there might be accuracy in some of these values below machine precision. In particular, one needs to compute the derivative of the PSWFs when calculating GGQ nodes, so those extra values might be useful.

> 
> (10) Eigenvalues. The help text says this is a "vector of length N in ascending order" but I think in fact it's a vector of eigenvalues coresponding to whatever is in the scalar or vector N. Thus if N = 5 we get one number, not 5, and if N = [3 1] we get two numbers in descending order. I'm guessing the functionality is fine and it's the help text that needs adjustment.

I agree. Fixed.

nickhale:
> (2) Numbering. Again we'll need to sort out n = 0,1,2,....

Done.

> (3) Reference. The user needs information about where these quadrature formulas come from.

To do 

> (5) Tests. A test file is needed.

See `tests/misc/test_pswf.m`
 
> (6) 'GGQ' parsing. For 'GGQ' mode, one has to give all three arguments: pswfpts(n,c,dom,'GGQ'). This should not be necessary.

Fixed.

> (7) Symmetry. The comments in the code mention that symmetry is not enforced in nodes or weights. But isn't it trivial to do this, at least in cases where the domain is of the form [-L,L], to avoid user annoyance such as middle node that is not quite 0? It seems to me all that's needed is x = mean([x -flipud(x)],2); and w = mean([w; fliplr(w)]); .

Done.

trefethen:
Thanks @nickhale for these further changes.  I've made a number of further modifications in help texts, and also separate the test file into separate tests `misc/test_pswf` and `misc/test_pswfpts`.  

One thing needs to be fixed, and then as far as I'm concerned this can be merged: the PSWFs computed now sometimes have the wrong sign.  You will see that as a consequence, the final test in `test_pswf` now fails.  I believe probably the standard normalization is that a PSWF is positive at x=0 if it's even, and has positive derivative if it's odd.

trefethen:
I'm afraid I added a few more tests of numerical values against WolframAlpha, and a couple of the signs differ.   So it would seem the problem of choice of signs is not yet settled.  See `misc/test_pswf.m` (tests 6 and 7).

trefethen:
Further investigation makes me suspect WolframAlpha is getting the signs wrong, not us.  I have adjusted the tests accordingly so now if @nickhale agrees he is welcome to merge this.  (And meanwhile I will check the sign issue with Jim Bremer.)

Issue #2376, opened 2021-02-24
Phaseplot color


NBoulle:
I'm wondering if there is an issue with the colorbar scale of phaseplot due to this line: https://github.com/chebfun/chebfun/blob/master/phaseplot.m#L80
Below is the current phaseplot on the function f(z)  = z, indicating that the angle is -pi on the positive real line while it should be 0.
![current_phaseplot](https://user-images.githubusercontent.com/31421323/108983519-dd929b00-7686-11eb-99fe-b379f05c47ae.png)

I found that to match the description (i.e. having the colorbar ranging from [0,2*pi]), one should replace the line by
```
h = surf(real(zz), imag(zz), -ones(size(zz)), phi(angle(-f(zz)))+pi);
caxis([0,2*pi])
```
**Edit:** It might still not be correct, probably the easiest is `mod(phi(angle(f(zz))), 2*pi)`
![match_description_phaseplot](https://user-images.githubusercontent.com/31421323/108983780-1f234600-7687-11eb-8a68-a4830fc9a1e7.png)

However, if we prefer to have the angle ranging from [-pi,pi], then the line should be changed to
`h = surf(real(zz), imag(zz), ones(size(zz)), phi(angle(f(zz))));
`
![angle_normalized_phaseplot](https://user-images.githubusercontent.com/31421323/108983881-3e21d800-7687-11eb-8877-094d575d91f5.png)

A nice feature would be to add an option to choose the angle normalization ([-pi,pi] by default or [0,2*pi] if asked by the user).

trefethen:
@NBoulle, that's interesting.  I've never put a colorbar on a phaseplot, so I hadn't noticed.

Would you be prepared to fix this?  Just now I have a dropbox problem that keeps me from updating Chebfun code until my next visit to the office.

NBoulle:
@trefethen Sure, should I add an option to change the axis from [-pi,pi] to [0,2*pi] ? And if so do you have any name suggestion ?

I think that the default should be plotting from [pi,pi] (the latest picture) so the description will change a bit as well as the color scheme your are used to see (0 will be blue by default instead of red).

trefethen:
Thanks!  I suggest maybe
`phaseplot(..., 'caxis', [theta1, theta2])`
to override the default.

jrus:
[...]

PR #2377, opened 2021-02-24
fix the color range of phaseplot

NBoulle:
Fix https://github.com/chebfun/chebfun/issues/2376 and we can now plot the angles with any convention [theta, theta+2*pi].

### DIFF
--- a/phaseplot.m
+++ b/phaseplot.m
@@ -1,12 +1,14 @@
 function varargout = phaseplot(f, varargin)  % Plot phase portrait.
 %PHASEPLOT   Phase (= argument) plot of a complex function.
-%   PHASEPLOT(F), where F is a function handle or a CHEBFUN2 defining a
-%   complex function, draws a phase plot of F(Z) in the complex plane.
+%   PHASEPLOT(F), where F is a function handle, a CHEBFUN, or a CHEBFUN2 
+%   defining a complex function, draws a phase plot of F(Z) in the complex 
+%   plane.
 %   As arg(f(z)) ranges over [-pi,pi] the colors run
 %   cyan -> blue -> magenta -> red -> yellow -> green.
 %   
 %   If HOLD is ON, the existing axes are used.  If HOLD is OFF, the axes
-%   are taken as the domain of F if it is a CHEBFUN2, otherwise [-1 1 -1 1].
+%   are taken as the domain of F if it is a CHEBFUN or a CHEBFUN2, otherwise 
+%   [-1 1 -1 1].
 %   PHASEPLOT(F, [A B C D]) uses the axes [A B C D].
 %
 %   PHASEPLOT(F, 'CLASSIC') uses the color scheme from [1] rather than
@@ -20,6 +22,7 @@ function varargout = phaseplot(f, varargin)  % Plot phase portrait.
 %   phaseplot(@(z) z.^2)
 %   phaseplot(@(z) exp(1./z.^2))
 %   phaseplot(@(z) besselj(6,z),[-12 12 -5 5])
+%   phaseplot(cheb.gallery('bessel'))
 %   phaseplot(cheb.gallery2('airycomplex'))
 %   phaseplot(@(z) z,'caxis',[0,2*pi])
 %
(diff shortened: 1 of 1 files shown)


trefethen:
I've taken a look at the functionality and I'm afraid I lot of problems have been introduced.  Unfortunately I am unable to edit code right now, so @NBoulle, may I ask you to fix the following?  It would seem that you did not run the examples in the help text in the old and the new `phaseplot`, or you would have spotted these; please run them and you will see what I mean.

* The colors now appear always in "classic" mode rather than by default in smoothed mode.
* The method of plotting has been changed in a fashion so that the blank ink on an existing plot largely fades out.
* The newly introduced Bessel example seems meaningless.  What's supposed to be going on here?  What do you mean by a phaseplot of a univariate chebfun?
* The new `caxis` feature has no effect on anything shown on the examples.  Did you mean to add `colorbar`?

NBoulle:
- Hmm, I don't see any issue with the examples: they are plotted in non classic mode by default (I checked that classic = 0 in the code).
- What do you mean by that ? I didn't change any plotting command.
- Since the existing code accepts function handles of one variable, I thought it could also accept chebfuns (for instance f(x) = chebfun(@(x)sin(x))) and plot it's phase as it was a complex function.
- The point is that if the user asks for a colorbar, the value displayed will be correct, it you don't ask for a colorbar then it won't change anything (it will appear visually similar)

NBoulle:
In addition I correct some other issues like the grid was on, and set the xlim and ylim to the domain we plot the phase.
Old code:
image

New code:
image

NBoulle:
Colors are now plotted in smooth mode again:
image

trefethen:
@NBoulle, I'm very sorry, but none of my four bullet points seem to have been fixed, at least not as the images appear on my laptop.  Let me explain each one in turn.

(1) "The colors now appear always in "classic" mode rather than by default in smoothed mode."  This has now been fixed sometimes, but not fully, for adding `caxis` seems to mess up the fix.  Look at these two images resulting from `phaseplot(@(z) z)` (correct) and `phaseplot(@(z) z, 'caxis',[0,2*pi])` (incorrect):

fig4
fig5

(2) "The method of plotting has been changed in a fashion so that the blank ink on an existing plot largely fades out."  Here is one of the examples from the help text, before (correct) and after (incorrect):
fig1
fig2

(3) "The newly introduced Bessel example seems meaningless. What's supposed to be going on here? What do you mean by a phaseplot of a univariate chebfun?"  Sorry, I really don't think this makes sense.  A 1D chebfun lives on an interval.  If a user really wants to do a phaseplot in the complex plane, that's easy enough with e.g. `phaseplot(@(z) f(z))`, but it doesn't makes sense for this to be automated as a standard operation.  The example now in the help text seems only to reinforce the oddity of this operation--what is the point of the example?  Here is the plot from it, `phaseplot(cheb.gallery('bessel'))`:

fig3

(4) "The new caxis feature has no effect on anything shown on the examples. Did you mean to add colorbar?"  I am really perplexed here.  Isn't the point of this pull request to make colorbars come out properly labeled?  What good is done by the example now in the help text, `phaseplot(@(z) z,'caxis',[0,2*pi])`, which lacks a colorbar and whose output is therefore identical to that of `phaseplot(@(z) z,)`, namely this:
fig6

It seems to me that the crucial thing is to illustrate colorbar in an example, and the obvious choice is simply `phaseplot(@(z) z), colorbar`, which now thanks to you correctly gives this image:

fig7

NBoulle:
I think I have resolved points 1,2,4 now. 

Concerning (3), I changed the example to `chebfun(@(z)sin(2*pi*z))`. 
If you don't want to advertise that one can plot the phase of Chebfun objects, I'm happy to leave it as a hidden feature (i.e. it would still work if one provides a chebfun input as in the current code but we won't mention it in the description and examples).
The update I made in the code was just to use the domain of the Chebfun as axes by default.

trefethen:
OK, thanks @NBoulle, I'm happy with three of my four points now and I will merge the PR.

I'm afraid I still don't like `f = chebfun(...); phaseplot(f)` as an option and I would like to request that you remove this example from the help text.  A 1D chebfun [-1,1] can in no sense be expected to be accurate in the complex unit square, and it is just asking for confusion to present to users with the inevitable implication that it's a sensible thing to do.  As it happens, the current example `phaseplot(chebfun(@(z)sin(2*pi*z)))` produces the right picture for sin(2 pi z), but if you change 2 to 20 the picture not right for sin(20 pi z):
fig

PR #2385, opened 2021-05-24
Never run chebtest again 

ajt60gaibb:
In this PR I add a `matlab.yml` file to the folder `.github/workflows/`.  This makes chebtest run automatically whenever there is a pull request. In this way, we never have to run chebtest again ourselves (as they run on an ubuntu server supported by GitHub) and GitHub colors the pull-requests (green tick, amber dot, red cross) depending on the status of the testing suite.  

To test this pull request I forked the project, checked the `matlab.yml` file on the forked repo, and then issued a pull request from the forked repo.

### DIFF
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -12,4 +12,4 @@ jobs:
       - name: Run commands
         uses: matlab-actions/run-command@v0
         with:
-          command: addpath('.'); results = chebfun.test; assert(all([results{:,2}] > 0))
+          command: addpath('.'); results = chebtest; assert(all([results{:,2}] > 0))
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -12,4 +12,4 @@ jobs:
       - name: Run commands
         uses: matlab-actions/run-command@v0
         with:
-          command: addpath('.'); results = chebtest; assert(all([results{:,2}] > 0))
+          command: addpath('.'); results = chebtest; assert(all([results{:,3}] > 0))
--- a/.github/workflows/matlab.yml
+++ b/.github/workflows/matlab.yml
@@ -1,0 +1,15 @@
+name: chebfun Tests
+on: [push]
+jobs:
+  chebfun-test:
+    name: Run chebfun tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Install MATLAB
+        uses: matlab-actions/setup-matlab@v0
+      - name: Run commands
+        uses: matlab-actions/run-command@v0
+        with:
+          command: addpath('.'); results = chebfun.test; assert(all([results{:,2}] > 0))
--- a/tests/singfun/test_feval.m
+++ b/tests/singfun/test_feval.m
@@ -34,7 +34,7 @@ data.exponents = [-a, -b];
 data.singType = {'sing', 'sing'};
 f = singfun(fh, data, pref);
 err = norm(feval(f,x) - feval(fh,x), inf);
-tol = 1e3*eps;
+tol = 2e3*eps;
 pass(3) = err < tol;
 
 %%


.github/workflows/matlab.yml
```diff
@@ -0,0 +1,15 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+      - name: Install MATLAB
+        uses: matlab-actions/setup-matlab@v0
+      - name: Run commands
+        uses: matlab-actions/run-command@v0
+        with:
+          command: addpath('.'); results = chebtest; assert(all([results{:,2}] > 0))
```
danfortunato, line 15: It looks like the third column of `results` contains the pass/fail info. So this should be `results{:,3}`.

NBoulle: The following tests failed:
```
chebop2/test_generalVariableCoefficients.m
singfun/test_feval.m
```
in your fork test: https://github.com/ajt60gaibb/chebfun-1/runs/2659300312 but the commit is labelled as "pass"

ajt60gaibb: @danfortunato, do you know how to find out how to find out why a test breaks on the ubuntu server? I imagine I need to increase a tolerance or two by a small factor. It also means that the tolerance checks in
```
chebop2/test_generalVariableCoefficients.m
singfun/test_feval.m
```
fails on the latest ubuntu version.

NBoulle: In the latest commit, only `singfun/test_feval.m` fails. This one `chebop2/test_generalVariableCoefficients.m` crashed before so I don't know why the two runs are different.

danfortunato: `singfun/test_feval.m` fails on my laptop as well. The third test in there has an error that is slightly above the tolerance.

danfortunato: `chebop2/test_generalVariableCoefficients.m` fails locally for me too. I believe there is a bug in `denseSolve`:

https://github.com/chebfun/chebfun/blob/5fde1a4ab57847c1a0843e88c2b3145bc5410fc9/%40chebop2/denseSolve.m#L88

It should be:
```
sz = size(CC{1,1}, 1) * size(CC{1,2}, 2);
```

danfortunato: If `chebtest` used MATLAB's built-in unit testing stuff, then we could use this GitHub workflow to see more detailed test output on GitHub: https://github.com/matlab-actions/run-tests.

ajt60gaibb: Can you make a new issue and pull request for the `denseSolve` bug?

danfortunato: Yes, here it is: https://github.com/chebfun/chebfun/pull/2386

ajt60gaibb: I increased the tolerance of the third test in `singfun/test_feval.m` by a factor of 2. It appears that this passes now.  I would be happy if this PR is merged, and we work on #2386 that should fix the broken test. (I am finding it awkward to test a GitHub workflow feature that cannot be tested locally on my machine.)

NBoulle: The third test of `singfun/test_feval.m` failed before on my Ubuntu laptop with error `2.842170943040401e-13`, while the tolerance was `2.220446049250313e-13`. I'm happy with the PR.
It would be good if later, chebtest displays which specific test failed (e.g. by printing the array `pass`).

ajt60gaibb: I agree, @NBoulle. I think I would support Chebfun moving to MATLAB's built-in unit testing so that GitHub can display detailed test outputs.

NBoulle: I think that would be quite complicated. A more simple option is to update `chebtest` to display more information when the test fails. I can have a try. 
Also, I'm wondering whether it would be better to run the test in GitHub with the quiet mode (otherwise there are 1500 lines of test to go through).

danfortunato: @NBoulle I agree, quiet mode is probably the best option for GitHub.

ajt60gaibb: Any thoughts @nickhale or @trefethen?   Are you happy with this PR?

trefethen:
On the advice of @abigopal I am merging this!

PR #2386, opened 2021-05-25
Fix incorrect sizing in chebop2 when X and Y discretizations differ

danfortunato:
`chebop2` can fail when the discretization sizes in X and Y are not the same, due to a bug in the sizing of Kronecker products. This PR fixes that bug.

The following test was failing, but now passes:
https://github.com/chebfun/chebfun/blob/5fde1a4ab57847c1a0843e88c2b3145bc5410fc9/tests/chebop2/test_generalVariableCoefficients.m#L61-L70


ajt60gaibb:
Thanks, @danfortunato.

PR #2397, opened 2021-09-16
minor-fix-hermpts

### LINKED ISSUE #2394: Newton method in hermpts_rec can be improved
Dear all,

there is a slight inconsistency in Newton method in hermpts_rec, I think. After the for loop, x0 and dval are not synchronized, since dval comes from the previous value of x0. I would add a further evaluation [val, dval] = hermpoly_rec(n, x0); immediately after the for loop. Without this correction, I see a slight loss of orthogonality in the Hermite polynomials (I can provide an example).

nickhale:
Update derivative values after Newton iteration has converged in hermpts_rec.

Closes #2394

### DIFF
--- a/hermpts.m
+++ b/hermpts.m
@@ -580,8 +580,9 @@ for kk = 1:10
     x0 = x0 - dx;
     if norm(dx, inf)<sqrt(eps), break; end
 end
+[~, dval] = hermpoly_rec(n, x0);      % Compute derivatives once more.
 x = x0/sqrt(2);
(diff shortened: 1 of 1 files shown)


trefethen:
Thanks @nickhale.  @ajt60gaibb, perhaps you can handle this (just a couple of lines).  If that's a problem let me know and I'll do it.    - Nick T.

ajt60gaibb:
Thanks, @nickhale.

Issue #2399, opened 2022-01-12
bug in minimax


trefethen:
`minimax` is failing with a very simple problem:
```
x = chebfun('x');
f = abs(x-0.5);
p = minimax(f,2);
plot(f-p), grid on
```
There should be at least four equioscillatory extrema, but there are just three:
![errorcurve](https://user-images.githubusercontent.com/3601488/149179642-0a09302c-0c54-4b95-9fe0-d4029127fe89.png)
:

PR #2400, opened 2022-01-14
minor fix to minimax

sfilip:
This pull request addresses #2399. Fixes a bug that would cause points from outside the approximation domain to be added to the alternating extrema set in some rare occasions.


trefethen:
Thank you @sfilip!

PR #2406, opened 2022-04-10
Create clsap

LourencoPeixoto:
Classical least squares approximation problem.

A new algorithm for the classical least-squares approximation of `N` samples by an algebraic polynomial of degree at most `n` when the number `N` of samples is very large. For a fixed `n`, the complexity in double precision accuracy is `O(1)`. It is faster and more precise than `POLYFIT()` command.

Includes `GRAMPTS()` to compute the nodes and weights of Gaussian type quadrature for sums.

See: https://doi.org/10.1137/19M1259936 .

### DIFF
--- a/clsap
+++ b/clsap
@@ -1,0 +1,704 @@
+function [c,p,rss,r2] = clsap(xj, f, n, varargin)
+%%  CLSAP   Classical least squares approximation problem
+%   
+%   [C] = CLSAP(XJ, F, N) computes the coefficients from the polynomial of
+%   least squares of degree N in Gram basis given the equidistant data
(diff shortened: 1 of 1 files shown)


trefethen:
Lourenço - I must comment from the "user's point of view," as is often my role with Chebfun.  If you have an important improvement to `polyfit` that really matters to the operation of Chebfun, then it would seem to me the aim should be to improve `polyfit` accordingly, and explain clearly what improvement you believe you have made, not to introduce a new code with an obscure new name.  I do not know if you believe you are in that position or not, and indeed I am not sure what you propose the role of `clsap` in Chebfun would be, since it doesn't seem to deal with chebfuns but rather with vectors and matrices.

PR #2408, opened 2022-04-18
update second AAA literature reference

daanhb:
Two references are given in the AAA code. In the second reference the name of the first author was misspelled. Also, it seems the paper appeared in 2020.


nakatsukasayuji:
Looks good to me! Thanks for the detailed check, I certainly didn't notice it.

PR #2409, opened 2022-04-21
fix-idct-help-text

nickhale:
Fix an inaccuracy in the chebfun.idct help text (specifically how it relates to the idct in the Signal Processing Toolbox).

```
% Note that the above means that chebfun.idct(U) is not the same as idct(U),
% where idct(U) is the implementation in the Matlab signal processing toolbox.
%    The two are related by 
%        idct(U) = CHEBFUN.DCT(E*U)
%    where E = sqrt(n/2)*eye(n); E(1,1) = sqrt(n);

U = rand(5);
n = size(U,1);
E = sqrt(n/2)*eye(n); E(1,1) = sqrt(n);
norm(chebfun.idct(E*U) - idct(U))
ans =
   3.9607e-16
```

### DIFF
--- a/@chebfun/idct.m
+++ b/@chebfun/idct.m
@@ -12,7 +12,7 @@ function y = idct(u, type)
 %   Note that the above means that CHEBFUN.IDCT(U) is not the same as IDCT(U),
 %   where IDCT(U) is the implementation in the Matlab signal processing toolbox.
 %   The two are related by 
-%       IDCT(U) = CHEBFUN.DCT(E*U)
+%       IDCT(U) = CHEBFUN.IDCT(E*U)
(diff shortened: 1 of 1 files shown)


ajt60gaibb:
Yes, this was clearly a typo. I guess we are accepting pull requests that break the online tests now. However, this PR certainly doesn't make the situation worse.

nickhale:
Thanks Alex.

Yes. The following tests failed or crashed:
     chebfun/test_trigpade.m
     chebmatrix/test_norm.m
     chebop2/test_transport.m

(Although clearly none of these were caused by the present pull request.)

test_trigpade and test_transport both pass on my machine, so I can't see the cause. test_norm is a tolerance issue.

PR #2411, opened 2022-05-04
Fix prz

trefethen:
The `prz` and `prztrig` methods are usable by users, but not advertised since they are basically tools for `aaa` and `aatrig` (curiously not for `minimax`, which uses its own code to compute poles, residues, and zeros of rational functions).  For historical reasons, `prz` and `prztrig` have an argument "r" that is never used.  This has now been deleted and the help text and `aaa` and `aaatrig` codes updated accordingly.


nakatsukasayuji:
I'm ok with looking at this but the 'All checks have failed' worries me a bit. FYI on my laptop the test chebmatrix/test_norm.m passed. Should this be dismissed?

nakatsukasayuji:
Otherwise looks good to me (and the test failure has nothing to do with the changes). I will merge.

Issue #2412, opened 2022-05-08
bug in minimax


trefethen:
The minimax command returns incorrect numerator p and denominator q in approximating cos(x) by rational functions of type (m,n) = (0,2), (1,2), (0,3), (1,3), (0,4) and (1,4).  I understand from comments by @sfilip that it's actually computing the rational function (barycentric form) correctly, but then not generating the right p and q.  @sfilip, I hope you can fix this?   Thanks!

sfilip:
Just created a pull request (#2413) that should resolve this.

PR #2413, opened 2022-05-09
fix bug in minimax for computing p and q

sfilip:
This pull request addresses #2412. Fixes a bug that would not generate correct p and q polynomials of the minimax rational approximation r when Chebyshev points used to generate p and q map to support points in the barycentric representation of r.


trefethen:
Looks good to me.  Thanks @sfilip!  (The test that fails concerns gpr, nothing to do with this code.)

PR #2415, opened 2022-05-28
Feature-pantograph

nickhale:
Add functionality for solving pantograph-type delay differential equations via Chebop.

(See `tests/chebop/test_pantograph.m` for some examples. An Example to be added soon.)

### DIFF
--- a/@adchebfun/adchebfun.m
+++ b/@adchebfun/adchebfun.m
@@ -555,7 +555,7 @@ classdef (InferiorClasses = {?chebfun}) adchebfun
             if ( isa(r,'chebfun') )
                 r = mat2cell(r);
             end
-            operatorBlock/operatorBlock.m
+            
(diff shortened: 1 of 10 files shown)


tobydriscoll:
In blockFunction, why did you switch the order of arguments in 75?

```
blockFunction( @(u) compose(g, u) )
```

nickhale:
> In blockFunction, why did you switch the order of arguments in 75?
> 
> ```
> blockFunction( @(u) compose(g, u) )
> ```

I think because `CHEBFUN/COMPOSE(F, OP) returns a CHEBFUN representing OP(F)` ?

I admit to being confused on this point, and getting it the wrong way round a few times, so this could be wrong. 

However, I have a vague recollection of it not working the other way around, which is maybe why I changed it.

PR #2416, opened 2022-06-05
Feature aaa cleanup

trefethen:
AAA has had a problem in approximating functions with singularities.  Poles should be exponentially clustered near the singularities with exponentially decreasing residues, but the "cleanup" procedure was removing them.  This has now been changed so that the test of size of residue is scaled by distance to the approximation set.  The same change was made in AAATRIG.

One complication came up.  Sometimes a pole is right on the boundary of the approximation set, and AAA is usually good at catching this.  To preserve this feature, the scaling of the residue test, which used to involve the maximum size of the function F on the set Z, is not based on the geometric mean size of F on Z.

Various tests have been run to confirm that this change in Cleanup should not have serious effects.  Of course the Chebfun tests have been run, and so have the examples in the AAA and AAATRIG test files.  Additionally, various AAA demonstration codes that I have in my files have been run.  These still work, and in some cases, the change has enabled me to remove "cleanup, off" instructions that never should have had to be there.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -83,7 +83,7 @@ function [r, pol, res, zer, zj, fj, wj, errvec, wt] = aaa(F, varargin)
 %   complex rational minimax approximation, SIAM J. Sci. Comp. 42 (2020),
 %   A3157-A3179.
 %
-% See also CF, CHEBPADE, MINIMAX, PADEAPPROX, RATINTERP.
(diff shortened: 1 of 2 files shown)


nakatsukasayuji:
Looks good to me. Cleanup might undergo further changes in the future (e.g. should it be turned on by default?)

PR #2417, opened 2022-06-08
remove unintended keyboard

nakatsukasayuji:


nakatsukasayuji:
So sorry, an unintended debugging mode was left when I pushed. Removing that and I hope nobody will kill me to go ahead and merge it.

PR #2418, opened 2022-06-14
Feature-Gauss-Jacobi-Lobatto-quad

nickhale:
The approach we use to compute the Gauss-Legendre-Lobatto nodes in `lobpts` extends to the Jacobi-Lobatto case. The same is true for Gauss-Jacobi-Radau points.

One could solve for the missing weights at the end(s), but there are even explicit formulas (due to Gautschi).

```
>> [x, w, v] = lobpts(4, .2, .7)
x =
  -1.000000000000000
  -0.338147393222043
   0.483074929453927
   1.000000000000000
w =
  Columns 1 through 2
   0.052000500524441   0.644040571638338
  Columns 3 through 4
   0.859532535973245   0.148340882172585
v =
  -0.320698598703037
   0.865616354236654
  -1.000000000000000
   0.455082244466383
```

### DIFF
--- a/lobpts.m
+++ b/lobpts.m
@@ -1,5 +1,5 @@
-function [x, w, v] = lobpts(n, varargin)
-%LOBPTS   Gauss-Legendre-Lobatto quadrature nodes and weights.
+function [x, w, v] = lobpts(n, alp, bet)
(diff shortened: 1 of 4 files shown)


trefethen:
Thanks @nickhale !

PR #2420, opened 2022-09-03
Update compose.m

federicofuentes:
Added support for 'trigx' and 'trigy'. This retains the mixed structure when operating over these objects, e.g. multiplying two 'trigx' objects via mtimes is still 'trigx'; real of 'trigx' is still 'trigx'.


ajt60gaibb:
Sounds very reasonable to me! 👍

ajt60gaibb:
Thank you, @federicofuentes. This is welcomed pull request.

PR #2424, opened 2022-12-15
spin:  Recurse on SUBSREF when accessing properties.

aaustin141:
If we don't do this, we can't further index properties that are returned without storing them in a separate variable.  This results in unexpected behavior, e.g.,

    >> S = spinop('ks');
    >> S.domain(2)
    ans =
       1.0e+02 *
                       0   1.005309649148734

With this fix in place, we get the expected result:

    >> S = spinop('ks');
    >> S.domain(2)
    ans =
         1.005309649148734e+02

### DIFF
--- a/@spinoperator/subsref.m
+++ b/@spinoperator/subsref.m
@@ -15,8 +15,15 @@ switch index(1).type
     case '.'
         
         % Allow access to any of the properties of S:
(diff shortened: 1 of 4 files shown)


trefethen:
Thanks @aaustin141.

PR #2425, opened 2023-01-23
fix help text typo

heatherw3521:
Just changing comments/helptext to correct a typo in diskfun help on sum and mean.


trefethen:
Looks good.

PR #2426, opened 2023-03-06
minor fixes for the chebfun3f constructor

cstroessner:
ensured that the chebfun3f constructor accepts chebfun3 objects as input functions

prevented collapsing dimensions in phase 3 of the chebfun3f constructor

ensured that the fixedRank setting works with the chebfun3f constructor 

now throws an error when the non-compatible flags 'trig' and 'chebfun3f' are set simultaneously

### DIFF
--- a/@chebfun3/chebfun3f.m
+++ b/@chebfun3/chebfun3f.m
@@ -327,14 +327,12 @@ while ~happy
         [K, QWK] = DEIM(QW);
 
         % Simplification:
-        % (Take a linear combination of the columns = faster)
         lenU = standardChop(chebvals2chebcoeffs(sum(Uf,2)), pref.chebfuneps);
         lenV = standardChop(chebvals2chebcoeffs(sum(Vf,2)), pref.chebfuneps);
         lenW = standardChop(chebvals2chebcoeffs(sum(Wf,2)), pref.chebfuneps);
-%         % (Full simplify = slower)
-%         lenU = length(simplify(chebfun(Uf, [dom(1),dom(2)], pref), pref.chebfuneps));
-%         lenV = length(simplify(chebfun(Vf, [dom(3),dom(4)], pref), pref.chebfuneps));
-%         lenW = length(simplify(chebfun(Wf, [dom(5),dom(6)], pref), pref.chebfuneps));
+        lenU = max(lenU,size(Uf,2));
+        lenV = max(lenV,size(Vf,2));
+        lenW = max(lenW,size(Wf,2));
 
         % Convert to coefficients and simplify:
         QU = chebvals2chebcoeffs(QU); QU = QU(1:lenU,:);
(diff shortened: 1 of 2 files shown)


cstroessner:
Currently the chebfun3f constructor is not the default option. Once the minor changes proposed here have been integrated into chebfun, the main issues that need to be overcome would be:

1. The objects computed with the chebfun3f constructor sometimes lead to issues when using the diff function. The coefficients stored in the factor matrices are close to machine precision (which is compensated by very large values in the core tensor). In this case the error of the diff function tends to be very large. This problem might potentially be circumvented by multiplying the factor matrix with a scalar before computing the derivative. However, this change would require modifications of chebfun3 outside of the constructor.
2. The chebfun3f constructor is not able to work with trigonometric basis functions yet. This change should not be too difficult. One would need to replace the Chebyshev grid points by the interpolation nodes for trigonometric interpolation. One would need to change the commands to compute the interpolants and the heuristic to determine accuracy. Lastly one needs to remove the current error message.
3. Once the previous two issues are resolved, chebtest('chebfun3') with chebfun3f as default constructor will pass all but three test cases. The cases chebfun3/test_fevalt.m and chebfun3/test_minus.m fail due to minor tolerance issues. The case chebfun3/test_subsref.m will never work with chebfun3f, since it checks if the function in the column is equal to the univariate input functions. Since we store a multiplicative factor in the core tensor this equality will only be satisfied up to scalar multiples and not exactly.

Once these changes have been made, I would recommend to use the chebfun3f constructor as default constructor for chebfun3 objects. This could be changed directly in the default parameter of prefStruct.constructor.

bhashemi:
The new changes look good to me and so if no one objects I'd be happy to merge this pull request. It reduces the number of chebfun3 tests that fail from 44 to 16 (on my laptop). All the 16 tests that fail belong to one of the three categories listed in the comments by @cstroessner.

For the record I'd like to point out an issue regarding `chebfun3/test_minus.m` that is perhaps not minor: 

```
% with the new chebfun3f
>> [x, ~, ~] = cheb.xyz; f = x;
>> g = (1-1e-7)*x; norm(f-g)
ans =
   1.6330e-07

>> g = (1-1e-8)*x; norm(f-g) % I keep getting zero for values less than 1e-8 as well
ans =
     0
```

Here is what happens with the default chebfun3 constructor:
```
>> [x, ~, ~] = cheb.xyz; f = x;
>> g = (1-1e-7)*x; norm(f-g)
ans =
     1.6330e-07

>> g = (1-1e-8)*x; norm(f-g)
ans =
   1.6330e-08

>> g = (1-1e-13)*x; norm(f-g)
ans =
   1.6335e-13

>> g = (1-1e-14)*x; norm(f-g)
ans =
     0
```

trefethen:
Thanks very much @cstroessner and @bhashemi !  I am happy for you to merge if that seems appropriate.  

To both of you -- is there something new here that should be highlighted in a Chebfun example?  As always, if so, I am happy to be involved in polishing and publishing.

bhashemi:
@trefethen, I just played a bit with the last two sections of [this example](http://www.chebfun.org/examples/approx3/Complexity.html) and got the following plot for _tough_ functions (section 3):
tough3d


Here is the corresponding plot for _easy_ functions (section 4 of the same example):
easy3d

@cstroessner, both plots show how nice Chebfun3f is! I will be sending you an email with the code that generated these plots.

PR #2427, opened 2023-03-10
Make chebfun3f the default chebfun3 constructor.

cstroessner:
I modified chebfun3f to be able to handle the trig flag. When set chebfun3f now correctly returns an approximation based on trigonometric functions on a uniform grid.

If we were to use the chebfun3f constructor as default constructor the following 10 test cases fail
   chebfun3/test_diff.m
   chebfun3/test_diffx.m
   chebfun3/test_diffy.m
   chebfun3/test_diffz.m
   chebfun3/test_feval.m -> minor tolerance issue
   chebfun3/test_fevalt.m -> minor tolerance issue
   chebfun3/test_gradient.m
   chebfun3/test_guide.m -> the trig approximation uses slightly more coefficients -> I do not simplify so far
   chebfun3/test_minus.m -> major tolerance issue
   chebfun3/test_subsref.m -> test 8 will never work since chebfun3f stores a multiplicative factor in the core tensor

The remaining failed test cases are related to the computation of derivative.


cstroessner:
By modifying the way the scaling is performed and by including the latest changes of the branch explore-chebfun3f. This version of chebfun3 now passes all sample tests when using the chebfun3f constructor.
The chebfun3f constructor now replaces the classical constructor. The classical constructor can still be accessed using the 'classical' flag.

trefethen:
Thanks so much @cstroessner and @bhashemi for your work on Chebfun3.

One really small thought: I think "classic" is better than "classical".

cstroessner:
I now set chebfun3f as default constructor. To access the old one needs to set the 'classic' flag.

I also slightly refactored the constructor. It now only does the parsing and then calls chebfun3classic, chebfun3f or chebfun3double depending on the situation.

PR #2428, opened 2023-03-13
revised aaa.m so mmax does not turn on Lawson

trefethen:
It has long bothered me that in the aaa code, specifying 'mmax', m is identical to specifying 'degree', m-1, and in particular, it turns on the Lawson iteration.  This is a problem since Lawson isn't (yet!) completely reliable.  

So I've made changes in the AAA code to separate mmax, which now doesn't turn on Lawson, from degree, which does.

For example:

```
X = chebpts(500); F = exp(X);
subplot(121), r = aaa(F,X,'degree',3), plot(X,F-r(X))
subplot(122), r = aaa(F,X,'mmax',4), plot(X,F-r(X))
```
fig

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -2,8 +2,8 @@ function [r, pol, res, zer, zj, fj, wj, errvec, wt] = aaa(F, varargin)
 %AAA   AAA and AAA-Lawson (near-minimax) real or complex rational approximation.
 %   R = AAA(F, Z) computes the AAA rational approximant R (function handle) to
 %   data F on the set of sample points Z.  F may be given by its values at Z,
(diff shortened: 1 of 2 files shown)


nakatsukasayuji:
Looks good to me. Just to note the original example isn't giving me an error; in both aaa codes (before or after this pull request),

f = @(x) abs(x+.5+.01i);
rng(8), X = 2*rand(1000,1) - 1;
[r,pol,~,~,~,~,~,errvec] = aaa(f(X),X,'mmax',60); r(1)
[r,pol,~,~,~,~,~,errvec] = aaa(f(X),X,'degree',59); r(1)
both give me 1.5000.

Issue #2423, opened 2022-12-02
AAA may return incorrect final error


WilliamRJohns:
For example if there are 2 support points and one is removed for having zero weight, interpolation is broken and the final error returned may be incorrect. In the following example the final l_inf error should be 1 but is returned as 0.
x=[1:1:5];
Fx=[ones(1,4) 2]
[aaa_aprox,~,~,~,zj,fj,wj,errvec,~]=aaa(Fx,x);
approx=aaa_aprox(x)
%Final error 0 returned but error is 1
errvec
%%

Fx =

     1     1     1     1     2

approx =

     1     1     1     1     1

errvec =

     1
     0

trefethen:
[aaa03_supportpts.pdf](https://github.com/chebfun/chebfun/files/10981999/aaa03_supportpts.pdf)

William, I'd be pleased to be in touch directly about this if you're interested.  If so, would you send me an email at trefethen@maths.ox.ac.uk ?  Thanks.

PR #2431, opened 2023-03-17
revised aaa for issue 2423

trefethen:
This is a fundamental change at the heart of AAA, prompted by issue #2423.  The problem is that AAA has never checked the error at any of the support points, assuming there is interpolation there.  But that assumption fails when a weight is zero.  So with this revision, AAA also checks the error at support points with zero weights, if any. 

I've written a memo explaining the details:
[aaa03_supportpts.pdf](https://github.com/chebfun/chebfun/files/11002556/aaa03_supportpts.pdf)


nakatsukasayuji:
The 11th aaa_test is failing. This is perhaps an error in the test code not aaa. 
Z = [0, 1];
F = [1, 2];
r = aaa(F, Z);
pass(10) = ( norm(F - r(Z), inf) < tol );
pass(11) = ( r(inf) == -inf );

trefethen:
Sorry @nakatsukasayuji, I thought I had fixed that test.  Anyway I think I've fixed it now.  Please try again.

nakatsukasayuji:
Looks good as far as I can see. (This change is a bit hard to check, as examples with weight=0 are not easy to find. In any case it works fine for the generic case, and the 3-point example above.)

PR #2432, opened 2023-03-21
new standalone AAA

trefethen:
I have modified AAA so that it can be used as a standalone MATLAB code.  In particular it now includes functions prz and reval, so these are no longer called as external Chebfun functions.  All tests still pass.

At the moment the code remains rather long, but most of the action is in the first 100 or so executable lines.  After that come functions PARSEINPUTS, CLEANUP, CLEANUP2, AAA_AUTO, PRZ, and REVAL.  In the upcoming weeks I hope further simplifications will be made especially in the CLEANUPs and in AAA_AUTO.

### DIFF
--- a/tests/chebfun/test_aaa.m
+++ b/tests/chebfun/test_aaa.m
@@ -1,13 +1,12 @@
 function pass = test_aaa(pref)
-% Test for aaa().  Many new tests added August 2019.
+% Test for aaa.  This has been written to be independent of Chebfun.  
 
 % Get preferences.
 if ( nargin < 1 )
-    pref = chebtech.techPref();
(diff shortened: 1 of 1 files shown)


trefethen:
Just changed a couple of lines to make this compatible with GNU Octave.  (Turns out MATLAB can compute the nullspace of a 0xn matrix, but Octave can't.)

trefethen:
I have removed the two "Developer notes".  Both of these commented on aaa.m as part of the Chebfun software system, whereas the point of standalone AAA is to be independent of the Chebfun software system (even though still hosted in the Chebfun repo).

PR #2434, opened 2023-03-22
corrected vectorize flag usage in chebfun3f

cstroessner:
Previously the vectorize flag was used incorrectly in Phase 2 of the chebfun3f constructor.


cstroessner:
When running test_imag.m in pass(2) the function ff in line 193 of chebfun3f changes sign depending on how it is evaluated. 
We have T = ff(X,Y,Z) with T(1,1) = -0.9, but ff(X(1,1),Y(1,1),Z(1,1)) = 0.9. This should not happen and leads to a crash.

PR #2435, opened 2023-03-25
improve-chebfun3-plus

bhashemi:
Ensure chebfun3/plus works fine with objects of different tech. 
Also added a test which did not work previously


cstroessner (approved):
These changes look good.

PR #2437, opened 2023-04-15
change prz to remove zero weights to avoid spurious poles

nakatsukasayuji:
The prz.m code, when the input vector wj contains an entry that is exactly 0, outputs a spurious pole that is not a pole of the rational function. This is because the generalized eigenvalue problem that the prz code is based on proceeds by multiplying the node polynomial to each partial fraction in the barycentric form, then finding the roots of that polynomial. This fix remedies this issue.


trefethen:
Thanks @nakatsukasayuji.  Looks good.

PR #2442, opened 2023-07-26
Vectorize chebfun3/feval in the case of unstructured evaluation points

danfortunato:
This pull request significantly speeds up `chebfun3`'s `feval` when the evaluation points are unstructured (i.e. not a tensor-product grid). See below for a comparison:

```
f = chebfun3(@(x,y,z) sin(x.*y.*z));

n = 50000;
x = 2*rand(n,1)-1;
y = 2*rand(n,1)-1;
z = 2*rand(n,1)-1;

tic
ff = f(x,y,z);
toc

% Before this PR: 2.247853s
% After this PR:  0.028487s
```

### DIFF
--- a/@chebfun3/feval.m
+++ b/@chebfun3/feval.m
@@ -113,14 +113,14 @@ if ( (m == 1) && (n>1) ) % Row vector instead of column vector
     zVec = zVec.';
 end
 
-outVec = zeros(size(xVec));
 colsVals = f.cols(xVec); % So, colsVals(:,j) = feval(f.cols(:,j),xVec).
 rowsVals = f.rows(yVec);
 tubesVals = f.tubes(zVec);
-for i = 1:size(xVec, 1) % TODO: How to vectorize this?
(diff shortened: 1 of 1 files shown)


bhashemi:
All test pass and I am, in principle, happy to merge this. The only thing I am not sure about is the fact that the new code will not work on MATLAB versions earlier than 2020b because `pagemtimes` was introduced then. Can someone please remind me of our procedure in such situations? Do we normally add something as follows:

```
 if ~isMATLABReleaseOlderThan("R2022a"). % Check if `pagemtimes' is available 

     "then use pagemtimes like this pull request"
else
     "use old lines which are removed in this pull request"
end
```

or we should not worry about backward compatibility? Perhaps @trefethen or @nickhale can comment on this?

trefethen:
My feeling is that 2020b would be a big concern if this were a core part of Chebfun, but since other things don't depend so much on Chebfun3, it's ok.

PR #2444, opened 2023-09-14
Fix syntax for prz and reval in aaa.m

sslchi:
Replace t-z' in subfunction prz with bsxfun(@minus, t, zj.'), and fix the syntax for Cauchy matrix in subfunction reval.


trefethen:
Thank you, but I am afraid we are aware that bsxfun is needed with older Matlabs, and we have decided not to write codes this way.  Matlab's post-2016 method of "implicit expansion" is so powerful and natural for Chebfun-style operations that we are embracing it.  We realize that poses a problem for people with old Matlabs.

PR #2448, opened 2024-05-23
spell check code using `codespell`, add "FYI" CI job

cbm755:
First, I fixed a few typos manually.

Then I tried to setup a CI job that runs `codespell` automatically as a GitHub Action.  For now, its set to only warn and not fail, b/c there are many typos to fix before it passes.

In my experience, `codespell` is worth it: there will be some exceptions to make for false positive but these are far out-numbered by the correct suggestions.


cbm755:
I run this locally like:
```
codespell -I .codespell-ignorewords -x .codespell-ignorelines -f -H -S .git --count
```

Currently, its outputting 333 errors.  @nickhale before I keep going, are you likely to merge this?

cbm755:
There was some concern in our meeting that this would create friction if someone adds new code b/c they might need to manage the exception list (say for variable names).

  - I've set the job so it never fails (it always passes with a green check)
  - The output looks like this currently:
    image
  - no one will see this unless they drill down into the CI report

So my plan is to fix remaining typos and then I propose this for merging.

PR #2450, opened 2024-07-15
adding the sign SVD modification option to aaa.m

trefethen:
I am planning to refactor `aaa.m` in the upcoming weeks.  This pull request, however, is not related to that (and the syntax may ultimately change when we refactor).  It adds an option `sign` to the aaa call to invoke a different use of the SVD which @nakatsukasayuji and I have found works much better for sign functions.  It also applies this option in the Lawson phase of AAA, if `lawson` as well as `sign` has been specified.

When we refactor, and after we have learned more, there is a chance we will want to make this or something like it the AAA default.  For the moment, however, the default operation is unchanged and `sign` is simply a new option.  Definitely worth using if you are approximating sign functions or similar!


nakatsukasayuji:
Looking good. Should I merge?

trefethen:
Yes please!

PR #2451, opened 2024-07-18
tweak to Lawson, only when sign on is specified

trefethen:
This is a tweak to AAA-Lawson (namely, underrelaxation) that seems to make a nice difference in suppressing oscillations that stop convergence.  It needs to be explored thoroughly, and may certainly evolve as we do that.  For the moment it's desirable to have it in Chebfun since we're doing a lot of approximation of sign functions for Zolotarev purposes, and the risk is negligible since the tweak only applies to a non-default Lawson option that was introduced just two days ago.


nakatsukasayuji:
looks fine

PR #2453, opened 2024-07-31
AAA Lawson damping

trefethen:
This change to AAA adds a new feature to improve performance in difficult AAA-Lawson cases: `'damping',d` to introduce a damping ratio other than the default of 1.  There will be no effect on codes that do not specify `'damping'`....

...with one small exception: the `'sign'` flag now only affects AAA iteration as before, not AAA-Lawson iteration, since its rather mysterious action in that case is now supplanted by this less mysterious damping option.

See Trefethen unpublished memo Rat348 for illustrations.  This change is likely to be improved upon the future, but for the moment, it increases AAA's ability to get minimax approximations in difficult cases, like the Fermi-Dirac example from Fig. 5 of Nakatsukasa & Trefethen 2020.

Hopefully @heatherw3521 can take a look, and @nakatsukasayuji will also be interested.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -33,6 +33,8 @@ function [r, pol, res, zer, zj, fj, wj, errvec, wt] = aaa(F, varargin)
 %   - 'lawson', NLAWSON: take NLAWSON iteratively reweighted least-squares steps
 %       to bring approximation closer to minimax.  Specifying NLAWSON = 0 
 %       ensures there is no Lawson iteration.  See next paragraph.
+%   - 'damping', DAMPRATIO: when running Lawson, apply a damping ratio at each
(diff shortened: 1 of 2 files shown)


heatherw3521:
@trefethen do we want to include `sign` and input related to damping in `aaa_autoZ` ? Currently those are not passed in.

trefethen:
Thanks @heatherw3521.  Well spotted.  I didn't think of this since I never use the automated feature, but I've added it in now.  I hope it looks ok?

As you know I dream of refactoring `aaa.m` systematically one day, and the dream is that when this happens, everything will become clearer.  Admittedly it isn't happening yet.

heatherw3521:
@trefethen @nakatsukasayuji  Looks good to me!

PR #2480, opened 2025-08-19
feature-improved-chebtest-display

nickhale:
Chebtest will now display which subtests are failing. This makes it easier to diagnose failures that occur on the CI but are not easily reproduced locally.


cbm755 (approved):

cbm755:
LGTM!  +1 for merge when you're happy.

PR #2485, opened 2025-08-27
modified AAA residues calculation

trefethen:
The `aaa.m` calculation of residues is inaccurate in the sense that much smaller residuals can often be achieved if we replace it by a residue calculation based on a least-squares fit.  This pull request does the replacement.

Approximation of functions with poles at infinity may be problematic.  We don't know everything about calculating residues!  See 
[rat385_residues.pdf](https://github.com/user-attachments/files/22011246/rat385_residues.pdf)
for information justifying (I hope) this change.

### DIFF
--- a/aaa.m
+++ b/aaa.m
@@ -268,6 +268,14 @@ if ( deriv_deg >= 1 )
     r = r_output;
 end
 
+% Compute more accurate residues, if nargout > 2:
+if ( nargout > 2 )
+    deg = max(0, length(zer)-length(pol) );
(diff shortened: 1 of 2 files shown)


trefethen:
Thanks @nakatsukasayuji, including for the new test.  

About `prz`, since it's a standalone code in Chebfun that takes as input only information about a barycentric fraction (not about an approximation domain or data values for AAA), I figured we had to leave it as is.  Luckily `aaa.m` has its own `prz` included internally, which could accordingly be bypassed in the `aaa.m` calculation of residues.

I'll go ahead and merge.

PR #2487, opened 2025-08-29
aaa: use `&&` and `||` in scalar logic

cbm755:
@trefethen I was brought up on the mantra that `&` is for bitwise operations on matrices and that one should use short-circuiting `&&` and `||` inside `if` statements.

I think this isn't mandatory on Matlab.  But `if (a & b)` is a warning in Octave.

(I checked and many parts of Chebfun use `&&` and `||`, although I haven't exhaustively checked.)


trefethen:
I'm happy with && instead of &, though I haven't attempted to check any of this.

Issue #2461, opened 2025-02-09
Minor bug and some questions about Fredholm operators


MichaelPPieper:
There are (at least) 2 versions of the fred command in Chebfun: one takes in a function handle for the kernel K, the other takes in a chebfun2. My primary question: is the latter the "better" option? Suppose for example I wanted to numerically approximate the eigenvalues of an integral operator. Is building a chebfun2 more stable, more accurate, more efficient, etc than just passing the function handle to chebfun/fred.m?

The minor bug is in the former version of the `fred` command. In chebfun/fred.m, line 58 sets the options as
`opt1 = {'resampling', false, 'splitting', true, 'blowup', 'off', 'vscale', normv}`
I don't think setting `'splitting'` to `true` actually activates splitting. chebfun.m seems to just check whether the string passed is equal to `'on'`; it then converts that to a logical value when building the `pref` passed to the constructor.

A secondary, less important question: does anyone recalls how these options were determined? Why is splitting always turned on, even if the user doesn't ask for it to be? Also, why is the vertical scale set to the norm of v? There is a brief note about this in the comments, but I guess I don't fully understand why this is the correct choice (or how vscale affects the constructor more generally).

Thank you!

ajt60gaibb:
Hi @MichaelPPieper, the first approach is usually better if the kernel is non-smooth (perhaps explaining why splitting is "on") while the latter is better when the kernel is smooth. 

vscale will affect the constructor: the constructor aims for a final accuracy that depends on vscale.  (I think that is the only place where vscale changes the constructor in any meaningful way.)

PR #2489, opened 2025-08-29
Fix-chebfun-fred

nickhale:
`chebfun(@(x) abs(x), 'splitting', 1)` now throws an error.

Previously it would simply not perform splitting and henceresult in a `Warning: Function not resolved using 65537 pts.` error for this particular example. 

This was pointed out in #2461 as `@chebfun/fred' was calling the constructor in this way.


cbm755:
Excellent, and adds a unit test too.

splitting is pretty common, no concern about folks relying on existing (buggy) behaviour?

If that's not a concern, happy to see this merge.

PR #2496, opened 2026-06-23
Feature aaa noise

tonylxu:
Add optional noise plateau detection and chopping algorithm to AAA.


trefethen (commented):
I like this code but I find it seems too conservative about chopping.  Suppose in particular we run this code:
```
np = 200;
X = linspace(-1,1,np);
F = sin(10*X);
rng(1)
for k = 1:12
   d = randi(8) + 4; ep = 10^(-d);
   F2 = F + ep*randn(size(X));
   [~,~,~,~,~,~,~,errvec] = aaa(F2,X,'noise_chop',1);
   subplot(4,3,k)
   semilogy(errvec,'.-'), grid on
   hold on,plot([1 length(errvec)],ep*[1 1],'--r'), hold off
   ylim([1e-16 1e4])
   set(gca,'ytick',10.^(-16:8:0))
end
```
Here is the result.  The trouble seems to be that the noise plateaus are not long enough for AAA to detect them.  But this seems wrong; to the eye these are obviously plateaus.

Image

